PR #27, opened 2014-07-30
Misc things: New textures, new creatures and new gold stacks + related code.

Bertram25:
@hwoarangmy : Could you review that one and merge if ok?

Thanks! :)


source/GameMap.cpp
```diff
@@ -1662,7 +1662,7 @@ std::vector<Room*> GameMap::getReachableRooms(const std::vector<Room*> &vec,
     return returnVector;
 }
 
-Room* GameMap::getRoomByName(std::string& name)
```
hwoarangmy, line 1665: Thanks for correcting all the const I forgot :)

PR #30, opened 2014-07-31
Real Time Shader System fixes

Bertram25:
This pull request is about fixing the RT Shader system initialization and all the material warnings.

This also fixes the handling of normal on linux on my machine.

The only bad counter part is that a GLSL shader script is triggering errors but this is a bug in Ogre.
I'll try to find a workaround in another commit for that.

Also, we should be able to start re-enabling normals after applying this one. :)


Bertram25:
@oyvindln Could you review my work? It seems to me you're the most at ease with all that kind of stuff...

Bertram25:
Thanks for the review.

PR #32, opened 2014-08-04
Full server implementation

hwoarangmy:


hwoarangmy:
Why did you added the game map as member of all the game entities?

=> Because it was already done through setGameMap. As you have seen, I have added another GameMap into the server. So, with setGameMap, it would have been possible to change an entity from a GameMap to another, which would be a problem. And finally, because there was a TODO in GameEntity saying that GameMap should be added through the constructor instead of setGameMap. Which I agree to.

Why did you move back the minimap to the ODFrameListener?
=> Because it is a rendered component and now, there are 2 GameMaps. So I had to take it out so that server GameMap do not create it.

hwoarangmy:
BTW, sorry for the big pull request. I have tried to not do anything else than what was needed for the full server paradigm. But it is quite a big change...

Bertram25:
> > Why did you added the game map as member of all the game entities?
> 
> Because it was already done through setGameMap. As you have seen, I have added another GameMap into the server. So, with setGameMap, it would have been possible to change an entity from a GameMap to another, which would be a problem. And finally, because there was a TODO in GameEntity saying that GameMap should be added through the constructor instead of setGameMap. Which I agree to.

Ah, yes. Makes sense. In fact, I wasn't sure there was a new game map object in the server. /me shouldn't review code at 10pm on work days. ;)
I also like the fact that a crucial member is set in the constructor as well.

> > Why did you move back the minimap to the ODFrameListener?
> 
> Because it is a rendered component and now, there are 2 GameMaps. So I had to take it out so that server GameMap do not create it

Indeed.

> BTW, sorry for the big pull request. I have tried to not do anything else than what was needed for the full server paradigm. But it is quite a big change...

Np. Your changes are quite readable so it's a pleasure so far. Btw, please don't take any of my questions as offenses or such. It's just me trying to stay in the loop. :)

Ok for the merge then.

hwoarangmy:
No problem, I do not take your questions as offence. Actually, I even think it's good that you really take time to read such a big pull request :)

PR #33, opened 2014-08-04
A few little fixes

Bertram25:


Bertram25:
Could you @hwoarangmy ? :)

source/RoomTreasury.cpp
```diff
@@ -167,18 +167,15 @@ RoomTreasury::TreasuryTileFullness RoomTreasury::getTreasuryTileFullness(int gol
 {
     if (gold <= 0)
         return noGold;
-
-    if (gold <= maxGoldinTile / 4)
```
hwoarangmy, line 172: All this else if are useless since they all return

source/RoomTreasury.cpp
```diff
@@ -167,18 +167,15 @@ RoomTreasury::TreasuryTileFullness RoomTreasury::getTreasuryTileFullness(int gol
...
         return half;
-
-    if (gold <= maxGoldinTile / 4 * 3)
+    else if (gold <= maxGoldinTile / 4 * 3)
         return threeQuarters;
 
-    if (gold <= maxGoldinTile)
-        return fullOfGold;
+    //if (gold <= maxGoldinTile)
+    return fullOfGold;
```
hwoarangmy, line 187: This would have been enough :)

akien-mga, line 187: Maybe it should also cater for gold > maxGoldinTile, or is it 100% sure that sure a situation cannot happen?

source/RoomTreasury.cpp
Bertram25, line 172: ok, I'll change that to stay with ifs.

source/RoomTreasury.cpp
Bertram25, line 187: Ok, i'll add a warning for such a possible case.
To be clear, I usually don't like triggering exceptions for such small things and I prefer warn the user/dev in logs instead. After all, the player doesn't care if he/she has too much gold because we code badly. He/she only wants to keep playing.

akien-mga, line 187: A warning in the log seems sensible to me; if it's printed out to the stdout when run from a terminal, any user could notice it and file a bug report if our code is bad :-)

source/RoomTreasury.cpp
hwoarangmy, line 172: should I wait for your change or will you do another pull request ?

Bertram25, line 172: I'll redo the commits within the same pull request. But it won't be before this evening.

If you have possible conflicting changes to propose, don't wait for me, I'll rebase if needed. :)

hwoarangmy, line 172: No, I don't think I will have something to do on these files anyway :)

Bertram25:
@hwoarangmy  @akien-mageia I've updated the pull request as discussed.

Cheers,

PR #43, opened 2014-08-06
Use of c++ polymorphism for GameEntity instead of C style switch/cast

hwoarangmy:


hwoarangmy:
@Bertram25 could you have a look please ?

hwoarangmy:
The 2nd commit should fix the room not beeing drawn. But concerning the not marked tiles, I don't see why. On my virtual ubuntu, it works.

Bertram25:
Patches looked fine. Do you really think the *Local suffix is needed?

EDIT: Oops, after a bit more testing. I've got this crash when placing a treasury tile:
An internal Ogre3D error ocurred: OGRE EXCEPTION(5:ItemIdentityException): SceneNode 'Treasury_13_12_9_node' not found. in SceneManager::getSceneNode at /tmp/buildd/ogre-1.9-1.9.0+dfsg1/OgreMain/src/OgreSceneManager.cpp (line 920)
Internal Ogre3D exception: OGRE EXCEPTION(5:ItemIdentityException): SceneNode 'Treasury_13_12_9_node' not found. in SceneManager::getSceneNode at /tmp/buildd/ogre-1.9-1.9.0+dfsg1/OgreMain/src/OgreSceneManager.cpp (line 920)
**\* ERROR: Internal Ogre3D exception: OGRE EXCEPTION(5:ItemIdentityException): SceneNode 'Treasury_13_12_9_node' not found. in SceneManager::getSceneNode at /tmp/buildd/ogre-1.9-1.9.0+dfsg1/OgreMain/src/OgreSceneManager.cpp (line 920)

Bertram25:
As for the not colored marked tile, I think I've got clues. It is linked on the use of the RT Shader system.
I'll make a fix for that.

Bertram25:
@hwoarangmy Sorry for all the messy comments here. I'll open proper issues. :)
Again, thanks for all the work done.

hwoarangmy:
The *Local suffix is not needed but it allows the functions to remember their state and not create/destroy meshes several times. With this, when the *Local function is called, there is no need to check. This way, the functions do not need to return a bool saying if the mesh is already created/destroyed and testing at each stage.

Bertram25:
Ah, so from what I read the *Local functions are used as unsafe ones, guarded by the safer no-suffix ones? I guess that's what you mean?

If I'm right, and only if, in other projects, we were using the prefix keyword 'actually' for such cases. i.e.: `ActuallyDeleteYourSelf().`

I've also seen people doing the reverse. for instance, call the guarded function *IfNeeded(). i.e:

``` c++
deleteYourselfIfNeeded()
{
    if (mDone)
        return;
    deleteYourself();
}
```

Don't misunderstand me, I think the simplification is good and should stay, I do even think we need more of such refactoring for better design. I'm just nit picking on function names here as I do think one of those will be more readable. ;) (You know, own colors and tastes)

If it's the case and if you're ok with it, I'll do a small pull request to change those names to use either one or the other, or something even better that you may know of?

Best regards,

Issue #15, opened 2014-07-29
Basic Keeper AI


Bertram25:
The current AI is working, yet it still needs a few additions:
- [x] The AI should be able to build treasure rooms, hatchery, quarters and a training hall
- [x] The AI should take care of building rooms to get the maximum "active spots" (represented by room models), by using squares surrounded by accessible and claimable dirt.
- [x] The AI should drop fighters near-by other enemy creatures to trigger some fun, once it is ready and has got a few fighters trained.

Bertram25:
Ah, while I'm thinking about it:
The AI should check it's available rooms from time to time. In case it already has them at start, or whenever it has got a few ones even partially destroyed.

Bertram25:
Also, IMHO, the AI shouldn't use traps atm, as they aren't fine tuned and may make the AI invincible.

hwoarangmy:
IMHO, we should create some indicators telling the creatures needs. For example, the average hunger, the number of creatures without a bed, the average level (for training room).
This way, it would be easy for the AI to know which room is needed and when to build a bigger/new room

Bertram25:
Yes, while you're already one step further. :)
I'd like the AI to create proper square/rectangle room shapes first and monitor their filling with convenient room tiles. Once this major step is done, the rest is just having fun with AI coding, IMHO. ;)

hwoarangmy:
Fun... and lot of headaches ;)

Bertram25:
Indeed!

Danimal696:
i can remember neither in Dk1 or dk2 the AI put traps/doors, all were pre-laid at map design, even if the ai player had a full workshop

Bertram25:
> i can remember neither in Dk1 or dk2 the AI put traps/doors, all were pre-laid at map design, even if the ai player had a full workshop

True. Yet, even with starting lighter, I hope we'll make it better even if it will never be perfect. :)

Bertram25:
Ok, it's official, I'll be on that from now on if nobody else strongly wants that one. ;)

hwoarangmy:
Concerning AI, according to what we said during the lan, I've also had a look and do you think all those classes are really needed ?
- A factory, when we will have, at most, 4 or 5 AI. IMHO, it can be removed (and eventually be replaced by the AIManager if we want a wrapper for AI creation).
[...]

PR #72, opened 2014-08-17
Add support for training and 2 segfault fixes

hwoarangmy:


hwoarangmy:
@Bertram25 Can you have a look ?

Bertram25:
Looks great, my small request apart. :)

I'll test it live this evening. Thanks a lot for all the good you've done!

source/RoomDojo.cpp
```diff
@@ -20,56 +20,258 @@
...
+}
+
+void RoomDojo::removeCreatureUsingRoom(Creature* c)
+{
+    Room::removeCreatureUsingRoom(c);
+    if(mCreaturesDummies.count(c) > 0)
     {
-        loadRoomObject(getGameMap(), "TrainingDummy1", mTopWallsActiveSpotTiles[i], 0.0);
+        Tile* tileDummy = mCreaturesDummies[c];
+        OD_ASSERT_TRUE(tileDummy != NULL);
```
hwoarangmy, line 193: To continue our talk about assertion, this example is the typical way I see for the use of the macro : Adding a log to the log file and try to continue. If we are lucky, we will not have a crash and the game will continue. If not, we will have a crash later but we will know looking at the log that something weird happened here.
As you can guess, assertion should never fail so in any case, even if it doesn't lead to a crash, a log with failed assertion should never be considered as normal.
And the good point with the macro is that I had though of adding information about the assertion in the macro as a second paramter supposed to be a string to be added in the assertion failed log containing info on why the assertion failed. If it is a macro, the string would be computed only if the assertion fails. If it is a function call, it will always be computed even if it the assertion doesn't fail.

Bertram25, line 193: Looks sane the way those asserts are used, to me. Do you think we should talk about those in the coding guidelines?

> If it is a function call, it will always be computed even if it the assertion doesn't fail.

Ah yes. I won't be nit-picking here since it's not worth it. The soul of my counterpoint about changing it to a function was more about code readability anyway. :)
Maybe, adding a few "\ & new-lines" will make it more readable, for instance?
If it's readable, I don't care if it's a macro to be honest.

hwoarangmy, line 193: > Looks sane the way those asserts are used, to me. Do you think we should talk about those in the coding guidelines?

If you think it is useful, yes. But we have to make clear that assertions should not fail and that code that produces failed assertions should not be included before being fixed. Otherwise, we will end up with logs full of useless failed assertions.

> Maybe, adding a few "\ & new-lines" will make it more readable, for instance?

Ok.

> If it's readable, I don't care if it's a macro to be honest.

Well, it's always interesting to have another coder's point of view since there might be some cases I've not thought about or even better ways to achieve the same goal that I don't know :)

Bertram25, line 193: > If you think it is useful, yes. But we have to make clear that assertions should not fail and that code that produces failed assertions should not be included before being fixed. Otherwise, we will end up with logs full of useless failed assertions.

Yeah, I was thinking about the philosophy of their use as well.

> Well, it's always interesting to have another coder's point of view since there might be some cases I've not thought about or even better ways to achieve the same goal that I don't know :)

Eh. :) there is the usual macro vs compiled code question, but here it is not worth the case, IMHO, and I'm sure you've already spoken of such things before with other ones, e.g.:
http://stackoverflow.com/questions/1137575/inline-functions-vs-preprocessor-macros

akien-mga:
@hwoarangmy's branch builds fine here. I'm now playing a while to see if the mentioned bugs are indeed fixed.

akien-mga:
I got a segfault when expanding a training room. We can see your assert function live :-)

> ERROR: Assert failed file /home/akien/Projects/opendungeons/OpenDungeons/source/ODClient.cpp line 584
> signal 11 (Segmentation fault), address is 0x40 from 0x5f0920
> 
> [bt]: (1) ./OpenDungeons.bin : Room::addRoomObject(Tile_, RoomObject_)+0x20 [0x5f0920]
> [bt]: (2) ./OpenDungeons.bin : Room::addRoomObject(Tile_, RoomObject_)+0x20 [0x5f0920]
> [bt]: (3) ./OpenDungeons.bin : ODClient::processOneClientSocketMessage()+0x259e [0x5c3c5e]
> [bt]: (4) ./OpenDungeons.bin : ODClient::processClientSocketMessages()+0x20 [0x5d08b0]
> [bt]: (5) ./OpenDungeons.bin : ODFrameListener::frameStarted(Ogre::FrameEvent const&)+0x192 [0x5d5bf2]
> [bt]: (6) /lib64/libOgreMain.so.1.9.0 : Ogre::Root::_fireFrameStarted(Ogre::FrameEvent&)+0x44 [0x7f44997aef04]
> [bt]: (7) /lib64/libOgreMain.so.1.9.0 : Ogre::Root::_fireFrameStarted()+0x21 [0x7f44997b07e1]
> [bt]: (8) /lib64/libOgreMain.so.1.9.0 : Ogre::Root::renderOneFrame()+0x9 [0x7f44997b08d9]
> [bt]: (9) /lib64/libOgreMain.so.1.9.0 : Ogre::Root::startRendering()+0x2d [0x7f44997b092d]
> [bt]: (10) ./OpenDungeons.bin : ODApplication::ODApplication()+0x3fc [0x5c0c9c]
> [bt]: (11) ./OpenDungeons.bin : main+0x1e [0x46747e]
> [bt]: (12) /lib64/libc.so.6 : __libc_start_main+0xf0 [0x7f44966a0f90]
> [bt]: (13) ./OpenDungeons.bin() [0x468844]

akien-mga:
I got another similar segfault when expanding a forge, so this narrows down the search for the faulty instruction to the room removal commit 301dcab.

akien-mga:
For the reference, for this last segfault, both the server and the client segfaulted with the same error and the same backtrace.

hwoarangmy:
Is it systematic when you expand a room ?

> I got another similar segfault when expanding a forge, so this narrows down the search for the faulty instruction to the room removal commit

No, the segfault comes from the room which is NULL (as the assertion tells) when adding the object. I have done quite a lot of changes in the way room objects are handled so there might be a bug. I will try to reproduce. If I cannot, I will add to OD_ASSERT_TRUE another parameter to be displayed (I was planning on doing that anyway) and some logs to see which rooms are created and why does it fail.

> For the reference, for this last segfault, both the server and the client segfaulted with the same error and the same backtrace.

That's the expected behaviour since client and server are supposed to work the same way

akien-mga:
It's not systematic no. From my few tests, I would say that it happens mostly (only?) when expanding a room using a selection that contains part of the existing room (e.g. I have a 4 tiles room, I select two of them to 4 new tiles to expand the room, and segfault).

I'll test some more to see if I can narrow it down to specific cases. I guess it can't be a latency issue between server and client that lets room be NULL for a few milliseconds?

akien-mga:
Another guess: could it be linked to the needed reorganisation of the dummies/forge elements when expanding the room? I can't reproduce the issue with the treasure room it seems, and I just encountered two consecutive crashes when expanding a training room with dummies (I started with a small training with no dummies, then expanded it little by little until there was enough place for dummies to appear, and the next expansion caused a crash).

akien-mga:
Ok it seems my latest tests confirm my guess from the last comment. Here is what I could gather:
- Treasury and Dormitory expansion do not cause segfaults.
- Training room, Forge and Library cause segfaults.
- When expanding a room with no room-specific models (dummies, forge elements, bookshelves + pedestal), no crash happens.
- Once a room is big enough to have room-specific models, a crash happens when you add one (1) tile to the room, with one exception: is the room is just big enough to have one model (3x3 or 2x4), expanding the room by one tile does not trigger a segfault, but expanding it a second time by another tile (or directly two tiles or more at once) triggers the segfault.

So I'd advise to have a look at the model placement code and how it could relate to the room being NULL.

akien-mga:
Dummies do get kicked by creatures now, btw :-D

akien-mga:
The branch stills builds fine with the addition of commit b1b2d82. The crashes still happen (I know the patch was not necessarily aimed at fixed that, but since it touches a bit of the room code, I wanted to make sure).

I'll be able to do further testing tomorrow in the evening if another patch comes in :-)

hwoarangmy:
Thanks for the tests Akien. As you already guessed, it was just a change of the names as discussed previously. So far, I couldn't reproduce the crash but I will try with what you just said.

hwoarangmy:
Ok, I can reproduce the issue. I've added some logs and I'll have a look. The good point is that now, I can see better where we should put logs to make them usefull to debug this kind of issue :)

hwoarangmy:
I've understood the issue. The bug is that active spots are refreshed when room tiles are removed (to allow removing active spots when a room becomes smaller). But in this case, room absorption, it makes the server send a message concerning room objects on the old room. The patch is on its way

akien-mga:
Good job hwoarangmy!

One small issue: you forgot to include LogManager.h in Room.cpp.
Otherwise I've built your branch (with the missing include) and the issue seems to be fixed (I just tried with making a big Library in several room absorptions).

I can do some more thorough tests in the evening, but I'd say that if Bertram can compile and does not see obvious regressions in game, he can merge the pull request (when the missing include has been added of course).

akien-mga:
I confirm that the game no longer segfaults when destroying an AI dungeon temple.

I've let the game idle for a while, now I see that many of my creatures are not reactive anymore (they're just sitting in the base), and I can't see to order diggers to expand my base. Maybe a consequence of my destroying of an AI base? (I did not see their behaviour change, I was afk).

Will do more tests later, but this issue should not prevent merge this pull request I think, it's most likely related to something else (or should be fix in a second step).

For the reference, I'm now getting these two warnings in the terminal:

> Computing turn 2128
> > 20 loops in Creature::doTurn name:Kobold205 colour: 4. Breaking out..
> > 20 loops in Creature::doTurn name:Kobold231 colour: 4. Breaking out..

_Edit:_ I dropped my yellow mechanical digger (don't know the exact creature's name) back into my base (it was roaming alone in a conquered enemy-base), and he woke up all my creatures :-)

akien-mga:
This game is starting to look real good :-D I even see AIs with training rooms and their units playing around with dummies!

akien-mga:
I've also confirmed that the game does not crash when removing room tiles (I don't know if it used to, but since @hwoarangmy spoke of it, it good to know that it's working properly).

Bertram25:
> This game is starting to look real good :-D I even see AIs with training rooms and their units playing around with dummies!

Eh eh. I made it so the AI create a training room south of the temple and a dormitory north of it when there is, well, room for it. :)
The last two steps to add as for AI is to make it place the needed rooms more dynamically, and have an attack phase.

> I confirm that the game no longer segfaults when destroying an AI dungeon temple.

Confirmed as well on my own. :) Eh eh I actually had a hard time destroying one since I didn't started the map very strategically. ;)

> I've also confirmed that the game does not crash when removing room tiles (I don't know if it used to, but since @hwoarangmy spoke of it, it good to know that it's working properly).

Confirmed as well. For the record and just in case you don't know it already, @hwoarangmy meant that when you add room tile next to a same-type room, the former room gets "absorbed" by the new one. It means the new room will take the room objects of the former one, its ground tiles, and the former room will be deleted.
The last crash was because the active spots (the creature available slot number to use the room) made the server try to update the room objects of the old room while it was deleted if i understood correctly.

I'll merge. Nice job @hwoarangmy :) I'm just sorry you have so much problems with that nasty room code.

Bertram25:
Note to self: Add this asap.

> One small issue: you forgot to include LogManager.h in Room.cpp.

EDIT: Done.

hwoarangmy:
> One small issue: you forgot to include LogManager.h in Room.cpp.

Yes, I used the repository where I had many changes concerning the hatchery. To make the 2 last patches, I had to remove the changes I had done (git gui is pretty useful for that). But it seems like I forgot LogManager. I should stop commiting after 2am :)

> I've let the game idle for a while, now I see that many of my creatures are not reactive anymore

I've seen it too but I did no investigation at all for that.

> Maybe a consequence of my destroying of an AI base?

When doing the training dummy patch, I used to use a map with no AI. And the behaviour was the same after a while. But as I said, I didn't look so I'm not sure. But I guess it is not related to this pull request. It is more likely that this behaviour was not seen because the game used to crash when the AI dungeon temple was killed and the game did not play enough to reach this stade. Again, I might be wrong.

> Eh eh I actually had a hard time destroying one since I didn't started the map very strategically. ;)

In this map, the only real attacking unit is the initial construct worker. As it can be dropped in dirt tile, you can just drop him near an AI dungeon temple and he will destroy everything alone :)
BTW, every other unit have weapons with 0 attack. But it seems they kill each other. I've not looked why but since the beginning I think it is strange as I was expecting them to fight forever.

> The last two steps to add as for AI is to make it place the needed rooms more dynamically, and have an attack phase.

Yes and I believe there are simple things we can do to make the AI seem more clever. Like creating rooms in the dirt (so that there are walls to increase the active spots) and drop 1 or 2 workers when it does so. We could even share the code when we will pick up idle workers (for AI or when you click on the worker icon).

> The last crash was because the active spots (the creature available slot number to use the room) made the server try to update the room objects of the old room while it was deleted if i understood correctly.

Yes, that's almost it. If you are interested, to say everything, the problem was that during room absorption, the room objects in the original room are moved in the new one and then, the tiles are removed. But because the activespots were refreshed when tiles were removed, if the room size was big enough to accept active spots after removing 1 tile, it created it. And a message was sent to the client from the old room with the consequences you already know.

Bertram25:
> > I've let the game idle for a while, now I see that many of my creatures are not reactive anymore
> 
> I've seen it too but I did no investigation at all for that.

The described behavior is here from the beginning. I can confirm it is also often linked to the warnings Akien has seen:

> Computing turn 2128
> 20 loops in Creature::doTurn name:Kobold205 colour: 4. Breaking out..
> 20 loops in Creature::doTurn name:Kobold231 colour: 4. Breaking out..

And indeed, the game wasn't that stable before so reaching the turn 2128 without encountering any other bug was unlikely before. ;) Thus it's a good step forward already.

> Yes and I believe there are simple things we can do to make the AI seem more clever. Like creating rooms in the dirt (so that there are walls to increase the active spots) and drop 1 or 2 workers when it does so. We could even share the code when we will pick up idle workers (for AI or when you click on the worker icon).

I've added the missing icon list in #8 and on the forums to request help, and I've added an issue about removing the obsolete ones in #74.
The AI improvement proposal is described in #15. But feel free to complete whatever you see fit. :)

> In this map, the only real attacking unit is the initial construct worker. As it can be dropped in dirt tile, you can just drop him near an AI dungeon temple and he will destroy everything alone :)

Yeah, and this is a flaw, since you will rather use your level 30 worker than your level 10 fighters. ;)
Maybe the worker max level should be set to something lower to prevent such a cheat. I'd like to open an issue about this. Ok if we make workers stuck at level 10?

(And later on, we shouldn't be able to drop a worker on a undiscovered location once we have some fog of war implemented.)

> BTW, every other unit have weapons with 0 attack. But it seems they kill each other. I've not looked why but since the beginning I think it is strange as I was expecting them to fight forever.

The damage done seems still hardcoded: https://github.com/OpenDungeons/OpenDungeons/blob/development/source/Creature.cpp#L1970
We'll need to balance this but I'd request to open a thread on the forums first to get everyone's input.

Bertram25:
Ah, and don't forget to reorder the room icons when adding the hatchery.
The hatchery should be placed in third position, after the Treasury and the Dormitory, in order to stay logical.

I know I'm demanding, sorry. ;)

hwoarangmy:
> Yeah, and this is a flaw, since you will rather use your level 30 worker than your level 10 fighters. ;)
> Maybe the worker max level should be set to something lower to prevent such a cheat. I'd like to open an issue about this. Ok if we make workers stuck at level 10?

In fact, this is not an issue. I let a powerfull worker (by having better attack than fighters) in this map because it was a test map. If fighters had better stats, I would have set the worker even better as he was there to allow to kill the other dungeons easily to test what happens.

> The hatchery should be placed in third position, after the Treasury and the Dormitory, in order to stay logical.

Ok, I will do another commit for that

> I know I'm demanding, sorry. ;)

No problem ;)

hwoarangmy:
> The damage done seems still hardcoded

I've not looked very deep but it seems the damage is related to the weapons : https://github.com/OpenDungeons/OpenDungeons/blob/development/source/Creature.cpp#L2183
But there must be something like that since we've all seen AI creatures killing each other so in the end, they end up doing damages

hwoarangmy:
> But there must be something like that since we've all seen AI creatures killing each other so in the end, they end up doing damages

I had not seen https://github.com/OpenDungeons/OpenDungeons/blob/development/source/Creature.cpp#L2180
That explains why damage is not 0...

Bertram25:
> I had not seen https://github.com/OpenDungeons/OpenDungeons/blob/development/source/Creature.cpp#L2180
> That explains why damage is not 0...

Ah ok. Having at least 1 as a base for damage looks fine to me, btw.

Bertram25:
> In fact, this is not an issue. I let a powerfull worker (by having better attack than fighters) in this map because it was a test map. If fighters had better stats, I would have set the worker even better as he was there to allow to kill the other dungeons easily to test what happens.

Sure. That said, the small one is perfect for quick skirmishes between friends. :)

PR #82, opened 2014-08-19
Restore stats windows when middle clic on creature

### LINKED ISSUE #78: Refresh creature infos when stats window open

hwoarangmy:
fixes #78
- Fixed a small bug in the hunger system (stupid me)

### DIFF
--- a/source/ClientNotification.cpp
+++ b/source/ClientNotification.cpp
@@ -48,6 +48,8 @@ std::string ClientNotification::typeString(ClientNotificationType type)
             return "askBuildTrap";
         case ClientNotificationType::ackNewTurn:
             return "ackNewTurn";
+        case ClientNotificationType::askCreatureInfos:
+            return "askCreatureInfos";
         default:
             LogManager::getSingleton().logMessage("ERROR: Unknown enum for ClientNotificationType="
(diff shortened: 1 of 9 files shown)


hwoarangmy:
@Bertram25 @akien-mageia Could you have a look ? It's not a big one this time :)

akien-mga:
Builds fine and the stats window works too. From my side it's okay to be pushed, maybe @Bertram25 will want to look through the code quickly before pushing.

Bertram25:
> maybe @Bertram25 will want to look through the code quickly before pushing.
> Yep, and code looks fine. :)

Thanks for bonus fix about hunger when dropped. I looked at the condition when reviewing the former patch but misinterpreted it. (I'm also stupid, eh. ;])
Merging...

hwoarangmy:
> I looked at the condition when reviewing the former patch but misinterpreted it. (I'm also stupid, eh. ;])

Yes, it can be tricky as hunger works the opposite as awakeness (from 0 to 100 instead of 100 to 0). But it was the most understandable IMO so I prefered to use it.

Bertram25:
> But it was the most understandable IMO so I prefered to use it.

I agree: Hunger grows when not eating. The reverse would be weird.

PR #96, opened 2014-08-20
Add replacement sound for dump.ogg and remove reference to non-imported model

akien-mga:
As asked for by @Danimal696 on http://forum.freegamedev.net/viewtopic.php?f=39&t=5695#p58857
The CREDITS file was also reworked for better legibility (this might not be the definitive version though, if we want to describe each file explicitly with its path).


hwoarangmy:
I let @Bertram25 have a look as I am not used to license stuff

Issue #42, opened 2014-08-05
Make sure the CMAKE_CXX_FLAGS are actually defined in CMakeLists.txt


akien-mga:
As reported by several users on the forum, compiling OD does not always work out of the box since most of the time the "-std=c++11" flags is not actually defined.

On the affected systems (among which Mageia 64bit), it seems that the CMAKE_COMPILER_IS_GNUCXX variable is empty, therefore the flags are not added to CMAKE_CXX_FLAGS.

I've found that moving the project declaration ( https://github.com/OpenDungeons/OpenDungeons/blob/development/CMakeLists.txt#L205-L229 ) at the top of the CMakeLists.txt (that is before the compiler flags definition) solves the issue on Mageia.

@Bertram25: IIRC you said something about this triggering issues on @tomluchowski's desktop?

On a side note, the whole CMakeLists.txt could probably use some refactoring; I'm not a cmake expert (yet), but some things look a bit fishy (e.g. listing of source files before project declaration and looking for dependencies).

Bertram25:
> @Bertram25: IIRC you said something about this triggering issues on @tomluchowski's desktop?

@akien-mageia Yep, moving the 'project' statement to the beginning of the file broke it for @tomluchowski AFAIK. That's why we did nothing and added the CMake flags by hand.

IMHO, the whole file needs to be redone anyway and follow more conventional ways of doing it as you said. I'd start it all over based on a better model and see whether things improve afterwards.
Just my two cents, though.

Bertram25:
As I said in #58, feel free to create a pull request that should fix that point.

Bertram25:
And you said you were only dabbling at it? It's working great here :D :+1: 

In fact, it's working so greatly that every warnings are enabled again and I get a lot of them. (Not complaining, we should have got them back weeks ago.)
This is working for me, and I hope it will work for @hwoarangmy too.
If so, I do think we should merge and deal with warnings afterwards.

akien-mga:
Hehe, in my pull request I just moved the whole chunk at the top of the file, it wasn't particularly complicated :-)
I've reverted my changes towards having proper make install instructions since it was a WIP, and I wanted something functional for a pull request. I've saved my work as git patches though, so I'll can use them to proceed with my work on the make install instructions (#58).

Bertram25:
Good strategy! :)

hwoarangmy:
Unfortunately, it doesn't :(
[...]

Issue #10, opened 2014-07-29
Move source/ files in sub-folders according to their subject.


Bertram25:
E.g: Move all the entities speaking files into a Entities/ sub-folder. Move the game map related files into a GameMap/ subfolder, …

A precise list of what should move is to be done first, though,

Bertram25:
A proposal:
**In source/entities/:**
BattleField.cpp/h
Building.h
Creature.cpp/h
CreatureAction.cpp/h
CreatureDefinition.cpp/h
CreatureSound.cpp/h
DirectionalTrap.cpp/h
GameEntity.cpp/h
MapLight.cpp/h
MissileObject.cpp/h
ProximityTrap.cpp/h
Tile.cpp/h
Trap.h
TrapBoulder.h
TrapCannon.h
Weapon.cpp

**In source/goals:**
AllGoals.h
Goal.cpp/h
GoalClaimNTiles.cpp/h
GoalKillAllEnemies.cpp/h
GoalMineNGold.cpp/h
GoalProtectCreature.cpp/h
GoalProtectDungeonTemple.cpp/h

**In source/rooms:**
Room.cpp/h
RoomDojo.cpp/h
RoomDungeonTemple.cpp/h
RoomForge.cpp/h
RoomHatchery.cpp/h
RoomLibrary.cpp/h
RoomObject.cpp/h
RoomPortal.cpp/h
RoomQuarters.cpp/h
RoomTreasury.cpp/h

**in source/network:**
ChatMessage.cpp/h
ClientNotification.cpp/h
ODClient.cpp/h
ODPacket.cpp/h
ODServer.cpp/h
ODSocketClient.cpp/h
ODSocketServer.cpp/h
ServerNotification.cpp/h

**In source/modes:** (Everything directly used by application modes)
AbstractApplicationMode.cpp/h
Console.cpp/h
ConsoleMode.cpp/h
Console_getPromptCommand.cpp
Console_getHelp.cpp
EditorMode.cpp/h
FppMode.cpp/h
GameMode.cpp/h
InputManager.cpp/h
MenuMode.cpp/h
MenuModeMultiplayer.cpp/h
MenuModeSingleplayer.cpp/h
ModeManager.cpp/h
PreefixTree.cpp/h

**In source/gamemap:**
GameMap.cpp/h
MapLoader.cpp/h
MiniMap.cpp/h
TileContainer.cpp/h
TileCoordinateMap.cpp/h

**In source/ai:**
[...]

PR #97, opened 2014-08-20
Fix definition of the CMAKE_CXX_FLAGS + linux fixes

akien-mga:
As discussed in issue #42, the CMake buildsystem is broken on some systems because the project definition does not come at the beginning of the CMakeLists.txt. Because of this, CMake does not know that this is a C++ project and the CMAKE_CXX_COMPILER variable is empty, therefore the --std=c++11 flag is not added properly to the CMAKE_CXX_FLAGS.

@tomluchowski seems to say that moving the project definition to the top broke the his buildsystem, but I would be more inclined to think that something is wrong on his end then. Projects should always be defined at the beginning of the CMakeLists.txt IMHO.

This pull request also adds an insignificant change that makes my life easier to compile OD on Mageia: the FIND_SFML_LIB_SUFFIX is now listed as a CMake argument, and the proper include dir is searched for (Mageia packaged SFML2 so that it can be co-installed with SFML, hence some tricks).


hwoarangmy:
I'm gonna test this with windows

hwoarangmy:
It doesn't work for me :(
I need the flag -std=gnu++11 instead of -std=c++11. But when I change that, the whole flag string is re-added :(
EDIT : with MinGW. I didn't try with MSVS

akien-mga:
@hwoarangmy: Can you debug by adding some `message(STATUS "Flags: " ${CMAKE_CXX_FLAGS})` before and after the definition of the flags?

Edit: Does MinGW use gcc too? Then maybe it's just the logic of the whole https://github.com/akien-mageia/OpenDungeons/blob/df197c7673d0e6f009990253e2b2698f1c8e821a/CMakeLists.txt#L51-L67 that should be fine-tuned, to make the distinction between GCC on Unix and MinGW?

I'll try to find some examples of how it's done in other C++11 projects.

Bertram25:
> Edit: Does MinGW use gcc too?

Normally, yes. MingW stands for Minimal gnu environment for Windows IIRC, and it is using a windows version of gcc/g++.
So I'd say, yes, the compilation flags will have to be adapted per compiler just as you suggested.
Following the current CMake script you have, it seems the c++11 flags are only set for gcc, and not for MSVC for instance, and the flag `-std=gnu++11` should be used for gcc. Something like:

``` cmake
if (MSVC)
    set ( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
    # We shouldn't set rubbish in the cmake variables...
    #set ( CMAKE_CXX_FLAGS_DEBUG "/DEBUG_FLAGS_GO_HERE")
    #set ( CMAKE_CXX_FLAGS_RELEASE "/RELEASE_FLAGS_GO_HERE" )
elseif (BORLAND) # Needed??
    #set ( CMAKE_CXX_FLAGS "/GLOBAL_FLAGS_GO_HERE")
    #set ( CMAKE_CXX_FLAGS_DEBUG "/DEBUG_FLAGS_GO_HERE")
    #set ( CMAKE_CXX_FLAGS_RELEASE "/RELEASE_FLAGS_GO_HERE" )
elseif (CMAKE_COMPILER_IS_GNUCXX)
    set ( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=gnu++11 -Wall -Wunused -Wno-deprecated -fno-strict-aliasing" CACHE STRING "compile flags" FORCE)
    list( APPEND CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS} -g -O2")
    set ( CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS} -O3 " )
else()
    # TODO Define a default behaviour?
endif ()
```

What do you think?

akien-mga:
Yes the current CMake script is kind of broken, we can now see it as a result of moving the project definition to its rightful place (@hwoarangmy did not have issues with the wrong flag being defined before, since no flag was defined at all for him).

I think we can drop the elseif (BORLAND) since nobody in the team seems to use it, and it doesn't do anything for now. In the main else() we could add a message telling the user to contact us so that the script can be adapted for his/her setup.

@hwoarangmy, @Bertram25: Do you know what would be the needed flags for MSVC and MinGW respectively? At least the needed flag for C++11 compilation, but maybe also optimisation parameters for the _debug and _release flags?

hwoarangmy:
Well, now that you are looking at cmake, you could also see with @Bertram25 why he has to remove the debug cegui file :
https://github.com/OpenDungeons/OpenDungeons/blob/development/CMakeLists.txt#L386

Until now, I use to remove the comment when I download the sources but it could be good to see why is there a problem since with MinGW, Debug version do not link without it (as expected).

Concerning the flags re-added, I will have a look and tell you what I see.

akien-mga:
> Well, now that you are looking at cmake, you could also see with @Bertram25 why he has to remove the debug cegui file :
> https://github.com/OpenDungeons/OpenDungeons/blob/development/CMakeLists.txt#L386

I guess something is not done properly in the FindCEGUI.cmake module if there was a need to explicitly link the _DEBUG and _RELEASE libraries. In my understanding, one should link CEGUI_LIBRARIES, and the latter would be _either_ the debug _or_ the release library depending on what you have specified as parameters, not both. Or do I understand it wrong?

Bertram25:
> Well, now that you are looking at cmake, you could also see with @Bertram25 why he has to remove the debug cegui file.

A right, I completely forgot about that one, sorry for the annoyance @hwoarangmy 

> In my understanding, one should link CEGUI_LIBRARIES, and the latter would be either the debug or the release library depending on what you have specified as parameters, not both. Or do I understand it wrong?

Your reasoning is sound and I do think the module should be improved. If we can find a better open-source FindCEGUI module, let's use it.
But if not, let's create a custom CEGUI Lib variable and set it to be what's needed according to the release type.
I don't know whether you're doing it, but you should also define a default release type early in the cmake script if not set. Something like:
https://github.com/Bertram25/ValyriaTear/blob/master/src/CMakeLists.txt#L21
And then use its value to define the compiling and linking strategies.

Best regards, :)

akien-mga:
I'm more and more thinking that perhaps the refactoring of the CMake scripts should not be postponed and done now :-) But I would want to do it properly (learning some CMake conventions along the way, not just based on my experience from other CMake scripts that I've read), so that would probably be when I'm on holidays in 2,5 weeks :-D

Until then I'll try to make the current system work for all our platforms.

Bertram25:
Np. There is no rush, even if @hwoarangmy 's pace is incredible. ;)

hwoarangmy:
cmake seems like an interesting tool to learn. I've googled a bit and it seems easy to end up in scripts for finding cegui. That may be worth giving a try.
But before pushing a change in cmake files, we should test as many environments as possible. I can test MinGW, MSVS and Ubuntu. But it would be nice to make sure at least all the devs can still work, including @tomluchowski if he's not on holydays as he seems to have special environments

Bertram25:
> But before pushing a change in cmake files, we should test as many environments as possible. I can test MinGW, MSVS and Ubuntu. But it would be nice to make sure at least all the devs can still work, including @tomluchowski if he's not on holydays as he seems to have special environments

I can test on MSVC 2012 and Debian sid. And I will for each steps. (This one being broken on MSVC for the reasons given above.)

Seems like @tomluchowski (aka Paul) is answering more usually on forums. I guess we'll have to ask him there once we're all three ok with the result.

> cmake seems like an interesting tool to learn. I've googled a bit and it seems easy to end up in scripts for finding cegui. That may be worth giving a try.

Yep.

Bertram25:
Good places to start:
https://bitbucket.org/cegui/cegui/src/fb162cdfee768ca5d5526fabe5156309c2f85f72/cmake/?at=default
https://bitbucket.org/cegui/cegui/pull-request/1/fixed-issue-when-some-cegui-paths-wouldnt/diff

Seems the CEGUI team is working on the cmake script those times. By seeing their script, it seems much better than our own, btw. I'm sure we should try to use it.
Note that @hwoarangmy and @MCMic made a few changes that might need to be integrated:
https://github.com/OpenDungeons/OpenDungeons/commit/351c7192318bca5a1b9ffbdf508e0b4ecb85bcdf

https://github.com/OpenDungeons/OpenDungeons/commit/e2f9ee00e3ea85988cdf1a4e430c6a9db1f6336c

akien-mga:
I made some changes to remove unused stuff and define the (hopefully) proper C++11 flag on MinGW. I used UNIX for the "-std=c++11" flag, since I understood it was the correct flag for both gcc 4.8+ and Clang. Time will tell.

WDYT about 6501258?

It's a dirty fix, for now I'm more aiming at having some thing "that works" (i.e. slighter better than the current state), but I will have a go at rewriting/improving the CMakeLists.txt and some modules in September.

hwoarangmy:
Sorry, I had not seen the update. I will check ASAP and tell you

hwoarangmy:
I tried the new patch. Now, CMAKE_CXX_FLAGS is not set at all. The strange thing is that I tried to add "CACHE STRING" and it is the same. But with this attribute, if I rename the flag to CMAKE_CXX_FLAGS2, I see the values in the new flag. But with CMAKE_CXX_FLAGS, it doesn't work. If you want me to do some tests, feel fre to ask. I will have a look on my side as I know it is hard to debug problems you don't see :)

akien-mga:
So I guess my `if(MINGW)` is not a valid instruction then :-)

@hwoarangmy Could you try with something along those lines?

```
if(MSVC)
  # C++11 compilation flags is activated by default
  # Optimisation flags can be added if deemed useful
elseif(CMAKE_COMPILER_IS_GNUCXX)
  if(WIN32)  # MinGW
    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=gnu++11 ${OPTFLAGS}")
  elseif(UNIX)  # GCC and Clang
    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 ${OPTFLAGS}")
  endif()
endif()
```

If it doesn't work, maybe try to debug it with the addition of some status messages with:
`message(STATUS "CXX flags: ${CMAKE_CXX_FLAGS}")`

hwoarangmy:
No, no, your construction is correct. I have added messages at each new part of the cmakefile to see where the flag is erased. The funny thing is that it is not. Each of my messages shows the good flags. But the gui do not show them. And they are not in cmakecache either.

akien-mga:
@hwoarangmy What if you add:
`set(CMAKE_CXX_FLAGS "" CACHE STRING "C++ compilation flags")`
before the `if(MSVC)`?

akien-mga:
> No, no, your construction is correct. I have added messages at each new part of the cmakefile to see where the flag is erased. The funny thing is that it is not. Each of my messages shows the good flags. But the gui do not show them. And they are not in cmakecache either.

It should build out of the box then, no? I mean, even if you don't see the CMAKE_CXX_FLAGS as a configuration option.

hwoarangmy:
That's what I tried. But I think it just ended up in an endless loop of warnings. I'm gonna have a try and tell you.

EDIT : It worked after all. There are loads of warnings but I could compile and run the game. I will make some more tests tomorrow.

CMakeLists.txt
```diff
@@ -25,24 +48,17 @@ endif()
...
-  set ( CMAKE_CXX_FLAGS_DEBUG "/DEBUG_FLAGS_GO_HERE")
-  set ( CMAKE_CXX_FLAGS_RELEASE  "/RELEASE_FLAGS_GO_HERE" )
-endif ()
 
-if (CMAKE_COMPILER_IS_GNUCXX)
-  set ( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wunused -Wno-deprecated -fno-strict-aliasing" CACHE STRING "compile flags" FORCE)
-  list( APPEND CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS}   -g -O2")
-  set ( CMAKE_CXX_FLAGS_RELEASE  "${CMAKE_CXX_FLAGS}   -O3 " )
-endif ()
+set(OPTFLAGS "-g -O2 -Wall -Wunused -Wno-deprecated -fno-strict-aliasing" CACHE STRING "CXX optimisation flags")
```
hwoarangmy: Why are you using -g for all ?
Same for -o2 ?

akien-mga: I don't have a strong reason for it, I don't really understand compilations flags that well yet.
It's just that those are among the default flags when packaging software in Mageia, so I thought it was good practice to use them for release build. But that can be changed of course :-)

For the reference, here are the "default" optflags in Mageia:
`-O2 -g -pipe -Wformat -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -fstack-protector --param=ssp-buffer-size=4 -fPIC`

akien-mga: @hwoarangmy @Bertram25 If you have experience with gcc flags, I'll be glad to define only those that are relevant for release and debug build respectively.

hwoarangmy: I don't know all compiler flags but the ones I know :
-O2 is an optimisation flag. It should not be used in debug mode as it can optimize and make the debugger go crazy
-g is for adding symbols (when there is a crash, you can have a readable callstack)
-fno-strict-aliasing is an optimisation flag (don't really know what it does)
-Wxxx are for warnings (enable or disable)

Concerning the -g and -O2 flags, they should not be set in the global configuration. CMake will add them himself when you choose the executable type you want to generate (Debug, Release, ReleaseWithDbgInfo - pretty explicit)

Bertram25: Hi guys, :)

> Concerning the -g and -O2 flags, they should not be set in the global configuration. CMake will add them himself when you choose the executable type you want to generate (Debug, Release, ReleaseWithDbgInfo - pretty explicit)

This is true. @akien-mga proposed opt flags are the ones used usually to create a ReleaseWithDebugInfo release type, which the best binary type used on open source to get more precise feedback on crashes, while maintaining correct performance.

That said, this is right, we should first set the default release type, and then set compilation and linking flags based upon this, per compiler & platform.
Thus, I'm also against setting default flags for everything.

That said, @akien-mga I do think you don't have to fix everything at once and if you can setup a correct list of IF() permitting to clearly view the compiler flags set per compiler and release type, we'll be able to optimize them little by little. :)

> EDIT : It worked after all. There are loads of warnings but I could compile and run the game. I will make some more tests tomorrow.

@hwoarangmy The previous cmake script made it so a lot of compilation warnings when inhibited. That's why we have so much of them now, but this is proving the improvements done by @akien-mga 's work. :)
We should fix those warnings in the code, for the bugs aren't in the CMake script.

Best regards,

hwoarangmy: > The previous cmake script made it so a lot of compilation warnings when inhibited. That's why we have so much of them now, but this is proving the improvements done by

I've not said it was because of akien's patch ;)
But there was so many that I even thought the compiler was crashed. Which was not the case since it compiled.
With the patch (after removing the unwanted flags), compilation seems to work and I definitely prefer showing the warnings as we should try to remove them when we can. The main problem for me is that the flag values are not displayed so they cannot be changed. I've not tried yet to change them to see what happens but it is not normal and it could hide a problem...

Bertram25: > The main problem for me is that the flag values are not displayed so they cannot be changed. I've not tried yet to change them to see what happens but it is not normal and it could hide a problem...

You mean like when using the cmake gui to setup the project? I haven't tested it yet, but i wonder whether there is an option to make cmake show its system variables.

hwoarangmy: > You mean like when using the cmake gui to setup the project

yes. I always use it :)
I suppose some variables like CXX flags are not handled by cache the same way as standard variables...

Bertram25: > I suppose some variables like CXX flags are not handled by cache the same way as standard variables...

Yeah, it is something to be tested, IMHO.
Also, you can add a few things like that in the script, after the corresponding part, when testing to make sure you're on the right track:
`message(STATUS "CXX flags: ${CMAKE_CXX_FLAGS}")`

hwoarangmy: I've already done that yersterday and the flags are correct. Moreover, when I generate the code blocks project and I use it, it works. But cmake shows an empty string....

akien-mga: I'll play around with cmake-gui and ccmake to see how to make CMAKE_CXX_FLAGS a displayed parameter (note that you should be able to override it with the command line cmake though).

I think the CACHE mecanism caters for that, but you seem to infer it does not work as expected, I'll have a look. I also saw some `optional( )` instructions in some cmake files, I'll read a bit to know what are the proper uses of these two things.

Bertram25: I might have lost track of progress or need of testing here. Feel free to remind me when you need help or test about anything here. :)

akien-mga: It's on hold for now because I don't have much time these days (I have an internship report to finish by the end of the week). I'm on holidays afterwards, so I'll have some time to look at it more thoroughly.

I just haven't tried to change anything for now so that the inline comments are not lost, there is some valuable information for my next attempt :-)

Bertram25: Ok. :)  (I meant not pressure, anyway. I swear! ;])  Good luck for your report. :)

akien-mga:
Ok, back on this.
@Bertram25 Do you know if the above comments will be lost if I rebase my branch to do a better implementation? If so I'll try to save the relevant bits in the corresponding issues.

Bertram25:
> @Bertram25 Do you know if the above comments will be lost if I rebase my branch to do a better implementation? If so I'll try to save the relevant bits in the corresponding issues.

Github changes a lot. Last time I rebased, the comments stayed telling they were about an outdated diff. But who knows? Please try to gather the relevant bits in a commit-independent comment before that as my instinct tells me to be careful. :S 
And many thanks for taking care of this \o/

akien-mga:
Relevant bits about the optimization flags in case the previous in-line comments are lost when rebasing:

From @hwoarangmy:

> I don't know all compiler flags but the ones I know :
> -O2 is an optimisation flag. It should not be used in debug mode as it can optimize and make the debugger go crazy
> -g is for adding symbols (when there is a crash, you can have a readable callstack)
> -fno-strict-aliasing is an optimisation flag (don't really know what it does)
> -Wxxx are for warnings (enable or disable)
> 
> Concerning the -g and -O2 flags, they should not be set in the global configuration. CMake will add them himself when you choose the executable type you want to generate (Debug, Release, ReleaseWithDbgInfo - pretty explicit)

From @Bertram25:

> > Concerning the -g and -O2 flags, they should not be set in the global configuration. CMake will add them himself when you choose the executable type you want to generate (Debug, Release, ReleaseWithDbgInfo - pretty explicit)
> 
> This is true. @akien-mga proposed opt flags are the ones used usually to create a ReleaseWithDebugInfo release type, which the best binary type used on open source to get more precise feedback on crashes, while maintaining correct performance.
> 
> That said, this is right, we should first set the default release type, and then set compilation and linking flags based upon this, per compiler & platform.
> Thus, I'm also against setting default flags for everything.
> 
> That said, @akien-mga I do think you don't have to fix everything at once and if you can setup a correct list of IF() permitting to clearly view the compiler flags set per compiler and release type, we'll be able to optimize them little by little. :)
> 
> > EDIT : It worked after all. There are loads of warnings but I could compile and run the game. I will make some more tests tomorrow.
> 
> @hwoarangmy The previous cmake script made it so a lot of compilation warnings when inhibited. That's why we have so much of them now, but this is proving the improvements done by @akien-mga 's work. :)
> We should fix those warnings in the code, for the bugs aren't in the CMake script.

akien-mga:
I've had a look at how to display the CMAKE_CXX_FLAGS in cmake-gui or ccmake, but indeed it seems to be handled differently from other variables. I saw reports of the use of "CACHE FORCE", but it doesn't work either it seems.

The CMAKE_CXX_FLAGS can be seen in cmake-gui and ccmake when toggling "Advanced" variables. Because of the way those graphical tools work though, it seems to display only the status of the variable _before_ the CMakeLists.txt is computed, i.e. the user-defined CMAKE_CXX_FLAGS.

So if you run `cmake -DCMAKE_CXX_FLAGS="-g -O2"` or define those in cmake-gui, you will only see "-g -O2" in cmake-gui, but the "-std=c++11" or "-std=gnu++11" will still be appended when configuring and used when building.

In my next commit, I'll add a simple workaround to show the actual values of CMAKE_CXX_FLAGS when configuring with `message(STATUS "CMake CXX Flags: " ${CMAKE_CXX_FLAGS})`.

So the logic is: use cmake-gui or -DCMAKE_CXX_FLAGS to define your _custom_ compilation and optimisation flags. Then our CMakeLists.txt script appends the _mandatory_ flags to the user-defined ones.

I've removed the hardcoded optimisation flags for now. As @Bertram25 said, some flags are automatically added depending on the CMAKE_BUILD_TYPE, so I don't think we need to add more flags for now. Let's improve this later on if we want e.g. to see more warnings.

Bertram25:
> So the logic is: use cmake-gui or -DCMAKE_CXX_FLAGS to define your custom compilation and optimisation flags. Then our CMakeLists.txt script appends the mandatory flags to the user-defined ones.
> 
> I've removed the hardcoded optimisation flags for now. As @Bertram25 said, some flags are automatically added depending on the CMAKE_BUILD_TYPE, so I don't think we need to add more flags for now. Let's improve this later on if we want e.g. to see more warnings.

All fine to me. I do no more than that for VT AFAIK. :)

akien-mga:
Actually the value of the "final" CMAKE_CXX_FLAGS can be displayed in cmake-gui if I use `set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11" CACHE STRING "Blabla" FORCE)`
The issue with this is that "-std=c++11" is appended to the flags every time you configure, so you end up with useless duplicates at the end of the CMAKE_CXX_FLAGS string.

akien-mga:
I'll also throw in a commit to make it more probable that CEGUI 0.8.x is used instead of CEGUI 0.7.x when both are installed on the system (basically the cmake module should look for "cegui-0" before looking for "cegui", since the latter would be CEGUI 0.7.x).

And I'll remove FindCEGUIOGRE.cmake since we do its job in FindCEGUI.cmake. Either way the whole CEGUI detection will have to be improved, but as pointed out earlier, upstream is working on an official FindCEGUI.cmake script.

Edit: Actually this won't fix the issue when having both CEGUI 0.7.x and CEGUI 0.8.x: it makes it possible to build against the correct lib, but for some obscure reason, the game crashes if libCEGUIBase.so (CEGUI 0.7.x) is installed, even though `ldd ./OpenDungeons.bin` does show that it is linked against libCEGUIBase-0.so (CEGUI 0.8.x). That's quite strange, but whatever, we don't particularly support having two concurrent versions of CEGUI's development libraries installed on the system (I only happened to have it once because Super Maryo Chronicles hasn't been ported to CEGUI 0.8.x yet).

akien-mga:
@hwoarangmy @Bertram25 Can you test this one? I tried to keep it as simple as possible to have something "that works", we can fine tune it afterwards (or rework the whole CMakeLists.txt and cmake modules).

Bertram25:
Thanks, I'll test asap. :)

While I see it, here, it is said something that may be useful for you:
http://www.cmake.org/pipermail/cmake/2007-April/013479.html

`Note that CMAKE_BUILD_TYPE is not initialized with a readable value at configuration time. 
This is because the user is free to select a build type at build time.`

> https://github.com/akien-mga/OpenDungeons/blob/4f63e2235c35a66ee3b68a4517560e6053b86ed4/CMakeLists.txt#L45

Once this works, my next request will be to add debug flags to the debug build, and -Wall -Wextra when using gcc/clang/MingW. (explaining why the build type must be set before compiler flags.)
I'll also request that the default build type is RelWithDebInfo

Once all of this is done, I'll stop bothering you. ;)

Best regards,

akien-mga:
> Once this works, my next request will be to add debug flags to the debug build, and -Wall -Wextra when using gcc/clang/MingW. (explaining why the build type must be set before compiler flags.)

Which debug flags would you want added to the debug build? Do you want -Wall -Wextra to all gcc/clang/MinGW builds, or only debug?

If I understand correctly, the compilation flags that actually get used are the CMAKE_CXX_FLAGS + CMAKE_CXX_FLAGS_<build type>, with the latter having the following default values:

> -- CMake CXX Flags - Release: -O3 -DNDEBUG
> -- CMake CXX Flags - RelWithDebInfo: -O2 -g -DNDEBUG
> -- CMake CXX Flags - Debug: -g
> -- CMake CXX Flags - MinSizeRel: -Os -DNDEBUG

akien-mga:
Simply adding `set(CMAKE_CXX_FLAGS_DEBUG "-g -Wall -Wextra")` for the relevant platforms seems to work fine:

> [  1%] Building CXX object CMakeFiles/OpenDungeons.bin.dir/source/AIFactory.cpp.o
> /usr/bin/c++   -DOD_DEBUG **-std=c++11 -g -Wall -Wextra** -I/home/akien/Projects/opendungeons/OpenDungeons-akien/source -I/home/akien/Projects/opendungeons/OpenDungeons-akien/dependencies -I/home/akien/Projects/opendungeons/OpenDungeons-akien/dependencies/tinygettext -I/home/akien/Projects/opendungeons/OpenDungeons-akien/dependencies/tinyxml2 -I/home/akien/Projects/opendungeons/OpenDungeons-akien/dependencies/angelscript/angelscript/include -I/home/akien/Projects/opendungeons/OpenDungeons-akien/dependencies/angelscript/angelscript/source -I/home/akien/Projects/opendungeons/OpenDungeons-akien/dependencies/angelscript/add_on/scriptarray -I/home/akien/Projects/opendungeons/OpenDungeons-akien/dependencies/angelscript/add_on/scriptbuilder -I/home/akien/Projects/opendungeons/OpenDungeons-akien/dependencies/angelscript/add_on/scripthelper -I/home/akien/Projects/opendungeons/OpenDungeons-akien/dependencies/angelscript/add_on/scriptstdstring -I/usr/include/cegui-0 -I/usr/include/SFML2 -I/usr/include/OGRE -I/usr/include/AL -I/usr/include/OIS    -o CMakeFiles/OpenDungeons.bin.dir/source/AIFactory.cpp.o -c /home/akien/Projects/opendungeons/OpenDungeons-akien/source/AIFactory.cpp

Bertram25:
> Which debug flags would you want added to the debug build?

As per compiler flag, I think nothing more. But in debug builds, we will want to link against debug libraries. That's what my twisted mind was thinking.

>  Do you want -Wall -Wextra to all gcc/clang/MinGW builds, or only debug?

For all. Whatever the build type is, we should never inhibit common warnings. The build should be warning free, at least on this platform, to me.

> Simply adding set(CMAKE_CXX_FLAGS_DEBUG "-g -Wall -Wextra") for the relevant platforms seems to work fine:

Cool :)

Bertram25:
Works fine on Debian here. A test on windows is left to be done.

Bertram25:
@akien-mga Do you want to merge that first, once the windows test is done, or do you prefer adding some other things?

regards :)

akien-mga:
@Bertram25 I'll just do a commit to add -Wall -Wextra on all builds.

I was thinking that I could add it in a way that makes it possible to override, e.g.

```
IF(NOT DEFINED CMAKE_CXX_FLAGS)
  SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra")
END()
```

Or I can add a WARNING=ON option that adds these flags?

akien-mga:
@Bertram25 Done the simple way for now, see cfca5d0.

Bertram25:
Thanks for working on this. :)
Unfortunately, on Debian CMAKE_CXX_FLAGS is considered defined even if empty. (I made sure to delete the CMakeCache.txt and CMakeFiles/ folder before trying.)
Hence, everything is working except setting the -Wall and -Wextra flags.
Is it working on your side maybe?

Does someone mind if we force the warnings for gcc? I'd like to see something like this added, for instance:

``` cmake
IF (CMAKE_COMPILER_IS_GNUCXX)
    # Help getting compilation warnings
    SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra")
    IF (WIN32)
        # This includes enough debug information to get something useful
        # from Dr. Mingw while keeping binary size down. Almost useless
        # with gdb, though.
        SET(CMAKE_CXX_FLAGS_RELWITHDEBINFO "-O2 -ggdb0 -gstabs2")
    ENDIF()
ENDIF()
```

akien-mga:
I'll admit I hadn't tested, I thought it would work :-)

I'll implement your code snippet instead.

akien-mga:
@Bertram25 I changed the previous commit for c473fd8.

The -Wall -Wextra flags are defined by default here on Mageia.

hwoarangmy:
Sorry, I couldn't test that this week end. Do you want me to test now or are you doing something ?

akien-mga:
@hwoarangmy It should be finished now, you can test :-)

Bertram25:
@akien-mga I've tested. It's working on linux. In fact, it's working so well I've got all those annoying warnings coming form Ogre source (Debian ones).

If I remove the -Wextra flag, then I have warnings coming from us, and the situation is much more correct, even if full of warnings.
I know I'm soo annoying, but could you remove the -Wextra flag at least until the Ogre guys have done something with their source?
This will my last request, the rest is ok, as for me. :D

akien-mga:
> I know I'm soo annoying, but could you remove the -Wextra flag at least until the Ogre guys have done something with their source?

Done in de2a6e6

Bertram25:
@akien-mga Ok, I'm fine with the change under linux. I don't know whether I'll have time to test under windows tonight though.

akien-mga:
There's no hurry :-)

hwoarangmy:
I've tried with windows. It works but compilation took 50 mins where it usually takes 10 minutes without all the warnings. Since AFAIK, removing the warnings is not a top priority, I think we should let the extra warnings as an option.

Concerning the gnu++11 flag, I let CXX_FLAGS empty and it worked (I've seen that they were set in cmake logs).

Bertram25:
> Since AFAIK, removing the warnings is not a top priority, I think we should let the extra warnings as an option.

Ok, if it is a problem, let's make it an option:
Something like this?

``` cmake
# At the beginning, near the project statement
OPTION(ENABLE_WARNINGS "Compile the game with all standard warnings enabled" OFF)

# ...

# Then when setting the cmake flags
IF(ENABLE_WARNINGS)
    # Set the cmake flags here
ENDIF(ENABLE_WARNINGS)
```

To invoke this, AFAIK:
`cmake -DENABLE_WARNINGS=on .`

akien-mga:
Done in 08c5518 :-)

Bertram25:
Perfect! Mergeable as for me. :) ok for you @hwoarangmy?

hwoarangmy:
Sorry, I can't really test at this moment. But if the warnings are optional, it's good for me :)

Bertram25:
@hwoarangmy Np. Take the time needed. If I can test on windows on MSVC before that, I'll report here as well.

Btw, once this PR will be merged, I'd really like to bite #10 once and for all. I know it's not urgent, but it's the kind of things that should be done rather sooner than later IMO, and I already did half the work by making the proposal anyway. I know this is quite some change according to files location.
Would you guys be ok with it?

hwoarangmy:
No problem for me. You can change source files as you please.

Bertram25:
ok, thanks. :)

akien-mga:
+1 from me, your proposal in #10 is quite good, and hopefully this should be an easy change. We'll have to merge it quickly to not hinder @hwoarangmy in his incredible coding pace with merging issues :-P

Bertram25:
> +1 from me, your proposal in #10 is quite good, and hopefully this should be an easy change. We'll have to merge it quickly to not hinder @hwoarangmy in his incredible coding pace with merging issues :-P

So true! :D

Bertram25:
@hwoarangmy I realize you said it was good for you. Does this mean I can merge?

hwoarangmy:
I have not tested since the warning options but if you tested, yes, you can merge. @akien-mga , when we change the flags in cmake gui, will it be displayed ?
If not, if you want, I can have a look at how flags are handled in SFML, for example, as I have a compiled version on my computer.

akien-mga:
I think the ENABLE_WARNINGS option should be displayed in cmake-gui. If you defined CMAKE_CXX_FLAGS in cmake-gui, they should be displayed in the terminal output together with additional flags added by our CMakeLists.txt script (i.e. warning flags if enabled and the C++11 flag).

hwoarangmy:
Ok. But they will not be displayed in the CMAKE_CXX_FLAGS editbox, right ? I remember that with your original patch, when I used to change CMAKE_CXX_FLAGS, my changes were added to the flags you add by default. For example, you used to add c++11 flag, when I tried to change for gnu++11, I ended up with both in the editbox.

Don't know if it will help much but as CMAKE_CXX_FLAGS seems to be handled weirdly by cmake, we may also want to not modify it at all and instead use checkboxes/combos/editboxes that will append to CMAKE_CXX_FLAGS the wanted flags.
In our case, we will have a checkbox for warnings (already done AFAIK) and a checkbox to choose the c++11 flag wanted (none or -std=gnu++11 or -std=c++11).

What do you think ?

akien-mga:
That's not a bad idea :-)

Someting like:

```
set(OPT_FLAGS "" CACHE STRING "Optimization and optional compilation flags")
set(CXX11_FLAGS "" CACHE STRING "Compilation flags to enable C++11 support")

if(MSVC)
    # Nothing for now
else()
    if(ENABLE_WARNINGS) 
        set(OPT_FLAGS "${OPT_FLAGS} -Wall"
    endif()
    if(MINGW)
        if(NOT DEFINED CXX11_FLAGS)
            set(CXX11_FLAGS "-std=gnu++11")
        endif()
        # This includes enough debug information to get something useful
        # from Dr. Mingw while keeping binary size down. Almost useless
        # with gdb, though.
        set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "-O2 -ggdb0 -gstabs2")
    elseif(UNIX and NOT DEFINED CXX11_FLAGS)
        set(CXX11_FLAGS "-std=c++11")
    endif()
endif()

set(CMAKE_CXX_FLAGS "${CXX11_FLAGS} ${OPT_FLAGS} ${CMAKE_CXX_FLAGS}")
```

The flags defined automatically might then still be overridden using CMAKE_CXX_FLAGS since it comes last in its own definition.

hwoarangmy:
Yes. Not allowing to change CMAKE_CXX_FLAGS, even if it is because of its weird handling, seems too much like a hack IMHO.
I think it is better like that.
With your change, if I check warnings, for example, and I add something in CMAKE_CXX_FLAGS, will the warning flag be appended or overriden ?

akien-mga:
> With your change, if I check warnings, for example, and I add something in CMAKE_CXX_FLAGS, will the warning flag be appended or overriden ?

If you check warnings, define e.g. CMAKE_CXX_FLAGS="-g -O3", and you are on MinGW with a RelWithDebInfo build, you should get in the end for CMAKE_CXX_FLAGS:
`-std=gnu++11 -Wall -g -O3 -O2 -ggdb0 -gstabs2`
i.e. `${CXX11_FLAGS} ${OPT_FLAGS} ${CMAKE_CXX_FLAGS} ${CMAKE_CXX_FLAGS_RELWITHDEBINFO}`

This means that ultimately the user-defined CMAKE_CXX_FLAGS flags are appended and can override any previous definition apart from the build dependent defition (e.g. here with -O2 from the RelWithDebInfo definition which overrides -O3).

This is still somewhat messy. I should have a look at some other projects to see how they manage this.

akien-mga:
> Yes. Not allowing to change CMAKE_CXX_FLAGS, even if it is because of its weird handling, seems too much like a hack IMHO.

Note that in the currently committed implementation, you can change the CMAKE_CXX_FLAGS, apart from the C++11 flag.

hwoarangmy:
Ok. I will try this evening what happens if I change CMAKE_CXX_FLAGS :)

akien-mga:
With the new proposal, the idea would be that the user defines:
- CXX11_FLAGS if he wants to override the default value (which should already be correct though), though I would have to make sure the redefinitions based on the compiler are not done if the CXX11_FLAGS has already be defined (thus putting them in `if(NOT DEFINED CXX11_FLAGS)`.
- OPT_FLAGS for any other optional C++ flag; the advantage being that this option is visible in cmake-gui contrarily to CMAKE_CXX_FLAGS. Apart from that, defining either OPT_FLAGS or CMAKE_CXX_FLAGS should have the same effect.

The way they are defined, the build-specific flags (CMAKE_CXX_FLAGS_RELWITHDEBINFO) cannot be overridden, but I think that's the way every cmake project works (I've seen some projects undefine all those build-specific flags to have full control over their flags regardless of the build type).

hwoarangmy:
I've tried your patch. It's much better without the warnings :)
Concerning the warning checkbox, could you rename it to CMAKE_CXX_ENABLE_WARNINGS so that it is displayed with the flags ?
Concerning CMAKE_CXX_FLAGS, in the gui, I've tried to set it to "-std=c++11". When I do that, I see in the logs that it is set to "-std=c++11 -std=gnu++11". I know if I let it empty, it works but we can imagine that we will, one day, add some flag (for mac for example). That would not be possible with the patch as it is now.
I think the most simple would be as we said to have a combo with c++11 setting and do not have anything in CMAKE_CXX_FLAGS. This way, there will be no problem :)

akien-mga:
> Concerning the warning checkbox, could you rename it to CMAKE_CXX_ENABLE_WARNINGS so that it is displayed with the flags ?

I'm not sure about it, because it would like an internal CMake variable. Maybe I can rename it to OD_ENABLE_WARNINGS?

> Concerning CMAKE_CXX_FLAGS, in the gui, I've tried to set it to "-std=c++11". When I do that, I see in the logs that it is set to "-std=c++11 -std=gnu++11". I know if I let it empty, it works but we can imagine that we will, one day, add some flag (for mac for example). That would not be possible with the patch as it is now.
> I think the most simple would be as we said to have a combo with c++11 setting and do not have anything in CMAKE_CXX_FLAGS. This way, there will be no problem :)

Yep I'll do that, based on the snippet I posted above.

akien-mga:
@hwoarangmy, @Bertram25 How is it with my new changes in 998c9c9?

Bertram25:
The file is looking good. I'm currently testing... :)

hwoarangmy:
I've tried but I see OD_CXX11_FLAGS empty. I was expecting -std=gnu++11 (in the log, I see -std=gnu++11 so it seems that it is just not displayed). I think it would be better to display it so that if one wants to use some other exotic flag, he can remove the one by default and replace by what he wants.

Bertram25:
> I've tried but I see OD_CXX11_FLAGS empty. I was expecting -std=gnu++11 (in the log, I see -std=gnu++11 so it seems that it is just not displayed). I think it would be better to display it so that if one wants to use some other exotic flag, he can remove the one bu default and replace by what he wants.

You mean before or after configuring? Is there a way to do that before the configuration phase without cache, btw?

Bertram25:
Working fine as for me.

akien-mga:
> I've tried but I see OD_CXX11_FLAGS empty. I was expecting -std=gnu++11 (in the log, I see -std=gnu++11 so it seems that it is just not displayed). I think it would be better to display it so that if one wants to use some other exotic flag, he can remove the one by default and replace by what he wants.

It's somewhat of a pain to get variables to output their cached values in the gui, so I guess that what you would like to have is not exactly in the scope of the CMake design. Still, I managed to achieve this result with the "FORCE" parameter in 028b9b2. It can only be used for absolute variable definitions though, since it's executed again each time you configure a project (thus I can't use it for OD_OPT_FLAGS, because it would append -Wall at each cmake call, and you end up with "-Wall -Wall -Wall -Wall" etc.).

hwoarangmy:
I guess you think I'm going to complain again about something. But no, this time, the old bear is not going to bite :)
That's good for me. I've tried to change the flag and it works in the output. Good work and sorry for all the complaints.

Bertram25:
Let's merge then? (Fine to me as well.)

akien-mga:
> That's good for me. I've tried to change the flag and it works in the output. Good work and sorry for all the complaints.

No problem, what's important is to have something efficient and user-friendly in the end :-)

Bertram25:
@akien-mga Congrats for the long standing effort on this. ;)

PR #99, opened 2014-08-20
Workers drop gold after digging - Creatures should get beds when dropped in dormitory

hwoarangmy:


akien-mga:
Some incredible work as always, that keeps me from going to sleep :-p
I'm building your branch, I will try to see if the announced changes are effective, as always without obvious regressions.

hwoarangmy:
You are right, I forgot to change Room::dojo and the according variables. I will do that later.
Concerning materials, I won't change that as I belive theses are referenced by the mesh files.

Bertram25:
> Concerning materials, I won't change that as I belive theses are referenced by the mesh files.

I confirm. Please, let only @Danimal696 touch the material inner & file names if possible. Thanks :)

hwoarangmy:
I've updated the pull request for the remaining dojos. Now there is still :
- credits : I don't know what to do
- OpenDungeons.cbp : this should be removed from repository IMHO
- materials : cannot be changed as they are referenced by the Mesh files

Concerning digging, my patch should not have changed how the workers get gold when they end up full. The only change was to make them deposit gold when there is nothing left to dig and they carry gold. The difference you see may come from the patch that shows all the pending actions for units (but has nothing to do with digging). As you said in another thread, we already know that there are things to change as we have all seen creatures not moving after a while in game. The patch showing the pending list will help debug such bugs (at least, we will see what's going on).

Bertram25:
Hi :)

> The only change was to make them deposit gold when there is nothing left to dig and they carry gold.

I do think @hwoarangmy is right, @akien-mageia. The change is about making the worker deposit gold when there is nothing left to dig. If there is still some, the case isn't handled by this change, IMHO.

> credits : I don't know what to do

I'll (or we) deal with the credits as a different problem. It could be safely renamed even if the actual solution is to list filename. So don't bother with that obsolete stuff.

> OpenDungeons.cbp : this should be removed from repository IMHO.

Yes, the cbp file can and should be created through cmake. It will ease our maintenance.

The code and the functionality is a good improvement. Nice work @hwoarangmy :)
I've a question left but it shouldn't prevent a merge, IMHO.

Bertram25:
Btw, @hwoarangmy is soo fast, he's used to kill bug 3 at a time. ;)

Issue #91, opened 2014-08-20
Review the code for the main conventions


Bertram25:
The code should have copyright headers, and follow the main style conventions:
...

More severe problems than formatting (like putting variable in the private space will have to dealt with another issue.)

Culprits found so far:
1- Misses header copyright (check pragma once practice): (Needed for release)
- [x] CullingQuad.cpp/h
- [x] DummyArrayClass.cpp/h
- [x] PrefixTree.h/cpp
- [x] SlopeWalk.cpp/h
- [x] SoundEffectsHelper.cpp/h

2- Needs variables renaming. (can be done afterwards.)
- [x] AIManager.cpp.h
- [x] CullingQuad.cpp/h (To be removed along with the culling manager)
- [x] DummyArrayClass.cpp/h
- [x] FastPODAllocator.h (To be removed along with the culling manager)
- [x] GameEntity.cpp/h
- [x] GameMap.ccp/h
- [x] LogManager.cpp/h: sem_t logLockSemaphore;
- [x] MenuModeSkirmish.cpp/h
- [x] MenuModeMultiplayer.cpp/h
- [x] NullAI.h: static AIFactoryRegister<NullAI> reg;
- [x] ODPacket.h/cpp: sf::Packet packet;
- [x] PrefixTree.h/cpp
- [x] RenderRequest.cpp/h
- [x] SlopeWalk.cpp/h (To be removed along with the culling manager)
- [x] SoundEffectsHelper.cpp/h
- [x] StackTracePrint.cpp/h
- [x] Tile.h
- [x] Vector3i.cpp/h (Now useless, will be removed along with the culling manager.)

Used? (Has problem 1. and 2.)
- [x] SocketException.h (was removed.)

hwoarangmy:
> SocketException.h

This one can be safely removed

hwoarangmy:
And I would add that MapLight, Weapon and Battlefield are not game entities. Looking at their code shows that they share almost all the needed to be. Is there a reason why ? To be honest, I have not looked enough to be sure there is no tricky reason but from what I've seen, it seems possible...

Bertram25:
> And I would add that MapLight, Weapon and Battlefield are not game entities. Looking at their code shows that they share almost all the needed to be. Is there a reason why ? 

I'm not certain but I know map lights are shown in the editor mode. Weapon also have meshes to represent them, and battleField values can be represented using visuals if some debug code is uncommented. This may explain it.

[...]

PR #105, opened 2014-08-21
Started dealing with #91

Bertram25:
- Added missing copyright headers. The issue milestone can be set to Later, afterwards.
- Also cleaned up the prefix tree class and fixed the DummyArrayClass variable name.

Part of #91

### DIFF
--- a/source/CullingQuad.cpp
+++ b/source/CullingQuad.cpp
@@ -1,3 +1,19 @@
+/*
+ *  Copyright (C) 2011-2014  OpenDungeons Team
+ *
+ *  This program is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
 
 #include "CullingQuad.h"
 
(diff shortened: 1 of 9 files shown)


akien-mga:
It builds fine (I wanted to make sure since you removed some includes), no obvious regressions.

hwoarangmy:
> It builds fine (I wanted to make sure since you removed some includes), no obvious regressions.

It is always good to try to build the game and launch it. Sometimes, it is hard to see the consequences of a patch. But I believe the less we could expect from a repository, even the development branch, is to compile and launch
Akien, I let you deal with that :)

Bertram25:
> It builds fine (I wanted to make sure since you removed some includes)

Yep, but I only removed includes in the cpp file that were already in the header, in fact.
I also tested the prefix tree on linux by opening the console, type the start of a command listed in:
https://github.com/OpenDungeons/OpenDungeons/blob/development/config/console_commands.txt
and press tab several times to see it completed by the possible commands.

Feel free to merge whenever you think this is ok. :)

akien-mga:
I've proofread the code (that was interesting since I don't know C++, but I did not spot any obvious issue in the little I could understand with general programming languages knowledge). Let's merge :-)

I shall take some time to learn the basics of C++ to be more confident reading OD code. I'm on Lua for now though :-)

Bertram25:
> I shall take some time to learn the basics of C++ to be more confident reading OD code. I'm on Lua for now though :-)

Beware, you're becoming more and more a developer every passing day.

@hwoarangmy ^ What have we done, we might have created a monster. O.O' ;)

akien-mga:
Don't you worry, I'm still _Akien the Archlich, evil torturer of the shallow devs_ (i.e. lead bug hunter in the team) ;-)

Bertram25:
> leader bug hunter in the team

Joke apart, this is very true. I saw that also on VT, btw. ;)

hwoarangmy:
> What have we done, we might have created a monster. O.O' ;)

lol

> lead bug hunter in the team

Yes, and we have to say you do that well :)

> I shall take some time to learn the basics of C++ to be more confident reading OD code

If you already know another language, as it seems to be the case, it should be fast as c++ is the root of most languages...

PR #109, opened 2014-08-22
Speed up creature pickup/drop/marking tiles/build room/build trap

### LINKED ISSUE #103: Speed up server acknowledge when player ask pickup/drop creature or mark tiles

hwoarangmy:
@Bertram25 @akien-mga Please test a little bit more this one as usual because I changed some sensible stuff (see details in the commit). But it seems pretty stable as I let the game play alone during 3 hours and I had no crash (I stopped it turn 12517). I also tried quitting the mission and restarting it and fixed the related bug I've seen (mostly about treasury room)
fixes #103

### DIFF
--- a/source/AIWrapper.cpp
+++ b/source/AIWrapper.cpp
@@ -20,6 +20,8 @@
 #include "Seat.h"
 #include "Player.h"
 #include "GameMap.h"
+#include "ODServer.h"
+#include "ServerNotification.h"
 #include "LogManager.h"
 
 AIWrapper::AIWrapper(GameMap& gameMap, Player& player)
@@ -37,8 +39,28 @@ bool AIWrapper::buildRoom(Room::RoomType newRoomType, int x1, int y1, int x2, in
     int goldRequired = coveredTiles.size() * Room::costPerTile(newRoomType);
     if(gameMap.withdrawFromTreasuries(goldRequired, player.getSeat()))
     {
-        Room* newRoom = Room::createRoom(&gameMap, newRoomType, coveredTiles,
-            player.getSeat()->getColor());
+        int color = player.getSeat()->getColor();
+        Room* newRoom = Room::createRoom(&gameMap, newRoomType, coveredTiles, color);
+        int nbTiles = coveredTiles.size();
+        try
+        {
+            const std::string& name = newRoom->getName();
+            ServerNotification *serverNotification = new ServerNotification(
+                ServerNotification::buildRoom, &player);
+            int intType = static_cast<int32_t>(newRoom->getType());
+            serverNotification->mPacket << name << intType << color << nbTiles;
+            for(std::vector<Tile*>::iterator it = coveredTiles.begin(); it != coveredTiles.end(); ++it)
+            {
+                Tile* tile = *it;
+                serverNotification->mPacket << tile;
+            }
+            ODServer::getSingleton().queueServerNotification(serverNotification);
+        }
+        catch (std::bad_alloc&)
+        {
+            OD_ASSERT_TRUE(false);
+            exit(1);
+        }
         Room::setupRoom(&gameMap, newRoom, &player);
         return true;
     }
--- a/source/BattleField.cpp
+++ b/source/BattleField.cpp
@@ -25,11 +25,10 @@
 
 #include <iostream>
 
-BattleField::BattleField(GameMap* gameMap) :
+BattleField::BattleField(GameMap* gameMap, const std::string& creatureName) :
     mHasMeshes(false)
 {
-    mName = "field_" + Ogre::StringConverter::toString(
-        gameMap->nextUniqueNumberBattlefield());
+    mName = creatureName + "_battlefield";
 }
 
 double BattleField::getTileSecurityLevel(int x, int y)
--- a/source/BattleField.h
+++ b/source/BattleField.h
@@ -67,7 +67,7 @@ typedef std::vector<SecurityTile> FieldType;
 class BattleField
 {
 public:
-    BattleField(GameMap* gameMap);
+    BattleField(GameMap* gameMap, const std::string& creatureName);
 
     ~BattleField()
     {}
--- a/source/ClientNotification.cpp
+++ b/source/ClientNotification.cpp
@@ -23,7 +23,7 @@
 ClientNotification::ClientNotification(ClientNotificationType type):
         mType(type)
 {
-    packet << type;
+    mPacket << type;
 }
 
 std::string ClientNotification::typeString(ClientNotificationType type)
--- a/source/ClientNotification.h
+++ b/source/ClientNotification.h
@@ -52,7 +52,7 @@ public:
     virtual ~ClientNotification()
     {}
 
-    ODPacket packet;
+    ODPacket mPacket;
 
     static std::string typeString(ClientNotificationType type);
 
--- a/source/Console_executePromptCommand.cpp
+++ b/source/Console_executePromptCommand.cpp
@@ -338,7 +338,7 @@ bool Console::executePromptCommand(const std::string& command, std::string argum
                     // Inform any connected clients about the change
                     ServerNotification *serverNotification = new ServerNotification(
                         ServerNotification::setTurnsPerSecond, NULL);
-                    serverNotification->packet << ODApplication::turnsPerSecond;
+                    serverNotification->mPacket << ODApplication::turnsPerSecond;
                     ODServer::getSingleton().queueServerNotification(serverNotification);
                 }
                 catch (std::bad_alloc&)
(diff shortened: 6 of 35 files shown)


hwoarangmy:
BTW, now that creatures are not removed from the gamemap when picked up, they move (idle action) while in hand. I let this behaviour because I thought it was funny. If you prefer, I can make them stop. But, in some far away version, we could have some different animation for creatures in hand :)

akien-mga:
> BTW, now that creatures are not removed from the gamemap when picked up, they move (idle action) while in hand. I let this behaviour because I thought it was funny. If you prefer, I can make them stop. But, in some far away version, we could have some different animation for creatures in hand :)

I like this! We could have animations of creatures kicking and roaring because they're angry that they're being grabbed :-D

akien-mga:
> Speed up creature pickup/drop/marking tiles/build room/build trap

That's a huge improvement! All those actions are now done instantly in singleplayer, I'm looking forward to testing it in LAN :-)

akien-mga:
I got a crash when expanding the existing treasury room on TestSingleplayer.level:

```
During this turn there were 92 calls to GameMap::path().Processing deletion queues on turn 187
Server sends newturn=187
ERROR: Assert failed file /home/akien/Projects/opendungeons/OpenDungeons/source/ODClient.cpp line 559 info : name=Treasury_2
signal 11 (Segmentation fault), address is 0x34 from 0x5c50b2

[bt]: (1) ./OpenDungeons.bin : ODClient::processOneClientSocketMessage()+0x752 [0x5c50b2]
[bt]: (2) ./OpenDungeons.bin : ODClient::processOneClientSocketMessage()+0x752 [0x5c50b2]
[bt]: (3) ./OpenDungeons.bin : ODClient::processClientSocketMessages()+0x20 [0x5dceb0]
[bt]: (4) ./OpenDungeons.bin : ODFrameListener::frameStarted(Ogre::FrameEvent const&)+0x192 [0x5e1f72]
[bt]: (5) /lib64/libOgreMain.so.1.9.0 : Ogre::Root::_fireFrameStarted(Ogre::FrameEvent&)+0x44 [0x7f1b5e2f8f04]
[bt]: (6) /lib64/libOgreMain.so.1.9.0 : Ogre::Root::_fireFrameStarted()+0x21 [0x7f1b5e2fa7e1]
[bt]: (7) /lib64/libOgreMain.so.1.9.0 : Ogre::Root::renderOneFrame()+0x9 [0x7f1b5e2fa8d9]
[bt]: (8) /lib64/libOgreMain.so.1.9.0 : Ogre::Root::startRendering()+0x2d [0x7f1b5e2fa92d]
[bt]: (9) ./OpenDungeons.bin : ODApplication::ODApplication()+0x3fc [0x5c3efc]
[bt]: (10) ./OpenDungeons.bin : main+0x1e [0x4682de]
[bt]: (11) /lib64/libc.so.6 : __libc_start_main+0xf0 [0x7f1b5b1eaf90]
[bt]: (12) ./OpenDungeons.bin() [0x4696f4]
```

It's easy to reproduce: just try to expand the treasury room.

hwoarangmy:
> It's easy to reproduce: just try to expand the treasury room.

Stupid me, I forgot to test after my lastest changes. I see what's happening. Patch incoming soon. Like RoomObjects, I will change room absorption so that the new room absorbs treasury indicators instead of destroying/redoing all.

> I'm looking forward to testing it in LAN :-)

From what I've seen on the last lan, I would guess that should be nice :)

> We could have animations of creatures kicking and roaring because they're angry that they're being grabbed

Yes, that's also what I had in mind :)

> It could have been nice to split this commit into two or three functional commits

The problem is that they were linked. Every change there is the result of what I had to do to make the same logic for every

hwoarangmy:
> Are they still doing their turn logic, animation apart?

When picked up, they are removed from the tile they were on. So they should not fight.Concerning their turn, nothing is done if the creature is not on map (a boolean set when pickup) :
https://github.com/hwoarangmy/OpenDungeons/blob/development/source/Creature.cpp#L539

But that's the main reason why I asked if it was possible to make some tests. I'm sure about the most obvious cases (creature turn and fighting) but there might be something more tricky that I didn't see...

hwoarangmy:
> The question is: Are there cases where we do not want to "force" the name of a creature/trap/room?
> The boolean forceName sounds like a temporary solution for me, unless there are cases where the creature name shouldn't be touched, while the 'name' parameter isn't empty?

Generally speaking, there are 3 cases :
- Entity creation on server (during portal/dungeon temple turn for creature, for example) : the name is automatically set
- Entity creation on client side (when message is received) : the name has to be the same as the server's
- Entity created temporary : used for refreshing the entity (used for tile, creature, ...)

I don't really like the boolean solution either. But I couldn't think of something better. If you have an idea, please tell me. I will also think about it myself

Bertram25:
> I don't really like the boolean solution either. But I couldn't think of something better. If you have an idea, please tell me. I will also think about it myself

Well, to be honest, I wondered whether simply forcing the entity name when the name is not empty would suffice, just like you implicitly did here for instance:
https://github.com/hwoarangmy/OpenDungeons/commit/11b384799a3b3fd0569a78a555843ef32dd14578#diff-ffe96aa9f153c9d8901adcfe3dd1d3dcR173

> But that's the main reason why I asked if it was possible to make some tests. I'm sure about the most obvious cases (creature turn and fighting) but there might be something more tricky that I didn't see...

Np, playing, err I mean testing is here for that. ;)
Seems not a lot of things can happen when they haven't got any position tiles anyway:
https://github.com/hwoarangmy/OpenDungeons/blob/development/source/Creature.cpp#L588

Regards,

hwoarangmy:
> Seems not a lot of things can happen when they haven't got any position tiles anyway

There is something tricky. When you remove the creature from the tile, positionTile can not return NULL as you also need to set an invalid value in position because :
https://github.com/hwoarangmy/OpenDungeons/blob/development/source/Creature.cpp#L2652

hwoarangmy:
From the start, looking at the treasury room, I've been thinking why treasures are not room objects ? Is that because they were static so it was not usefull to have the whole animated objects mecanisms ?

Bertram25:
> From the start, looking at the treasury room, I've been thinking why treasures are not room objects ? Is that because they were static so it was not usefull to have the whole animated objects mecanisms ?

I think yes, and also because it was likely the oldest room coded explaining the code roughness there, maybe.
Note that there is no reason for them not to be room objects  (plus not-technically speaking they are), so they can become ones without any regret, IMHO.

> There is something tricky. When you remove the creature from the tile, positionTile can not return NULL as you also need to set an invalid value in position because :
> https://github.com/hwoarangmy/OpenDungeons/blob/development/source/Creature.cpp#L2652

Ah, of course, silly me. :)
If it happens to be needed, I wonder whether we could readd the boolean telling whether the creature "is on map", without actually removing it from the game map object, to disable a few things and clearly show its state. If bugs happen, I'd suggest we keep this in mind.

Good evening guys :)

hwoarangmy:
> I wonder whether we could readd the boolean telling whether the creature "is on map", without actually removing it from the game map object

It exists (and I used him to stub doTurn when the creature is picked up). I also removed the creature from the tile it is in (from what I've seen, it will avoid enemy creatures to try to attack it). That's all I've seen. But there might be something hidden that I missed. Even as I said, we will have time to debug when the bug comes :) Anyway, it seems more logical to me to let a picked up creature in the gamemap as it is still part of it.

Bertram25:
> It exists (and I used him to stub doTurn when the creature is picked up). I also removed the creature from the tile it is in (from what I've seen, it will avoid enemy creatures to try to attack it). That's all I've seen. But there might be something hidden that I missed. Even as I said, we will have time to debug when the bug comes :)

Ah? Sure, nevermind then. :)

> Anyway, it seems more logical to me to let a picked up creature in the gamemap as it is still part of it.

Agreed. And I'm sure less complicated to handle.

hwoarangmy:
> Ah? Sure, nevermind then. :)

Yep :
https://github.com/hwoarangmy/OpenDungeons/blob/development/source/Creature.h#L337

> Agreed. And I'm sure less complicated to handle.

Yes. I reminded having seen somewhere where we checked for creatures in the gamemap and then in every player hands but I couldn't find it. I thought I had dreamed or something had been removed but I found it. It is in the goal keeping a creature alive. It's much simpler now :)

I made a new commit that uses room objects for treasury indicators. I think the bug is fixed. Please fill free to try :)
The other good thing is that if an artist wants to make effects (like having a gold bag that grows instead of removing the bag and adding a new one), it is possible

akien-mga:
Crash fixed :-)
That's a nice code simplification. I'll let @Bertram25 do the final review before merging.

On a side note, the wall-claiming algorithm seems somewhat strange, creatures claim a wall, then go 10 tiles away to claim another one, then back, then to another location... But I'll open another issue for that (unless it's per design).

Bertram25:
The question on whether we should get rid of the forceName boolean is still left to be seen, but i Think in another commit. :)

I'll merge.

PR #113, opened 2014-08-26
Adding chicken coop in hatchery

### LINKED ISSUE #106: Add the chicken coop in the hatchery

### LINKED ISSUE #112: Change chicken spawning logic in the hatchery
I've been thinking about how to display the chicken coops in the hatchery. I think it could be nice to have one per active spot to know how many chickens are generated even if they are all eaten.
The problem is that for 3x3 rooms, we could have 5 activespots. With the chickens, it will be 10 room objects for 9 tiles.
We could add support to have several room objects on the same tile (at least 2).

@Bertram25 What do you think ?

hwoarangmy:
fixes #112 
fixes #106


akien-mga:
I get this error in the terminal at each turn:

> ERROR: Assert failed file /home/akien/Projects/opendungeons/OpenDungeons/source/RoomHatchery.cpp line 208
> ERROR: Assert failed file /home/akien/Projects/opendungeons/OpenDungeons/source/RoomHatchery.cpp line 262

Apart from that the model loads fine, it looks great :-)

akien-mga:
Tremble before my impressive artistic skills:

od-hatchery-creation

As you can see on the screenshot, there is an issue with the way chicken coop spots are handled when absording rooms: the red square is the area I used first for my hatchery, and it produced two chicken coops (red dots, don't mind the ghost claimed walls). Then I expanded the room to fill the whole space, and the new chicken coop spots were correctly computed and placed, but the previous chicken coops should also be removed IMO.

Same thing at the top of my screenshot, where I created the hatchery in several steps (three or four).

Bertram25:
> ERROR: Assert failed file /home/akien/Projects/opendungeons/OpenDungeons/source/RoomHatchery.cpp line 208
> ERROR: Assert failed file /home/akien/Projects/opendungeons/OpenDungeons/source/RoomHatchery.cpp line 262

Hi @hwoarangmy 
In `void RoomHatchery::removeCoveredTile(Tile* tile)`, (and `void notifyRemoveActivespot()` but that might be normal here), it seems you never remove/erase the tile in `mCreaturesChickens`. Is that desired?
Is a call to `removeCreatureUsingRoom(Creature* c)` missing in that case maybe?

Seems to me, you're almost there :D 

Good luck and best regards,

Bertram25:
> Tremble before my impressive artistic skills:

Oh, god O.O`

hwoarangmy:
> You are making the room code remove any dummies before recreating them here? Am I Right?

No, when there is a room absorbed, its room objects are keeped in the new room. I moved the code in the training room because it is more logical to like that but in reality, there is no impact on the unused dummies when the room is absorbed so it doesn't change anything. But if later, we add code that is impacted by tiles removal (like hatchery), it will be less confusing.

> Then I expanded the room to fill the whole space, and the new chicken coop spots were correctly computed and placed, but the previous chicken coops should also be removed IMO.

You have seen something I did not. I had never thought that expanding a room could change the activespots position (even if it is logical now that I think about it). I need to change that since it is a problem that should also impact all the other rooms.In fact, if the room active spots move, the old ones should be removed. Nice catch.

> it seems you never remove/erase the tile in mCreaturesChickens. Is that desired?

As I writed in the comments (seems like not clear enough), it is the call to changeEatRoom that will remove the creature from the room.

> index should also be unsigned int. This will create a warning once we adopt Akien's work. ;)

Yes, I need to have a look on that. But it will be easier when akien's work will be there. I just hope we will not have too many non-avoidable warnings due to boost.

> You are making the room code remove any dummies before recreating them here?

It is the same as I said previously. Because room objects are absorbed, when there is a room absorbtion, the room objects are added to the new room and removed from the old one. This way, when the tiles are removed from the old room, the room objects are not removed.

EDIT : This bug with room objects handling might also be the reason why there are assertions as room objects are not where they are expected to be.

Bertram25:
> As I writed in the comments (seems like not clear enough), it is the call to changeEatRoom that will remove the creature from the room.

Ah, everything makes sense now. Sorry for lagging behind like a dead trail, but I'm trying to follow the closest possible. ;)

hwoarangmy:
> Ah, everything makes sense now. Sorry for lagging behind like a dead trail, but I'm trying to follow the closest possible. ;)

No problem. I know it is not that easy to follow code you didn't write :)

akien-mga:
> You have seen something I did not. I had never thought that expanding a room could change the activespots position (even if it is logical now that I think about it). I need to change that since it is a problem that should also impact all the other rooms.In fact, if the room active spots move, the old ones should be removed. Nice catch.

Do you want to cater for this issue in this pull request, or should we open another issue for that (since you say it affects all rooms)?

hwoarangmy:
I will do that here. It is a buggy behaviour that may lead to crashes

hwoarangmy:
Here is the patch :)
@Bertram25 @akien-mga , can you have a look ?

EDIT : you can play as much as you want merging rooms. There should be no more bugs like in the snapshot :)

Bertram25:
Nice patch! :D 

Nothing to say about the code, and seeing the chickens pick the ground is so lovely. :)
IMHO, looks good to be merged.
I couldn't trigger any segfault, while playing with every rooms.
I'll wait on @akien-mga approval, too.

Great job!

hwoarangmy:
> I couldn't trigger any segfault, while playing with every rooms.

And the good thing is that I've added logs (that I removed before the pull request) to display the active spots. Now, they are the same when a room is created from scratch or when it is merged. That's the reason why there was so many chicken coops in @akien-mga 's snapshot.

akien-mga:
I'll have to work a bit on my linking issue with CEGUI 0.8.4, so I can't test the new patch for now. But based on @Bertram25's comment, I think this is good enough to be merged.

PR #119, opened 2014-08-29
Removed ODFrameListener from Goals. Check if name is used when new entity is created

hwoarangmy:


hwoarangmy:
yes, I will fix the missing while in creature (stupid me) and differentiate more the getGameMap

hwoarangmy:
Concerning AS, I don't know how it works well enough to patch that. AFAIK, it is not used yet, right ?
Do you know how it works, in particular concerning threading ?

akien-mga:
It builds and runs fine.
As far as I can tell the only way to really test these changes in-game is to let a the oldTest run in background and see that it doesn't segfault when new creatures are generated.

hwoarangmy:
> As far as I can tell the only way to really test these changes in-game is to let a the oldTest run in background and see that it doesn't segfault when new creatures are generated.

The problem is that it would segfault only if the good creature spaws at the right moment. For example, If I remember well, there was a wizard7. So if the 7th creature is a wizard and it is called Wizard8, it works. But if you really want to test, I guess the best way would be to edit the level game and add creatures with matchable names. Like having 10 wizards from wizard1 to wizard10 and hope a wizard pops soon :)

akien-mga:
Sounds like a good testcase, I'll try that :-)

akien-mga:
Ok it works fine.
I tried a modified TestSingleplayerSmall with only one wizard named Wizard3, and the possibility to spawn only wizards for a given player. All subsequent wizards are named Wizard(n+1) where n was the number of the last spawned creature (regardless of its class).

hwoarangmy:
ok, cool. Let's wait for @Bertram25 if he can see another of my mistakes and then, we can push :)

akien-mga:
Note that there seems to be conflicts to resolve before the pull request can be pushed.

hwoarangmy:
no pb. When I have your go, I will resolve conflicts and merge

Bertram25:
> no pb. When I have your go, I will resolve conflicts and merge

You can go ahead. It's working nice here. :)

hwoarangmy:
ok, I will push tonight :)

hwoarangmy:
It's done. But it's strange, there was no conflict after all...

Bertram25:
Can happen, especially with intertwined merge points, AFAIK. A good thing done. :)

PR #123, opened 2014-09-01
Add console command to add gold to players and support to add gold in a ...

### LINKED ISSUE #117: Add a console command to give gold to a player
This is for debug purposes but it can be used to change starting gold in level files. The logic should be too dispatch through the treasury rooms the gold. If more gold than treasury rooms available, the gold is dropped.

hwoarangmy:
...level file.

If treasuries are full, remaining gold is dropped
fixes #117

Added map "Stonekeep siege.level" from @Danimal696 

To test it :
- Check  that the gold amount in a level file is added to the treasuries
- In the console, use addgold command to add gold in you treasuries


akien-mga:
Nice work!

The initial gold amount seems to be working, since I started for example the TestSingleplayerSmall maps with 1000 gold.
It seems the addgold command doesn't work here though (neither `addgold` nor `addgold 1000`). But maybe there is some cheat code to type in the console before such commands are available?

The Stonekeep siege level works fine and looks great! :-)
It would have been nice to add it in a separate commit though, since it's not related to the addgold feature. This way if we wanted to revert the addgold commit, we wouldn't delete the level.
But we won't revert it so it's alright :-P

hwoarangmy:
> It seems the addgold command doesn't work here though

To open the console, you need to press ².
If you want a list of available commands, you can enter "help".
To use addgold, you need to give the player to whom give the gold (from color). In TestSingleplayerSmall, for the human player, it would be "addgold 1 1000"

> It would have been nice to add it in a separate commit though

That's right. But in fact, it is not so unrelated since I had to modify all maps to add the starting gold ;)
But I agree it would have been better to do 2 commits.

akien-mga:
> To open the console, you need to press ².
> If you want a list of available commands, you can enter "help".
> To use addgold, you need to give the player to whom give the gold (from color). In TestSingleplayerSmall, for the human player, it would be "addgold 1 1000"

Here it is F12 to open the console :-)

I've used addgold with its two mandatory arguments, it works fine.

I've noticed a few issues with the console as a whole though. I'll probably open an issue for them tomorrow.
- Long lines are not cropped, e.g. if you type "help addgold", the definition is on a single line (and thus you can't read it completely before it reaches the end of the screen).
- After I've issued "help", it seems every command that I write also runs "help" again (so if I run "addgold 1 1000", I get my 1000 gold but I also see the help once again.

Bertram25:
Fine. :) I'll merge.

Issue #120, opened 2014-09-01
Crash on missile spawn.


Bertram25:
As reported by @Danimal696 here: http://forum.freegamedev.net/viewtopic.php?f=37&t=5786

The crash exception log is:
`Internal Ogre 3d exception: SceneNode "MissileObject_Missile_Object_1_node" not found. ...`
This happens on the 2014-08-20 rolling release windows binary.

Some work has been done on units names so this might already be fixed in master, but we have to make sure.

hwoarangmy:
I think it is already fixed but I am waiting for @Danimal696 to add more information on how to reproduce the bug.

Bertram25:
> I think it is already fixed but I am waiting for @Danimal696 to add more information on how to reproduce the bug.

I think it too. AFAIK, to reproduce the bug, you have to place cannons within a level file (so they are loaded at map start) such as in the oldtest.level file, and attract enemy creatures to make the cannon fire.
The crash should happen at that moment, or no more. :)
@Danimal696 may have more info, though.

hwoarangmy:
It only happens on canons loaded from file ? I've never tried such cases...

Bertram25:
> It only happens on canons loaded from file ?

I don't know yet, but I reproduced only that way. And I didn't want to interfere or tell things I haven't actually tested.

hwoarangmy:
Well, I have already tested quite a lot canons and didn't see this crash. I remember having fixed a bug where the missile object number was wrongly handled. But I don't remember if it was after or before the new release...

akien-mga:
The release was on August 21st around 1 a.m., you can check if your commit was after that :-)

Danimal696:
> It only happens on canons loaded from file

Hi guys, yes, that the case; i tried and it happened each time a cannon tried to fire to something. im sure thats the cause.

Bertram25:
Ok, then the test is simple enough. :)
If we can test @Danimal696 's level and it doesn't crash when a cannon is firing on master, then the bug is fixed. I'm eager to see that level anyway. :)

@Danimal696 
If the bug is indeed fixed, I'll upload a new binary for you asap, so you can go ahead with the level more freely. :)

Danimal696:
Thanks, but everything would go smoother if the editor worked in the next exe so i dont have to jump between two OD versions

Bertram25:
> Thanks, but everything would go smoother if the editor worked in the next exe so i dont have to jump between two OD versions

[...]

PR #124, opened 2014-09-02
Updated Danimal's new map to its latest version.

Bertram25:
I could test cannons and no crash seen so far with firing ones. :)

IMHO this can be used to prove #120 is fixed.


hwoarangmy:
ok, perfect. We will open another issue if we see another crash then.

PR #130, opened 2014-09-09
Lava and water passability should be set independently

### LINKED ISSUE #101: Water and lava should be set passable/impassable independently for creatures

hwoarangmy:
fixes #101 

Patch content :
- Improved algorithm for A\* (using diagonals) and removed cutCorners
- Improved creature behaviour (Now, they will decide what to do on idle mode only). I've not looked at the fighting behaviour yet.
- Speed can be set independently on ground/water/lava. The speed is used to compute the best way. If speed is set to 0, the corresponding tile is not passable.
- Fixed a bug where synchronisation was not very good between server and clients when creatures were moving. This resulted in creatures going through walls :)
- Added support for floodfilling (used by A\* to find path) to work with ground/water/lava
- Added command console to force a creature to go somewhere in the map and another to display floodfill in logs

How to test :
2 new console commands :
logfloodfill
setdest <creatureName> <tileX> <tileY>

Test floodfill :
In TestSingleplayerSmallPassability.level :
logfloodfill
You should see the floodfill values in the logs.
The 1st value corresponds to the value for ground units (if tiles have different values, there is no way between them).
The 2d value is for creatures than can go on water + ground
The 3rd for creatures than can go on lava + ground
The 4th for creatures than can go on water + lava + ground
In this map, there is 1 tile (0;0) surrounded by diggable walls. If you dig them, floodfill should change (same value for this tile and the ones linked to it)

In TestSingleplayerSmallPassability.level, there are 2 creatures called Lava (PitDemon, can go on lava - very slow) and Water (slime, can go on water - faster on water than ground) :
setdest Lava 9 0
You can try to play with both creatures. Note that slime is faster on water than ground so it should consider going on water when faster. On the other hand, PitDemon is slow on lava and should try to avoid it when faster.


Danimal696:
Good job! thats some monster patch there, are flying creatures affected by this? from what i understand they just need to have the same speed on all grounds?

hwoarangmy:
> Good job! thats some monster patch there, are flying creatures affected by this? from what i understand they just need to have the same speed on all grounds?

Correct :)

Currently, I've changed creatures def so that no creature except PitDemon can go on lava (but with a huge speed penalty). Flying units have no penalty on water, Kobolds have 50% penalty and slime goes 100% faster on water. This can be adjusted later but it's pretty usefull to test the patch :)

hwoarangmy:
@Bertram25 @akien-mga now that conflicts are resolved, can you have a look ?

akien-mga:
I'm having a look, it's an impressive (and awesome) pull request :-D

akien-mga:
It seems you merged your own branch in itself, and now you have all commits twice in the log.

It also tells there is a conflict:

> commit e1992649f81581300fa0bd76d9a644aeae43d7ff
> Merge: 892a21a 8a82d1b
> Author: hwoarangmy hwoarangmy@hotmail.com
> Date:   Tue Sep 9 23:06:45 2014 +0200
> 
>    Merge branch 'development' of https://github.com/hwoarangmy/OpenDungeons into development
> 
>   Conflicts:
>         source/MenuModeSkirmish.cpp

akien-mga:
> How to test :
> 2 new console commands :
> logfloodfill

I did not manage to see the output of this command. Maybe I need a debug build to get the correspond log messages?

> setdest 

Works like a charm. I think it requires 3 positions since there should also be the z-axis, so `setdest Lava 9 5 0` for example to point the pit demon to the right side of the lava river.
I could test that the pit demon and the slime try to use the fastest way depending on their position.

hwoarangmy:
> It seems you merged your own branch in itself, and now you have all commits twice in the log.

Yes, that's the first time I resolve a conflict. It seemed logic tome to merge that in my branch but it might not be the "usual" way to do so. Do you know how I should have done ? If not, maybe @Bertram25 will tell us when he will get back.

hwoarangmy:
> I think it requires 3 positions

No, only 2 numbers are needed. The 3rd is ignored.

> I did not manage to see the output of this command

Strange, it should be in te terminal like other logs. For my part, I use a modified main to output everything in a file becaus on windows, I get no log otherwise.
They should look something like (for each tile) :
Tile floodfill : [1,1] - fullness=0, [0=1], [1=20], [2=8], [3=21]

Explanation :
[1,1] -> tile 1,1
fullness=0 -> Fullness is... 0 :)
[0=1] -> floodfill on ground : 1
[1=20] -> floodfill on ground : 20
[2=8] -> floodfill on ground : 8
[3=21] -> floodfill on ground : 21

Bertram25:
@hwoarangmy Nice work, and I can feel it's been some harder one. Congrats. :)

I don't really know whether it is the issue you had, but when merging the main branch into yours, you should use the --rebase option, this will reapply your commits on top of history, and sometimes force you to handle conflicts right away, though.
I don't care if you do it for the current merge, though.

I've got a few questions before merging that I'll let you see inline.

Best regards,

hwoarangmy:
> when merging the main branch into yours

Yes. Actually, it's not the way I use. In reallity, I usually list the files that changed between the main repository and the time when I forked. Then, I delete my repository and refork original. I replace all the files that did not change and apply my changes on the others.
But I will try your way next time. It surely is better but I'm not sure about the exact command I should use. I will search a bit.

Bertram25:
> Yes. Actually, it's not the way I use. In reallity, I usually list the files that changed between the main repository and the time when I forked. Then, I delete my repository and refork original. I replace all the files that did not change and apply my changes on the others.
> But I will try your way next time. It surely is better but I'm not sure about the exact command I should use. I will search a bit.

Note that I'm not into forcing you into one way or another. Experience has proven that what is to be done is not the best practice 100% of the time. :)
So feel free to have a go at trying the git pull --rebase command but please don't turn off your instinct. ;) You might still be right for certain obscure cases.

hwoarangmy:
> please don't turn off your instinct

My instinct is to not delete my repository each time I make a pull request ;) But it's an easy way. I can do that here because there are not too many coders and I can keep up. But on a bigger project... That's why I prefer to use the "standard" way :)

Bertram25:
IMHO, this is mergeable now, the size() == 0 to empty() change apart. Do you want to change this in another pull request?

hwoarangmy:
sorry, i forgot
let me do that now

Bertram25:
That's some great step ahead. Thank you @hwoarangmy :)

PR #135, opened 2014-09-10
Sounds deinit + fix 133

Bertram25:
I improved a bit the way sounds are deinitialized. It's better on my own but feel free to tell me on yours. :)

Also disabled the normal appliance on the luminescent tiles.


hwoarangmy:
Not much to say. If SFML says it is the way to do it, we should.

PR #136, opened 2014-09-10
Add Danimal last models, display room/trap price when building and avoid unwanted selected tile when toggling in windowed mode

### LINKED ISSUE #89: Add a tooltip showing the price of a room tile/trap when hovering its icon
As the topic says. For now it's by trial and error that you know whether you can place a room tile.

### LINKED ISSUE #131: When toggling application (Alt + tab), when OD is re-entered, mouse position is considered like a tile selection
Even if mouse buttons are not pressed

hwoarangmy:
fixes #131 
fixes #89


akien-mga:
Maybe you can also add TrainingDummy3? http://forum.freegamedev.net/viewtopic.php?f=38&t=5528&start=150#p59351

hwoarangmy:
> Is the TrainingDummy2 really meant to replace TrainingDummy1,

As said in the forum, the TrainingDummy2 is not very nice when against a wall because when triggered, it goes through. That's why I have added them on central spots. TrainingDummy1 are on wall spots.

> Maybe you can also add TrainingDummy3

I will have a look to see where it fits well :)

akien-mga:
> > Is the TrainingDummy2 really meant to replace TrainingDummy1,
> 
> As said in the forum, the TrainingDummy2 is not very nice when against a wall because when triggered, it goes through. That's why I have added them on central spots. TrainingDummy1 are on wall spots.

That makes sense yes, I hadn't looked at the context in the commit :-)

Bertram25:
Nice patches. :) This will be useful and refreshing.

@hwoarangmy Note that #89 is speaking of also showing a tooltip when hovering the room/trap icons.
I may have misread the code but I do think you are showing the name and price only in game when hovering/dragging tiles.
In that case, the other tooltip is yet to be done before closing it.

Unless it is ok for you guys? We may even move the rest of the issue for later?

Regards,

hwoarangmy:
> Note that #89 is speaking of also showing a tooltip when hovering the room/trap icons

Yes I know. I've had a look at how to display tooltips but from what I've seen on CEGUI forums, everything is done in OD code to make them work. But they don't show up. And since I wanted to display the total price of a room, anyway, I didn't want to spend too much time on that and just skipped the tooltip. In any case, displaying tooltips for that or something else should be nice but it can be done later...

Bertram25:
> In any case, displaying tooltips for that or something else should be nice but it can be done later...

Ok. Moving the issue to later. :)

Bertram25:
@hwoarangmy I feel like this one has slipped through everything we said here. ;)

> Do you want to merge now, or do you prefer adding the training dummy 3 before that?

hwoarangmy:
No, I'm going to add training dummy 3. It's not a big deal and it fits well here :)

akien-mga:
Giving it a try in-game, and then I'll merge.

akien-mga:
Tested locally, it looks great :-)

Just one concern over the placement of the active spots, see the training room on the left side of this screenshot. One would expect some dummies on the wall on the right, but it seems the 4xX type of room is not that well handled (maybe it's the same for all rooms with an even width?).
odscreenshot_1

hwoarangmy:
> it seems the 4xX type of room is not that well handled

For having tested it for a while, I confirm that. But it is not related to this patch. It is the way active spots were handled before I started to play with them. I don't really know if it is an expected behaviour to force certain room sizes but I would guess it's a bug :)

Bertram25:
Fine to merge. I'll deal with credits discrepancies myself.

hwoarangmy:
thanks :)

Bertram25:
Np. That's the least I can do to help. :)

Credits updated.

Bertram25:
> For having tested it for a while, I confirm that. But it is not related to this patch. It is the way active spots were handled before I started to play with them. I don't really know if it is an expected behaviour to force certain room sizes but I would guess it's a bug :)

Btw, I confirm. It's a bug, even if I didn't make it that way in the first version of active spot (or I got something wrong somehow :S?). 

The most weird-looking part is that there is no active spots on the right part of the room (on screenshot's left training room). I do think it's fine if the central spots aren't perfectly aligned because of the room tile number being even, though.
I'll open an issue about this.

PR #140, opened 2014-09-11
Added support to cross-fade some battle music when encountering enemies.

### LINKED ISSUE #23: Music handling enhancement
At the moment, music handling is very basic.
- [x] Change music tracks handling from integer index to a string based array.
- [x] Add fade in/fade out and cross-fading support.
- [x] Adds support to make the music played configurable in the level a map data.
- [x] Adds support for combat and peaceful music using a cross-fade.

Bertram25:
As client game map info parameters were loaded after the game mode was activated
due my recent change about loading the level only once the server tells which to load,
I had to fix the game mode to activate only after the client game map level loading.

This fixed the calm music not played anymore when starting the game.

I also made the server tell the specific client when attacked or attacking because
the attack and damage taking logic was only done on server-side.

Fixes #23


akien-mga:
It works great! :-)
The cross-fade works particularly well, even we fading out after a battle and then abruptly fading in again because a new monster shows up.

My only concern is that the battle music tends to produce some "saturation noise" on the loudest drum blows. It's not the first time I experience this in OD, when starting a game I always here some of those noises from time to time when a new sound plays, and usually it stops after a while.

I think it's on the software end because I have pretty good speakers on my laptop, and I'm not playing with the volume maxed out. Maybe I should open a new issue for this?

Bertram25:
> My only concern is that the battle music tends to produce some "saturation noise" on the loudest drum blows. It's not the first time I experience this in OD, when starting a game I always here some of those noises from time to time when a new sound plays, and usually it stops after a while.
> I think it's on the software end because I have pretty good speakers on my laptop, and I'm not playing with the volume maxed out. Maybe I should open a new issue for this?

True. I'll normalize the music file. This will make the issue much less frequent.
The issue seems to come from somewhere between SFML and openAL. @tomluchowski already told us about it several times, but there is not much we can do about it, apart from making sure to normalize sounds/music.

Thus, I don't think you should open an issue about the root of it since it's external, but if you catch some unnormalized sound/music, feel free to tell. :)
EDIT: Issue on SFML forums and the given solution: http://en.sfml-dev.org/forums/index.php?topic=2410.0
EDIT2: The official SFML issue: https://github.com/LaurentGomila/SFML/issues/310

Bertram25:
@hwoarangmy Should be done now. :)

@akien-mga I also normalized the battle music. It should be better now.

akien-mga:
It's perfect now :-)

I'll try to track down the other sounds that need to be normalised and to list them in an issue.

hwoarangmy:
That will be a good improvement. I will try as is.
Can you just have a look at the int thing ?

hwoarangmy:
It works well. Do you want to change the int now or later ?

Bertram25:
Should be fixed now. thanks for the input. :)

Bertram25:
I had to rebase to fix small conflicts in the map loader code, since I merged the editor PR.

source/MapLoader.cpp
```diff
@@ -167,14 +167,18 @@ bool readGameMapFromFile(const std::string& fileName, GameMap& gameMap)
         param = "Music\t";
         if (nextParam.compare(0, param.size(), param) == 0)
         {
-            gameMap.setLevelMusicFile(nextParam.substr(param.size()));
+            std::string musicFile = nextParam.substr(param.size());
+            gameMap.setLevelMusicFile(musicFile);
+            LogManager::getSingleton().logMessage("Level Music: " + musicfile);
```
hwoarangmy: Typo here : musicfile -> musicFile

Bertram25: /me hides in utter shame XD

Bertram25:
Should be ok now.

hwoarangmy:
ok, seems good. As expected, it breaks the editor but that's something I had already seen :)
I'm gonna fix that.

Bertram25:
Thanks :) For fixing the editor, too :D

hwoarangmy:
> Thanks :) For fixing the editor, too :D

No problem. It's just because we made 2 pull request linked at the same time :)

PR #142, opened 2014-09-12
Fixed editor launch.

### LINKED ISSUE #17: Editor: Fix cursor behaviour when moving a creature around.
Currently, moving a creature around makes the tile selector duplicate itself from the dragging start to the dragging end.
While this is correct when selecting tiles to add or placing rooms, this is not when doing something else.

The bug was somehow introduced with this commit: https://github.com/OpenDungeons/OpenDungeons/commit/eb7806ed2b1c70109cf9a67a61a279979565de78

hwoarangmy:
Fixed editor tile selector
Rooms and Traps layout is now shared between Editor and game.
Mouse handling copied/pasted from GameMap to have the same logic. For example, creatures can be picked up like ingame.
fixes #17
By default or when right mouse button is pressed, nothing is selected in editor mode. To add room/trap/tiles, you have to clic on the corresponding icon.
Added server messages for editor. In editor mode, if a room/trap is built on dirt tiles, gold tile or rock, they are claimed/digged automatically.
Rooms/traps cannot be built on tiles where there are already creatures/traps/rooms
Added a parameter to Server for him to know if we started editor or game mode. In editor mode, creatures have no turn :)
A map can be saved only if no creature is in the hand
Within the editor, creatures can be dropped everywhere they can walk (dirt, claimed - same color or not, water, lava).
Fixed a bug when reading map info (space is considered as a separator char by stringstream)

### DIFF
--- a/gui/OpenDungeons.layout
+++ b/gui/OpenDungeons.layout
@@ -95,61 +95,8 @@
             <Property name="Area" value="{{0,0},{1,-133},{1,0},{1,0}}" />
             <Property name="TabHeight" value="{0,30}" />
             <Property name="RiseOnClickEnabled" value="False" />
-            <Window type="DefaultWindow" name="Rooms" >
-                <Property name="Area" value="{{0,0},{0,0},{1,0},{1,0}}" />
-                <Property name="Text" value="Rooms" />
-                <Property name="MaxSize" value="{{1,0},{1,0}}" />
-                <Property name="RiseOnClickEnabled" value="False" />
-                <Window type="OD/ImageButton" name="TreasuryButton" >
-                    <Property name="Area" value="{{0,20},{0,20},{0,80},{0,80}}" />
-                    <Property name="NormalImage" value="OpenDungeonsIcons/TreasuryButton" />
-                    <Property name="TooltipText" value="Build a treasury" />
-                    <Property name="InheritsAlpha" value="False" />
-                </Window>
-                <Window type="OD/ImageButton" name="DormitoryButton" >
-                    <Property name="Area" value="{{0,80},{0,20},{0,140},{0,80}}" />
-                    <Property name="NormalImage" value="OpenDungeonsIcons/DormitoryButton" />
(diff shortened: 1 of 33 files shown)


Bertram25:
Hi @hwoarangmy :)

I've not quite to say about the patch. The code looks good and funnily I think that using the server logic for map edition could lead one day to collaborative editing. ;)
Plus, there are a few nice bits there in usability as for rooms and traps. :)

Now I tested it live and I'm unable to select a creature in hand, move it and drop it somewhere else, at least on linux. Is there something I missed concerning how to get a creature?

I would also have another small request: To stop the music while you're editing.

Best regards, :)

hwoarangmy:
> Is there something I missed concerning how to get a creature?

Can't you pick them up like in game ? I've tested it pretty much by dropping creatures on water/lava/dirt to see if it works and I've never seen any problem.

> I would also have another small request: To stop the music while you're editing.

If you want, I will do that on my next pull request. I have copied/pasted the skirmish menu to allow to choose a level to edit. But I have a crash when reloading a map. It looks like I enter editor mode before the client gamemap is loaded. So I believe it is linked to your pull request and I will wait for it before my next one.

Bertram25:
> Can't you pick them up like in game ? I've tested it pretty much by dropping creatures on water/lava/dirt to see if it works and I've never seen any problem.

I copied/pasted the dungeon siege.level into levels/Test.level and I was able to take no creature at all. Might there be you made it so one can take only own-team creatures?

Bertram25:
> So I believe it is linked to your pull request and I will wait for it before my next one.

Yeah, might be. I should be done now, btw. :)

hwoarangmy:
> I copied/pasted the dungeon siege.level into levels/Test.level and I was able to take no creature at all. Might there be you made it so one can take only own-team creatures?

Yes, you may be right. I will have a look.

Bertram25:
Tried the patch. It works fine now. :)

PR #159, opened 2014-09-18
Workaround for plugin path + update ogre cmake

oyvindln:
fixed the workaround for the plugin path, and updated the ogre cmake files.


Bertram25:
There is a few wrong indentation used here and there. I will test the whole beast and tell you asap how it turned. Thanks for your help! :)

oyvindln:
FindOGRE.cmake is from the ogre project, it's not a file I've edited. I just updated the files to the ones in OGRE 1.9.

Bertram25:
> FindOGRE.cmake is from the ogre project, it's not a file I've edited. I just updated the files to the ones in OGRE 1.9.

Oh, my apologies. Nevermind that, then.

hwoarangmy:
I've tested the patch. On my windows computer (using MinGW), the plugin folder is still set to plugins where it has to be empty to work. It might be my installation that is weird since I compiled all the modules (boost, ogre, cegui, sfml, openal).
It doesn't break anything to me. But like every PR heavily changing cmake files, it has to be tested by as many people as possible :)

@Bertram25 , @akien-mga , have you had a look ?

oyvindln:
The plugin path change was only for linux. If you look at the code, you can see that the windows code simply sets the path to "plugins". Not sure if there's an easy want to auto-find libraries on windows.

As for screenshot patch, I added that later, didn't cross my mind that that would be automatically added to the pull request too, though it shouldn't cause any issues.

akien-mga:
I'll give it a try on Mageia.

@oyvindln: Indeed, when you do a pull request from a given branch, all further commits to this branch (or rebasing) get added to the pull request. That's pretty useful :-)

I don't know how @hwoarangmy and @Bertram25 handle it, personally I used a dedicated branch for my last pull request, but I know they use their development branch. So either they push further commits to another temporary branch, or they refrain from coding/committing :-P

hwoarangmy:
Yes, if you want to separe 2 commits, you have to create a branch.
In my case, as people are quite reactive in the team, I usually use my development branch and start working locally until the PR is merged. When someone merges it, if I did no heavy change, I usually delete and re-fork the main repository to keep updated with the latest sources.
And if I did heavy changes, I usually continue to work on my repository until next PR (depending on how the main repository have been updated since my fork) :)

But in your case, when you change cmake files heavily, you should probably do a branch for that because that kind of request is usually long to merge since everybody has to test on all environments available to make sure they are not broken. Which usually takes more time than the rest :)

oyvindln:
@hwoarangmy It's not really a heavy change as such, I only changed a line in CmakeLists.txt adding an extra step to fill in the plugin path in case it was missing. The rest of the changes is from updating cmake modules to the ones from Ogre 1.9, which I assume are well-tested already given that the version has been marked stable and is included in the ubuntu repos.  If you just look at the commit log it may look like a heavy change though, I probably should have stated that more clearly in the commit message :) I updated them mainly to see if they fixed the issue, but they didn't, though I thought they probably fixed other issues, so I didn't see any reason to use the outdated version.

Bertram25:
The overall change is fine to me.
- Working on linux, and the time is applied on screenshot names.

What could be done later is to prettify the date by making it look like OpenDungeons_YYMMDD_HHMMSS.png instead of a raw number.

Bertram25:
I'd propose that if the change doesn't break anything on windows, it should be merged as it is still an improvement.

hwoarangmy:
I've tested with MinGW, it is OK (no visible change). I will test with MSVS ASAP. Let's also wait for @akien-mga to make sure there is no problem. If it is OK for everybody, we can merge :)

> It's not really a heavy change as such

Yes, I've read your comment. But sometimes, weird things can happen when cmake files are changed. For example, I remember that when we declared the project at the beginning of cmake file (where it suited the best), it used to break @tomluchowski 's config. If you look at CmakeLists.txt, you will see that cegui is commented on debug mode for linking because it breaks @Bertram25 's config. In my case, as I mostly use debug version, I have to uncomment it to make it work. Just to say that sometimes, things that look like an improvement can break unexpected things ;)

Bertram25:
True, I thought @akien-mga had already tested the change. Ok, let's wait for him. :)

akien-mga:
I've tested and it works fine. It can now find the OgreVolume module that it couldn't locate previously.

I just wonder if it wouldn't be better for us to prefix the OGRE cmake modules with "Ogre" (and adapt the scripts to use the proper names in their includes)? This sync with OGRE upstream adds a lot of cmake modules with relatively generic names.

oyvindln:
I think the find functions in cmake are dependent on the filenames so I don't think we can rename them. Maybe we could seperate the different libraries into folders though. Everything in the modules folder is things from the respective libraries. Given that the files are used to locate the libraries in the first place, they kind of have to be included with the project.

Bertram25:
> I think the find functions in cmake are dependent on the filenames so I don't think we can rename them. 

True: http://www.cmake.org/Wiki/CMake:How_To_Find_Libraries#How_package_finding_works

IMHO, I'd rather not change upstream find modules if we can avoid it. It will help us stay in sync more easily when necessary. 
That said, if the ogre team does something weird, we should tell them. :) But here, I don't think it's that bad.

akien-mga:
Ok, let's merge then :-)

BTW I'm back with my connection issue when starting a local or multiplayer game; I'm now positive it is linked with my wireless connection, I'm just not sure yet if it's an issue in the router or in my configuration. Since I'm moving on Friday, I won't spend too much time on this, using another connection should be a radical fix :-)

Bertram25:
@akien-mga Your wireless connection problem related to the local loop is really weird. I hope you'll get help about that soon. :S

akien-mga:
It's definitely on my end though, I've noticed I can't download content through the libre gaming library Lutris, even though wget or curl work just fine to retrieve the same content.

Bertram25:
Sounds crazy to me, and though certainly perfectly logical once you'll have the reason...

akien-mga:
> Sounds crazy to me, and though certainly perfectly logical once you'll have the reason...

I have no clue about how to fix it, but I found a workaround: I just have to disconnect my wireless connection, and then I can join the server. I'll see if some Mageia wireless gurus have a clue about this; but as I said this will hopefully be forever fixed when I'll move on Friday :-)

PR #162, opened 2014-09-19
Fixed creature behaviour when fighting. With this patch :

### LINKED ISSUE #73: Have creatures face each other when fighting
Currently fighting creatures are not necessarily facing each other when battling.

They sometime occupy the exact same position (and then regardless of the direction of their battling animation, they can't face each other). In other times they occupy adjacent positions, but their animations are not directed toward each other (for example both creatures can have their animation in the same direction, making as if one creature was hitting the other but the other one is fighting with air).

It's not that clear on my screenshot, but it's pretty obvious in game.
od-battle1

### LINKED ISSUE #90: Make workers flee only when receiving first enemy blow.
It has been discussed on the forum that the workers should continue to work until they get damage (or at least until an enemy actually attacks it).

In the small map, it is quite obvious that there is something weird with the current behaviour as it is really hard to get workers claim tiles when they are not far from the enemy base.

### LINKED ISSUE #104: Creatures stop moving after a while - problem with action list
I let a game play during some time and I've seen strange things in action list like the one posted. I guess the root comes from maneuver action that is not popped for an unknown reason (I've not had look yet)

bug

### LINKED ISSUE #154: Combat behaviour upgrade
As discussed here: http://forum.freegamedev.net/viewtopic.php?p=59460#p59460

Currently, the creature behaviour in fight is rather passive.

In order to fix this, the creature thinking logic before entering battle should follow those rules:
- The fighter creatures should check their "health" + "mood" before going or keep going to combat or fleeing.
- During the fight or before, if HP < 10%, flee to bed.
- Before fighting, check HP + mood (hunger + awakness). If too low, the creature has up to 20% chances of fleeing (never more than 20% or the AI will never fight). Moreover, a creature that chooses to fight will not check its mood until it dies or until the battle is over (or its ennemies are not visible because they flee)
- Note that mood is decreased by hunger, drowsiness, and later, other creatures, lack of pay wage, and other things. For now, hunger + drowsiness can be checked < 10% instead.

Fighting logic should be :
1 - If it is at the range distance, it should shoot (if melee unit, in a neighboor tile)
2 - If too near or too far, move (at its range distance)
3 - During attack cooldown, like every other unit, it cannot move. That would allow not too slow units to retaliate even if a little bit slower than the ranged unit.
4 - Go to 1

The creatures should try to attack the closest ennemy to them.

We can remove Battlefields from code.

### LINKED ISSUE #161: Crash when loading a map
When selecting TestSingleplayerSmallEmpty.level:

[bt]: (1) ./OpenDungeons.bin : Seat::isAlliedSeat(Seat_)+0 [0x5e48b0]
[bt]: (2) ./OpenDungeons.bin : Seat::isAlliedSeat(Seat_)+0 [0x5e48b0]
[bt]: (3) ./OpenDungeons.bin : GameMap::getVisibleForce(std::vector<Tile*, std::allocator<Tile*> >, Seat_, bool)+0x9f [0x55866f]
[bt]: (4) ./OpenDungeons.bin : Creature::getVisibleForce(Seat_, bool)+0xc1 [0x532a81]
[bt]: (5) ./OpenDungeons.bin : Creature::doTurn()+0x14e [0x53f45e]
[bt]: (6) ./OpenDungeons.bin : GameMap::doCreatureTurns()+0x4a [0x5514da]
[bt]: (7) ./OpenDungeons.bin : GameMap::doTurn()+0x43 [0x559853]
[bt]: (8) ./OpenDungeons.bin : ODServer::startNewTurn(double)+0x474 [0x5b3344]
[bt]: (9) ./OpenDungeons.bin : ODServer::serverThread()+0x15d [0x5bbbed]
[bt]: (10) /usr/lib/x86_64-linux-gnu/libsfml-system.so.2 : +0x6508 [0x7fc319918508]
[bt]: (11) /lib/x86_64-linux-gnu/libpthread.so.0 : +0x80a4 [0x7fc319d360a4]
[bt]: (12) /lib/x86_64-linux-gnu/libc.so.6 : clone+0x6d [0x7fc318c12c2d]

hwoarangmy:
- Workers attack workers only.
- If workers are attacked by something else than a worker (creature, trap, ...), they flee
- When a creature or a worker is weak, he flees from combat
- If a fighter is weak and there are no ennemies around, he tries to sleep
- When a creature flees, it tries to go closer to its dungeon heart. If it cannot reach it, it flees randomly on visible tiles
- When a creature flees near the dungeon heart, it flees randomly on visible tiles
- Creatures always try to go to the farest position (depending on range) they can before engaging (within visible tiles). That means that ranged unit need melee units to avoid running away for ages.
- If a creature is in bad mood, it has 20% chances of fleeing instead of fighting when seeing an ennemy. For now, it is in bad mood if awakness < 10% and hunger > 90% at the same time

fixed #161 
fixed #154 
fixed #104 
fixed #90 
fixed #73 

To test it, I've changed the map TestSingleplayerSmall.level. Player 1 spawns with 1 BigKnight (range 1) and 1 Advanturer (range 3). Moreover, there is 1 claimed tile in the dungeon bottom right to be able to drop units and see them fight/flee.


Bertram25:
Well, I just couldn't go to sleep because of this PR. Best stuff ever seen so far. :D 

I almost lost the first time playing the small test single map because I was too confident. ;D
It's so good to see how reactive the creatures are now. I had a bunch of creatures running after a red one fleeing up to his dungeon were I had to retreat. Was fun!

There are a few things I've noticed but it will be hard to say if for some it's regressions due to this PR and not another one. The good being much superior to the potential bad, I'd be for merging and investigate later. :)

All seen on the Skirmish/TestSingleplayer.level file:
- The most weird: An enemy kobold was able to dig one of my claimed walls. It shouldn't have been able to, though, as a claimed walls shouldn't be diggable when at least one ground claim tile is next to it and it was the case. My bet is that the AI had previously marked a path to gold, and I claimed the wall on this path.
- Kobolds seem to claim tiles under enemy room and traps. This doesn't affect the enemy rooms but looks weird.
- When an enemy is attacking a kobold which is claiming ground tiles. The kobold doesn't move it seems if it's too far from his dungeon temple.
- When dying (or was it when killed by a cannon only? IDK), the Kobold doesn't stay in dead animation. After a few seconds, it goes back to idle and then disappear a few seconds later.

Seen on the small test skirmish map:
- FPS were giving the game hiccups at the beginning of the small test map, because probably of all the fights happening rather quickly. It was largely manageable, but we might have to do some profiling on that later. After the first ten minutes when there were less creatures alive, everything was back to normal.

Clearly not related to this PR but I wanted to write it down while I see it:
- Cannons deal way too much damage and can kiil any lvl 30 creatures.
- Cannons keep shooting at dead creatures until they disappear.
- Cannons now attack rooms, yeah! :D (Not a bug, just me being happy)

My hand is itching to merge this, so if nobody gives a good reason not to do so by tomorrow noon, I'll merge. ;)

Best regards,

hwoarangmy:
I would add to actual problems you listed :
- Creatures level up much too fast. So fast that HP increases and they become weak without fighting :)

Concerning what you said, I will have a look. But as you said, I think we can merge and handle that later :)

But I'm glad you enjoy the patch ^^

hwoarangmy:
BTW, concerning the points you listed, I would guess that :

> When dying (or was it when killed by a cannon only? IDK), the Kobold doesn't stay in dead animation. After a few seconds, it goes back to idle and then disappear a few seconds later.

That probably comes from the patch

> An enemy kobold was able to dig one of my claimed walls

Never seen that. You guess is interesting, I will have a look. But I think it is not related to the patch

> Kobolds seem to claim tiles under enemy room and traps

I've not changed anythin in Creature behaviour other than fighting. I think it is because when Kobolds use to see an ennemy, they transformed into Chuck Norris and began to kill everything nearby. With the patch, their behaviour is more expected...

> When an enemy is attacking a kobold which is claiming ground tiles. The kobold doesn't move it seems if it's too far from his dungeon temple

I will have a look. They are supposed to flee like other creatures :
- If the Dungeon temple is reachable, they should flee in its direction
- If the Dungeon temple is not reachable, somewhere far within the visible tiles

akien-mga:
> My hand is itching to merge this, so if nobody gives a good reason not to do so by tomorrow noon, I'll merge. ;)

I'll be testing it now. If I haven't tested by noon, you can merge, since I'll be away this afternoon :-)

akien-mga:
> Kobolds seem to claim tiles under enemy room and traps

Confirmed. As @hwoarangmy said, I think previously kobold used to crush enemy rooms, now that they don't anymore, they claim it. What a subversive little race :-p

akien-mga:
Awesome pull request :-D
I'll snatch it away from @Bertram25 and merge it now :-P

Bertram25:
Hey :D 

> Creatures level up much too fast. So fast that HP increases and they become weak without fighting :)

Yep, part of the problem is that the XP aren't reset to 0 once leveled up. This is trivial to add, so the first to do it. ;)

> Never seen that. You guess is interesting, I will have a look. But I think it is not related to the patch

From a quick glance to the code, it seems to be indeed that any marked tile is dug by kobolds, there is no conditions check on that. I do think the a marked tile should be unmarked when turning to a claimed wall, or be dug, and that's what I must have forgotten to add at the time.

> I've not changed anythin in Creature behaviour other than fighting. I think it is because when Kobolds use to see an ennemy, they transformed into Chuck Norris and began to kill everything nearby. With the patch, their behaviour is more expected...
> 
> Confirmed. As @hwoarangmy said, I think previously kobold used to crush enemy rooms, now that they don't anymore, they claim it. What a subversive little race :-p

XD. Yes, +1 for the Chuck Norris reference, btw.

> I'll snatch it away from @Bertram25 and merge it now :-P

Arr, again something we'll have to talk about on our next lan. :>

PR #163, opened 2014-09-20
Reworked the main menu not to look like a christian cross anymore.

Bertram25:


akien-mga:
Hehe, I wonder how many free software projects have references to religious symbols in their commit logs :-P

I like the new layout :-)

Issue #165, opened 2014-09-20
Kobolds can claim ennemy rooms/traps


hwoarangmy:
See comments in #162 
An easy fix would be to check if the tile has a room/trap before trying to claim it. But it would avoid kobolds to go there if it is a trap.
So, kobolds should not consider claiming rooms but they should test if a trap is on a tile only after they moved on it.

Bertram25:
Sounds fine to me. :)

hwoarangmy:
Concerning claiming trap, should we use the same as in DK ? ie: when a tile is claimed the corresponding trap is destroyed ? If I remember well, it was @Danimal696 opinion to do that (I remember he said that in DK2, it was too easy to kill traps by attacking). I think we all agreed but I am not sure. If someone remember, please help my failing memory :)

Bertram25:
My opinion is I'd rather say that until the trap is destroyed, one cannot claim the tile. Same for rooms.
But there might be flaws in this as well.

hwoarangmy:
That brings us to how to destroy a trap ? At first, it should be hidden. If a player triggers the trap, it will become visible. Then, can creatures attack it ? If there are no creature to attack, during how much time will the trap become visible ?

Bertram25:
You're right, certain traps should be hidden, but I think cannons shouldn't, so this should remain an option, IMHO.

I also do think the trap should remain visible once discovered as it won't move anyway, and it would look weird to let creatures rediscover it again and again.
Once discovered, then yes, I'd be for letting creatures destroy it before being able to claim the tile.
Thus, traps should have custom HP (not just the default 10 of a tile).
Feel free to correct me if you see something. :)

hwoarangmy:
It's not a correction, just another discussion about the subject. I was able to find what I was talking about :
#139

Danimal696:
My opinion is not mandatory, but im in favor of traps not being directly attacked even when discovered

Bertram25:
If you don't want them attacked by fighters when discovered, then what should happen, in your own opinion?

Danimal696:
they have to be triggered a number of times before becoming deactivated or claimed fast enougth so it doesnt trigger again when your kobolds are still over it (and die from it)

Bertram25:
I'm fine for this with spikes, but this doesn't fit well for cannons, right?

akien-mga:
I guess turrets and traps can't be handled exactly the same way.

hwoarangmy:
[...]

PR #169, opened 2014-09-22
Small fixes : workers claim ennemy room + Cannonns keep shooting + sometimes, ennemy kobold can claim claimed walls

hwoarangmy:
To test this patch, the most simple is to play multiplayer map small free for all (it is the same as multiplayer small but no alliance)
With player 2, try to mark tiles from dungeon 2 to dungeon 1. During kobolds are digging, with player 1, claim the tile in the way. The tile should not be marked anymore in player 2 screen and it should not be possible to mark it.
If you do the same with the multiplayer small 2v2 map, the claimed wall should be left marked and kobolds should dig it.
To test "kobolds can claim ennemy rooms", the most simple is to use the editor to build a room not too far from your dungeon :) Be careful of not doing it too near or your creatures will attack it


hwoarangmy:
BTW, #165 room part is fixed. The trap part will wait until we decide what we do about them

Bertram25:
Will test asap. The code change looks fine. :)

Bertram25:
I've tested it. Working fine for every points. IMHO, this can be merged. Anyone else wants to test before I do?

akien-mga:
I've tested that marked tiles are effectively unmarked if they are claimed by an enemy.
For the other points I'll rely on your tests, I think you can merge @Bertram25 if the code changes are good.

PR #179, opened 2014-09-24
Turn warnings on + active spots + renaming camera manager in ODFrameListener

hwoarangmy:


hwoarangmy:
conflict resolution on its way

Issue #58, opened 2014-08-09
Make OD system-wide installable on Linux


akien-mga:
I've started working on proper "make install" instructions for Linux. I will soon commit some changes to my fork and we can discuss them here before I do a proper pull request.

I managed to install OD on my system, putting the two binaries (script and OpenDungeons.bin) into /usr/games and the data into /usr/share/games/opendungeons
The data is:

```
$ ls /usr/share/games/opendungeons/
gui/  levels/  materials/  models/  music/  plugins.cfg  resources.cfg  resources.zip.cfg  sounds/
```

Now, I'm having an issue with the "levels" folder, since it is not linked in resources.cfg nor in resources.zip.cfg. I tried adding a "[Level]" or "[Levels]" category, but obviously it does not work, so my guess is that the "./levels" path is hardcoded for now. Any clue about reading it from the resources.cfg too?

EDIT: The above is also valid for the "scripts" folder. For now I worked around the issue by adding "cd /usr/share/games/opendungeons" in the launch script, so that the executable is started from a folder containing the "levels" and "scripts" folder. That's really messy :-p

akien-mga:
I have three commits there starting the work to make OD system-wide installable: https://github.com/akien-mageia/OpenDungeons/compare/OpenDungeons:development...development

"make install" now should work fine, but the approach I tried to use to make it still possible to run the game from the build directory seems not to be functional: I defined the paths in resources.cfg using ${CMAKE_SOURCE_DIR}, so that the binary retrieves the data from the root dir, but it seems that OGRE doesn't like the absolute path. Replacing the absolute path with "../" (since I built in a "build" subdirectory) does the trick, but I don't know a cmake variable that would be the ${CMAKE_SOURCE_DIR} as a relative path.
Maybe I should try another approach.

It might be that my changes broken the Windows build too, or at least it probably doesn't fix an already broken "make install" on Windows :)

My branch is a WIP, feel free to comment and give advice on how to do it in a better way.

hwoarangmy:
The level dir is hardcoded in MenuModeSingleplayer.cpp and MenuModeMultiplayer.cpp

Concerning the media, I believe the ressource file is used by ogre so I'm not sure if we can safely add new entries nor if we can read it on our own after having sent it to ogre. Anyway, I agree that it would be better to have the important dir in a config file rather than hardcoded.

Bertram25:
IMO, there are two ways here:
[...]

Issue #57, opened 2014-08-09
Linux: Drop the ".bin" binary suffix (i.e. encode the data path into the binary?)


akien-mga:
I've started some work on the CMakeLists.txt (I'll push it to my fork soon so that it can be reviewed), and dropping the ".bin" binary suffix is not as straight-forward as it looks like.

Currently the file cmake/config/OpenDungeons.in is used to generate a bash script that runs OpenDungeons.bin with the needed environment variable to let OD find its data:

```
#!/bin/sh
OPENDUNGEONS_DATA_PATH="/usr/share/games/opendungeons" exec "/usr/games/OpenDungeons.bin" "$@"
```

If we want to drop the ".bin" suffix (and set the binary name in lower case on Linux if possible), we need to find another way to let OpenDungeons know the location of the data directory (currently defined as OPENDUNGEONS_DATA_PATH here: https://github.com/OpenDungeons/OpenDungeons/blob/development/CMakeLists.txt#L507 ).

Any idea about how this could be done (i.e. how most games do it)? It looks somewhat messy to have a broken (as in, you can't execute it properly) OpenDungeons.bin binary in /usr/games and a script to make it work.

Bertram25:
Hi @akien-mageia :)

Well, I'll need to have a more careful look at how it is done, but there is something surprising me. Maybe you can tell:
- In what way renaming the binary is preventing the current script from running? (renaming stuff in it apart?) I fail to get it. 
- Currently, the code is getting the data path here, for instance:
  https://github.com/OpenDungeons/OpenDungeons/blob/development/source/ResourceManager.cpp#L127

What I saw elsewhere is that usually, such a variable is hard-coded by cmake at compilation time depending on the platform, making the launch script useless.

Example:
`DATA_PATH="/usr/share/game/opendungeons"` on linux
`DATA_PATH="."` on windows

...
IMO, it's what we should do.

What do you think?

akien-mga:
> In what way renaming the binary is preventing the current script from running? (renaming stuff in it apart?) I fail to get it.

I think the only reason is that the initial design of this system would be to put both the "OpenDungeons" script and the "OpenDungeons.bin" binary in the /usr/games (or /usr/bin) folder. That's somewhat messy because the OpenDungeons.bin binary can't be run as is, so it shouldn't be in the $PATH; but likewise it's bad practice (or so I've gathered from my packaging experience) to include binaries in data dirs such /usr/share/games/opendungeons.
[...]

Issue #53, opened 2014-08-09
Switch remaining DOS-like line endings (CRLF) files to Unix-like (LF)


akien-mga:
The following files in the repo use DOS-like line endings (CRLF):

```
[akien@cauldron OpenDungeons (development)]$ find -exec file {} + | grep CRLF     
./dependencies/angelscript/tests/test_build_performance/projects/msvc9/msvc9.sln:               UTF-8 Unicode (with BOM) text, with CRLF line terminators
./dependencies/angelscript/tests/test_feature/projects/msvc7/msvc7.sln:                         ASCII text, with CRLF line terminators
./dependencies/angelscript/tests/test_feature/projects/msvc10/msvc10.sln:                       UTF-8 Unicode (with BOM) text, with CRLF line terminators
./dependencies/angelscript/tests/test_feature/projects/msvc2012/msvc2012.sln:                   UTF-8 Unicode (with BOM) text, with CRLF line terminators
./dependencies/angelscript/tests/test_feature/projects/msvc9/msvc9.sln:                         UTF-8 Unicode (with BOM) text, with CRLF line terminators
./dependencies/angelscript/tests/test_feature/projects/msvc6/msvc6.dsp:                         ASCII text, with very long lines, with CRLF line terminators
./dependencies/angelscript/tests/test_feature/projects/msvc8/msvc8.sln:                         UTF-8 Unicode (with BOM) text, with CRLF line terminators
./dependencies/angelscript/tests/test_performance/projects/msvc9/msvc9.sln:                     UTF-8 Unicode (with BOM) text, with CRLF line terminators
./dependencies/angelscript/add_on/autowrapper/generator/generator.sln:                          UTF-8 Unicode (with BOM) text, with CRLF line terminators
./dependencies/angelscript/angelscript/source/as_callfunc_x64_msvc_asm.asm:                     ASCII text, with CRLF line terminators
./dependencies/angelscript/angelscript/source/as_callfunc_arm_xcode.S:                          assembler source, ASCII text, with CRLF line terminators
./dependencies/angelscript/angelscript/source/as_callfunc_arm_gcc.S:                            ISO-8859 text, with CRLF line terminators
./dependencies/angelscript/angelscript/projects/dev-cpp/AngelScript.dev:                        ASCII text, with CRLF line terminators
./dependencies/angelscript/angelscript/projects/dev-cpp/Makefile.win:                           makefile script, ASCII text, with very long lines, with CRLF line terminators
[...]

Issue #25, opened 2014-07-29
Made OD follow paths policies on Windows, Mac and Linux.


Bertram25:
At the moment, the game app is writing log files and shader debug data in the same place that it is running.

While this is useful when debugging, this should not happen on release mode and follow system standard data paths instead.
User saved games should also do the same.

akien-mga:
Some details about the paths to use on Linux (and Mac?): http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html
Basically config files should go in XDG_CONFIG_HOME/OpenDungeons (or lower case name, depending on what we decide for the linux binary in #57) i.e. ~/.config/OpenDungeons, and user data (saves and custom maps or custom tilesets, etc.) should go to XDG_DATA_HOME/OpenDungeons i.e. ~/.local/share/OpenDungeons.

Bertram25:
As told in a743cb2:

The ogre.cfg and other cfg files should be in: ~/.config/OpenDungeons
Later, the saves should be in: ~/.local/share/OpenDungeons

Such paths shouldn't be hardcoded but rather taken through XDG\* variables using getenv() when possible.
I'll deal with this issue while I'm on such things anyway.

Bertram25:
Ah, to be clear about it, I'll let the code setting paths for Mac as is. I can't test that one since I don't own a mac. I'll deal with windows and linux, though.

Bertram25:
@oyvindln On developer builds, when the data path is overriden to "./". Do you think I should also make it so the ShaderCache should be written there, too? Or should I let it written to: ~/.local/share/opendungeons/ShaderDataCache/ instead?

akien-mga:
IMO, everything should be in the source tree for developers builds.

But then the question would be, what about end-users who compile the game themselves and run it in the source tree? I think we can expect them to run a "sudo make install" and make their build a "non-developer" build. If not, well let them have the save games and everything in the build tree, it's probably simpler that way (and then we developers can do as many builds and saves and caches as we want without wreaking havoc in our ~/.local/share/opendungeons folder where we have our on-going playthrough of Danimal's latest awesome campaign :-P

Bertram25:
You have a point. :) Ok, I'll do that.

akien-mga:
[...]

PR #184, opened 2014-09-25
Made the resource manager retrieve the OD installation path at load time...

Bertram25:
I also made it so the ./resources.cfg file is tested first to override the data path to "./",
to still permit developer builds when necessary

Works fine for me but needs testing though to make sure.

Should help with #58

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,13 +4,13 @@ cmake_minimum_required(VERSION 2.8)
 #### Project settings ############
 ##################################
 
-project(OpenDungeons)
+project(opendungeons)
 
 #postfix bin if we are on linux as we want to launch with a script
 if(WIN32)
     set(PROJECT_BINARY_NAME "OpenDungeons")
 else()
-    set(PROJECT_BINARY_NAME "OpenDungeons.bin")
+    set(PROJECT_BINARY_NAME "opendungeons.bin")
 endif()
 
 # Project version
@@ -20,6 +20,17 @@ set(OD_PATCH_LEVEL   9)
 
 # Set the project version ready to be used in the code.
 set(OD_VERSION "-DOD_VERSION=\\\"${OD_MAJOR_VERSION}.${OD_MINOR_VERSION}.${OD_PATCH_LEVEL}\\\"")
+
+# Set the data path depending on the platform
+if(WIN32)
+    set(OPENDUNGEONS_DATA_PATH ".")
+    set(OPENDUNGEONS_DATA_DEST ".")
+else()
+    #Set data location to where we install if we want to use the installer
+    set(OPENDUNGEONS_DATA_DEST share/games/${PROJECT_NAME})
+    set(OPENDUNGEONS_DATA_PATH ${CMAKE_INSTALL_PREFIX}/${OPENDUNGEONS_DATA_DEST} CACHE PATH "")
+endif()
+set(OD_DATA_PATH "-DOD_DATA_PATH=\\\"${OPENDUNGEONS_DATA_PATH}\\\"")
 
 # Where we want the binary
 set(EXECUTABLE_OUTPUT_PATH ${PROJECT_BINARY_DIR})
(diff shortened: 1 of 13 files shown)


Bertram25:
Should be almost done for #25 , too.

akien-mga:
As far as I understand it this PR fixes most of #57. I'll test whether the OpenDungeons.bin binary can be run both from the source directory and from any other directory when data is located in OPENDUNGEONS_DATA_PATH; if it works, fixing #57 simply means dropping the IF here: https://github.com/OpenDungeons/OpenDungeons/blob/development/CMakeLists.txt#L10-L14

Though we might want to define PROJECT_BINARY_NAME in lower-case on Unix, and then use PROJECT_BINARY_NAME to define the default installation path for the data.

akien-mga:
There is still an issue with loading the levels/ and scripts/ directories it seems (when the data is installed system wide). See the discussion about it on https://github.com/OpenDungeons/OpenDungeons/issues/58#issuecomment-51886016

oyvindln:
Ah, I had just made a patch removing dirent, now it's going to be a mess to merge :/

Bertram25:
> There is still an issue with loading the levels/ and scripts/ directories it seems (when the data is installed system wide). See the discussion about it on #58 (comment)

@akien-mga Ok, I'll have a look.

> Ah, I had just made a patch removing dirent, now it's going to be a mess to merge :/

@oyvindln  Feel free to create it and we'll merge yours after review first. I'll rebase afterwards. :)

Bertram25:
Own TODO so far:
- [x] Use boost::filesystem instead of custom file exist code.
- [x] Use boost filesystem also for the createDirIfNotExist() function, too, if @oyvindln hasn't already done it.
- [x] Use a lower-cased project binary name and use it when setting the installation path.
- [x] Make sure to use the resource path prefix when searching for music, sounds, levels, and scripts.

Feel free to point out other blockers. :)
Once all this is done and merged, I'll deal with XDG stuff ( #25 ). Akien will then be able to drop the launch script (AFAIK there was one, right?) and .bin in the binary name.

Bertram25:
@akien-mga @hwoarangmy @oyvindln  I should be done now according to the TODO above. Feel free to point me any errors. :)

Bertram25:
Ah, to get a lowercase binary and installation path, you will need to clear your cmake cache.

Bertram25:
> Just to be clear, this applies at runtime only? So if there is data in "./" when building the source, it won't remove the OD_DATA_PATH from the path? But it will do temporarily when running the game?

Yes. :) Useful for us, poor devs. ;)

Bertram25:
Ah, btw. Don't start ./OpenDungeons.bin once build, now it is ./opendungeons.bin
I fell on that one a few times before removing the old binary. XD

Bertram25:
I'll stop the commit frenzy. See ya later guys. :)

akien-mga:
Code-wise it's looking good (as far as I can understand the code I read, that is :-p), apart from the line ending issue with scripts/console.as (another proof that we should take care of #53, at least for our own files, maybe not for the AS upstream files that need updating).

I'll test it live now.

akien-mga:
This works GREAT :-D

@hwoarangmy, @oyvindln To test the system wide installation on Linux, since there are no "make install" instructions yet (I'll be on it once this gets merged), you can put all the relevant data manually in /usr/local/share/games/opendungeons (or do a symbolic link to your source directory), and then move the opendungeons.bin binary anywhere, and run it. It should work :-)

hwoarangmy:
I won't be able to test until tonight. If someone test and sees it is ok, no need to wait for me before merging :)

akien-mga:
@Bertram25 Once this gets merged, I think you might be able to merge the whole development branch in stable and do a Windows release (if you want to).

Bertram25:
> @Bertram25 Once this gets merged, I think you might be able to merge the whole development branch in stable and do a Windows release (if you want to).

Ok, I'll do it. :) But before that, I need to take care of a certain commit :crying_cat_face: >:}]

Bertram25:
@hwoarangmy @akien-mga Sunny day, I might not do much before evening. ;)

Bertram25:
@akien-mga as script line ending commit fixed. :)

hwoarangmy:
It's ok for me on windows/MinGW. I will let someone else using linux merge that as I think it needs more test with this OS.

Bertram25:
Good to hear it's ok on windows. :) I'll let @akien-mga merge once/if he's ok.

PR #189, opened 2014-09-27
Move removal of some warnings to the OD_OPT_FLAGS variable on MinGW

akien-mga:
As discussed on https://github.com/OpenDungeons/OpenDungeons/commit/8299752bf611159a5f9d7dedc974d470349786b6#commitcomment-7948989

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -79,8 +79,12 @@ else()
     endif()
     if(MINGW)
         if(NOT OD_CXX11_FLAGS)
-            set(OD_CXX11_FLAGS "-std=gnu++11 -Wno-unused-local-typedefs -Wno-reorder -Wno-format" CACHE STRING "Compilation flags to enable C++11 support" FORCE)
+            set(OD_CXX11_FLAGS "-std=gnu++11" CACHE STRING "Compilation flags to enable C++11 support" FORCE)
         endif()
+	# Disable some warnings on MinGW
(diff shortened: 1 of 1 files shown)


Bertram25:
Is fine for me. @hwoarangmy will surely have a closer look. :)

akien-mga:
Yes I would like to know if those warnings should be disabled only for MinGW (as done previously), or if it would also be relevant to disable them on all GCC platforms.

Bertram25:
I'll try a full compile shortly. :)

Bertram25:
I've got the same root making @hwoarangmy request for disabling certain warnings. I'd say yes for every gcc, at least for now.

Bertram25:
> I've got the same root making @hwoarangmy request for disabling certain warnings. I'd say yes for every gcc, at least for now.

@akien-mga Nevermind what I said, I'm fixing all reordering warnings + some bonus. ;)

Bertram25:
I've added the reordering fix in #190 
For GCC, I think you can let -Wall alone. Only the AS strict-aliasing warnings are left AFAIK.
(And those can't be removed like that IIRC.)

hwoarangmy:
Ok, I had not seen the discussion on the warning philosophy, sorry for using a non suited flag. If you feel it's better like this, I have nothing to say.

Bertram25:
Np, :) Merging...

PR #201, opened 2014-09-29
Fix #198

### LINKED ISSUE #198: The loading menus aren't centered on screen and behave not quite good when resized.
I'm on it.

Bertram25:


akien-mga:
I'll have a look. Don't make me wait too long on #198, I'm itching to do a PR for https://github.com/akien-mga/OpenDungeons/tree/linux-install ;-)

akien-mga:
**Edit:** Or I'm silly and I forgot to run `git up` (my alias for `git pull --rebase`).

It doesn't seem to be working that well here on 1920x1080.

od-gui-1920x1080

od-gui2-1920x1080

akien-mga:
It works fine, the windows are always centred.

My only concern is that we propose those two resolutions in the Ogre Rendering Settings: 640x480 and 800x600, but the menu is completely broken for the first one, and usable but not that great for the second one.

Can't we have CEGUI resize the menu buttons when the window resolution goes under a given threshold, to keep a nice aspect ratio?

Bertram25:
> It doesn't seem to be working that well here on 1920x1080.
> Edit: Or I'm silly and I forgot to run git up (my alias for git pull --rebase).

Ah, for a sec. I was like: Woot?? O.O' XD

> My only concern is that we propose those two resolutions in the Ogre Rendering Settings: 640x480 and 800x600, but the menu is completely broken for the first one, and usable but not that great for the second one.
> Can't we have CEGUI resize the menu buttons when the window resolution goes under a given threshold, to keep a nice aspect ratio?

Good question:
I made it so it is looking nice on 1024x and above, but if lower resolutions are desired, I can review all that.
What resolution do we want to support, guys? @Danimal696 @oyvindln @hwoarangmy 

Anyway, the windows should behave better than before so I'd request  a merge and I'll handle that afterwards according to the decision taken. :)

Danimal696:
does anyone still use lower than 1204x768?

oyvindln:
It's useful to run in windowed mode when developing, so having 800x600 working would be nice.

akien-mga:
I agree with @oyvindln. The game already runs quite fine in low resolution, there's only the GUI that needs some downscaling, so I guess it's worth the effort. I'll open another issue for this.

Bertram25:
Ok, 800x600 or even 640x480?

akien-mga:
I'd say if we can scale the GUI down for 800x600, then let's do it for 640x480 too.

IMO we should not be limited by the support of lesser resolutions, i.e. if some new features won't work at all with these resolutions, then we can drop the support of those resolutions. But as long as they are working OK in-game, let's continue to support them if it's not too much work.

akien-mga:
Opened #203 for this, maybe we can continue the discussion there.

Issue #202, opened 2014-09-29
Link game icon with GUI on Linux


akien-mga:
**Edit:** Icon embedded for Windows has already been achieved for 0.5.0, what's left to do (or decide not to do if it's too complex) is to get the icon shown by Linux window managers in the title bar and task bar.

On Linux at least, the OD icon is not linked with the GUI, so the game window appears with the default Xorg icon (both in the taskbar and in the title bar).

A prerequisite to fixing this might be to design a nice-looking low-res icon (it should look good in 16x16 and 32x32, for higher resolutions the current icon is just fine) based on the current OD logo.

Bertram25:
I guess I'm the icon guy ;) But isn't the default icon size requested by XDG specs 48x48?

I would concentrate on that one and that one alone. WDYT?

akien-mga:
What I mean is that the current icon ( https://github.com/OpenDungeons/OpenDungeons-media-source/blob/master/Logo/OpenDungeons-Logo.svg ) looks quite good already when resized to 48x48, see:
![opendungeons](https://cloud.githubusercontent.com/assets/4701338/4447568/bc0a8b9c-480b-11e4-84df-476cfde64828.png)

But with smaller resolutions, you can't really see the details of demonic "D", and it becomes hard to read.
The XDG specs require at least a 48x48 icon, but you can provide more in the hicolor theme, with the help of the various subdirectories in /usr/share/icons/hicolor, e.g. with:
/usr/share/icons/hicolor/16x16/apps/opendungeons.png
/usr/share/icons/hicolor/48x48/apps/opendungeons.png

If your desktop environment needs an icon small or equal to 16x16, it will use the one in /usr/share/icons/hicolor/16x16/apps/opendungeons.png and scale it down if need be.
If it needs something between 16x16 and 48x48, it will use /usr/share/icons/hicolor/48x48/apps/opendungeons.png
This way we can provide a different icon in those two paths.

Bertram25:
Ok, now I see the point completely. I'll work tomorrow on the 32x32 and 16x16 versions if nobody else does it. :)

Bertram25:
What do you think of those?
![od-16x16](https://cloud.githubusercontent.com/assets/765917/4809166/12c712da-5ea9-11e4-81fd-614c0ce51c35.png)
![od-32x32](https://cloud.githubusercontent.com/assets/765917/4809168/168b5962-5ea9-11e4-8d31-86a177892423.png)
[...]

PR #204, opened 2014-09-29
Add installation instructions for Linux + some distribution bonus (desktop file, icon)

### LINKED ISSUE #58: Make OD system-wide installable on Linux

akien-mga:
Fixes #58.


Bertram25:
Testing now :)

akien-mga:
Note that on some distributions, /usr/local/games might not be in the PATH, so it could be that you won't be able to run "opendungeons" directly. I don't think that's something we have to cater for though, I already asked on the Mageia development mailing list if /usr/local/games could be added to the PATH (there is /usr/bin, /usr/games and /usr/local/bin already).

Same thing for the icons in /usr/local/share/icons/hicolor, it seems they are not picked up automatically here, so I had to configure the desktop file to point to the icon directly. So for the desktop file the icon field is `Icon=/usr/local/share/icons/hicolor/scalable/apps/opendungeons.svg`. I fear that this might break the purpose of the hicolor theme that should retrieve the correct size of the icon if it's available (as discussed in #202).
But that's something that can be improved afterwards (e.g. test whether the CMAKE_INSTALL_PREFIX is /usr, and use directly `Icon=opendungeons` in this case (since the distro should search for it directly in /usr/share/icons/hicolor).

Bertram25:
I wouldn't fear too much the /usr/local stuff. You are the true packager here, but IMHO, people compiling will play using the developer behaviour, the other are users or packagers using both /usr.
My two cents, though.

Bertram25:
Tested with latest patch. Works like a charm. :)

akien-mga:
> I wouldn't fear too much the /usr/local stuff. You are the true packager here, but IMHO, people compiling will play using the developer behaviour, the other are users or packagers using both /usr.
> My two cents, though.

You have a point. Do you think I should drop the configuration step for the desktop file and just use:

```
Exec=opendungeons
Icon=opendungeons
```

Bertram25:
If you are certain it is the common practice for desktop files, then yes I trust you upon this.

akien-mga:
I'll think I will tweak the script in order to have, when installed in /usr:

```
Exec=/usr/games/opendungeons
Icon=opendungeons
```

and when installed in /usr/local:

```
Exec=/usr/local/games/opendungeons
Icon=/usr/local/share/icons/hicolor/scalable/apps/opendungeons.svg
```

The `Exec=opendungeons` works fine on Mageia and would on Debian, but I don't trust all distros to have /usr/games in their default PATH, so it doesn't cost much to be more specific.
As pointed out earlier for the icon, I think the `Icon=opendungeons` is the best solution to have a flexible icon (with different icons depending on the size), but would not work (at least on Mageia) if installed in /usr/local.

With my proposal from above, it should work in both prefixes, with the only fallback that your incoming 16x16 and 32x32 icons might not be used on /usr/local.

Bertram25:
> With my proposal from above, it should work in both prefixes, with the only fallback that your incoming 16x16 and 32x32 icons might not be used on /usr/local.

The proposal looks more than fine, IMHO. As for the icons, well, I'm sure not a lot people will notice.
As it would be a very specific set of users, IMHO.

akien-mga:
Commit b1483c0 implements the proposal described above.

I've noticed that if you run cmake with the default prefix, and then run it again with -DCMAKE_INSTALL_PREFIX=/usr, the Exec field in the desktop file won't be reconfigured. But it's not really important IMO, we don't support successive CMake calls with variations :-)

Bertram25:
> I've noticed that if you run cmake with the default prefix, and then run it again with -DCMAKE_INSTALL_PREFIX=/usr, the Exec field in the desktop file won't be reconfigured. But it's not really important IMO, we don't support successive CMake calls with variations :-)

True. Works like a charm. :)
If you haven't anything left for doubts, I'll merge. ;)

akien-mga:
Nice :-D

akien-mga:
@Bertram25 Maybe now would be a good time to merge the beast into stable?

Bertram25:
Done! :)

PR #206, opened 2014-09-29
Make it possible to run the game out of the source dir

### LINKED ISSUE #121: Fix run in place when building OD out of source
Partly related to #58.

I'll have a look at fixing the CMake scripts to make it possible to run OD even when it is built out of the root source directory (for example I always build OD in a "build" subdirectory). For the reference, the build directory is stored in the CMAKE_BINARY_DIR variable.

On Linux, the simplest way to do this would be to add symlinks to the needed assets directories (which can be easily located using CMAKE_SOURCE_DIR).

The current implementation uses two resources.cfg files [1]: one for the system wide installation (currently broken) and one for launching OD from the source dir. The latter could also be adjusted to reference the assets directory (but last time I tried I couldn't find a way to reference them with a relative path, and Ogre3D did not like the absolute path).

[1] https://github.com/OpenDungeons/OpenDungeons/blob/development/cmake/config/resources.cfg.in

akien-mga:
Create symbolic links to the game resources on Linux when building out of the CMAKE_SOURCE_DIR (fixes #121)
Also replace CMAKE_CURRENT_BINARY_DIR by CMAKE_BINARY_DIR (we have only one CMakeLists.txt)

To test:

```
mkdir build && cd build
cmake ..
./opendungeons
```

This failed previously because of missing resources, it should work now.


akien-mga:
The symlinks create are currently absolute, so if you move your source directory (+ the build directory, keeping the same relative paths), they won't work anymore. That's not a big deal since these symlinks are just a convenience for us developers.

I've had a look at how to retrieve the relative path for the symlinks, since it would be way cleaner. It turns out there is no dedicated function for that in bash, and I couldn't find it either in CMake. The best solutions are python and perl one-liners, but I don't know if we can assume that people will always have python or perl on their system?

Taken from: http://stackoverflow.com/questions/2564634/bash-convert-absolute-path-into-relative-path-given-a-current-directory/
`$ python -c "import os.path; print os.path.relpath('/foo/bar', '/foo/baz/foo')"`
`$ perl -MFile::Spec -e "print File::Spec->abs2rel(q(/foo/bar),q(/foo/baz/foo))"`

Bertram25:
> but I don't know if we can assume that people will always have python or perl on their system?

No, IMHO we can't. AFAIK, almost every python app I know has a troubleshooting section listing python dependencies problems. Quite the same for perl, or even worse...

As you said, it is some convenience function. I wouldn't worry too much. You already made our day. :)

Bertram25:
Looks fine. I'll test this evening. :)

hwoarangmy:
> No, IMHO we can't

+1. We should avoid adding additional dependencies as much as we can. That's something I find really annoying when you want to try to compile some open source game and you end up downloading 15 dependencies...

Concerning the patch, will it work with windows ?

hwoarangmy:
Ok, I've add a look, it's only for linux. Could you copy the required directories on windows, to stay multiplatform, please ? Or in another issue if you prefer ?

akien-mga:
@hwoarangmy My commit is just an additional convenience for Linux, it doesn't change anything for Windows. I don't think I can create symbolic links on Windows, and duplicating ~90 MB of resources is not an improvement IMO :-)

If you want to support out of source builds for Windows, we could probably for the binary to be created in CMAKE_SOURCE_DIR, along with the plugins.cfg and resources.cfg?
But I'm not sure that's worth the effort: apart from developers, no Windows users will try to compile the game from source and run it. It's different on Linux where many advanced users might give it a try this way, and it's nice to be able to run ./opendungeons directly :-)

akien-mga:
Actually there might be a way to do this on Windows too.

@Bertram25, @hwoarangmy: Could you experiment with the `mklink` command to see if you can create symlinks similar to what `ln -s` does on Linux?

See: http://www.howtogeek.com/howto/16226/complete-guide-to-symbolic-links-symlinks-on-windows-or-linux/ (browse down to "Symlinks via command prompt")

`mklink /prefix link_path file/folder_path`

``````
First, choose the correct prefix.  Mklink can create several types of links, including the following:

    /D – creates a soft symbolic link, which is similar to a standard folder or file shortcut in Windows.  This is the default option, and mklink will use it if you do not enter a prefix.
    /H – creates a hard link to a file
    /J – creates a hard link to a directory or folder```
``````

**Comment by akien-mga** [comment]
Ah ah, it's good to read man pages: I discovered the `-r` option for `ln` which makes the symlinks use relative paths :-D

**Comment by akien-mga** [comment]
I'll rebase and add some tests instructions to create symlinks on windows too.

**Comment by hwoarangmy** [comment]
> Could you experiment with the mklink

it seems to work on windows 7. But not with XP (at least on my computers)

> I don't think I can create symbolic links on Windows, and duplicating ~90 MB of resources is not an improvement IMO

That's what I do at every build ;)
If it is done like on linux, by testing if the path is the same, I see no problem with that. For me, it would be an improvement ;)

**Comment by akien-mga** [comment]
@hwoarangmy It seems there was `junction -s` on WinXP:
`junction –s link_path file/folder_path`

I thought people were no longer using XP :-p
Now, is there a way to check with CMake whether the Windows version is XP or newer?

**Edit:** Ah junction is a third-party tool, so it won't work no.

**Comment by akien-mga** [comment]
@hwoarangmy, @Bertram25 Pushed a new version that should work on Windows too, and that uses relative symlinks on Linux.

**Comment by Bertram25** [comment]
Works fine to me under linux. I'll let @hwoarangmy do the windows part.

**Comment by hwoarangmy** [comment]
Thanks for adding the windows part. I should not have to copy every ressource each time I compile a new build from now on :)
I will test this tonight and merge if no problem ^^

**Comment by hwoarangmy** [comment]
I've tried the patch but the links are not created. I'm gonna make more tests tonight when I have time

**Comment by akien-mga** [comment]
@hwoarangmy It seems I was a bit optimistic about how to run a command-line instructions on Windows :-P

I found some details here: http://stackoverflow.com/questions/13959434/cmake-out-of-source-build-python-files
They use something like that:

```
        if (UNIX)
            set(command ln -s ${target} ${link})
        else()
            set(command cmd.exe /c mklink ${link} ${target})
        endif()
```

Could you try to replace the simple `mklink` call in my PR with `cmd.exe /c mklink`? And maybe check whether the flag I used after mklink is pertinent :-) (you can try this locally to see if the results of mklink commands are proper symlinks as we need here).

hwoarangmy:
From what I've seen, mklink without parameter is for files and /j is for directory so using it seems pertinent ;)
I've tried to play with cmd /c, quotes and everything but could not manage to make it work within cmake. In a comand console, it works (by adding quotes to directories).
I will try again tomorow and tell you if I find something.

akien-mga:
Maybe it's an issue with administrator rights? See http://superuser.com/questions/124679/how-do-i-create-a-link-in-windows-7-home-premium-as-a-regular-user

Of course that's only if the `cmd.exe /c mklink /j foo bar` syntax is correct and can be used within cmake.

hwoarangmy:
I could make it work only by using a batch file. I created a file "winlink.bat" in OD root containing :
mklink /j "%1" "%2"
In the makefile I replaced :
add_custom_command(...)
by
execute_process(COMMAND cmd.exe /c ${CMAKE_SOURCE_DIR}/winlink.bat ${CMAKE_BINARY_DIR}/${RES} ${CMAKE_SOURCE_DIR}/${RES})

There might be a better way to do that, through, but I'm not a cmake expert. I don't know how execute_process translates the command (when adding double quotes).

EDIT : I've just thought that there will probably be problems with paths containing spaces...

EDIT 2 : I think it would be Ok with "winlink.bat" containing :
mklink /j "%~1" "%~2"

akien-mga:
A one-liner command in cmake would be classier than having to rely on a .bat script, but if it's the only solution why not.

I've found some info about how to get some info about what's failing: http://stackoverflow.com/questions/5300572/show-execute-process-output-on-stdout

From what you're saying, I gather that this did not work?
`execute_process(COMMAND cmd.exe /c mklink /j "${CMAKE_BINARY_DIR}/${RES}" "${CMAKE_SOURCE_DIR}/${RES}")`

hwoarangmy:
As you have guessed, I've tried this and it doesn't work. I've also tried with paths containing spaces bu using a custom batch file that only prints arguments (to know what was going on). The problem is that if the path contains spaces, double quotes are added by cmake. If no space, no quotes.
That means that we cannot add manually the double quotes. Moreover, if I try the command you suggested, I see no difference by adding quotes or not (same result from batch file).

Like you, I would prefer to not use a batch file but that's the only solution I've found so far :-(

akien-mga:
You mean it doesn't work with your batch file when the build folder has spaces?

Then maybe you should let cmake handle the quotes and remove them from your batch file? Then there would be quotes only when cmake deems it necessary.

hwoarangmy:
no, it works with the batch. The problem is handling the quotes within cmake...

akien-mga:
Alright, let's go with the batch file for now :-)

What about putting the batch file in the root of the cmake/ directory? We already have lots of folders in the repo root directory, I'd like to avoid adding scripts that are not relevant for end-users or for us on a daily basis :-)

hwoarangmy:
I'm ok :)

akien-mga:
Seen a typo, will rebase again :-)

akien-mga:
@hwoarangmy @Bertram25 33f0e88 should work better :-)

akien-mga:
Note that now the symlinks should be created when running cmake, you don't have to compile everything to get them :-)

Bertram25:
Long story, but very interesting. :)
I'll test as soon as the newest build is uploading. ;)

hwoarangmy:
gonna test it ASAP. This will be usefull for me ^^

hwoarangmy:
That's good for me. As @Bertram25 is testing, I will let him merge.
I was planning on doing a script to copy the ressources. Thanks to you @akien-mga , I won't have to :)

Bertram25:
Works fine. Merging :)

PR #211, opened 2014-10-02
updated model format so ogre won't complain

oyvindln:
I managed to update the models by using OgreXMLConverter, converting to xml and back again. Seems to work fine aferwards.


akien-mga:
Works fine here, all warnings about the need to update the models are gone.
There are still a couple warnings on some meshes, e.g.:

> WARNING: the mesh 'Chicken.mesh' includes vertices without bone assignments. Those vertices will transform to wrong position when skeletal animation enabled. To eliminate this, assign at least one bone assignment per vertex on your mesh.

But these deserve their own issue so that @Danimal696 can fix it.

hwoarangmy:
Yes. Do not hesitate to open an issue listing all the meshes you have seen

Bertram25:
@oyvindln Cool! I'll test it after @akien-mga 's one. :)

Bertram25:
Works fine. The latest warning have been put in: #213

PR #220, opened 2014-10-04
Only create symlinks if they don't already exist

oyvindln:
If the second parameter to ln is a directory (or link to one) it creates a symlink inside it (unless the -T) option is specified. Changed the loop so it only creates symlinks if they don't exist to avoid geting symlinks in the symlinked direcdtories linking to the symlinked directory. (Symlinkception!)

### DIFF
--- a/source/MapLight.h
+++ b/source/MapLight.h
@@ -65,7 +65,7 @@ public:
         setAttenuation(range, constant, linear, quadratic);
     }
 
-    ~MapLight()
+    virtual ~MapLight()
     {}
 
     static const std::string MAPLIGHT_NAME_PREFIX;
--- a/source/MapLight.h
+++ b/source/MapLight.h
@@ -65,7 +65,7 @@ public:
         setAttenuation(range, constant, linear, quadratic);
     }
 
-    virtual ~MapLight()
+    ~MapLight()
     {}
 
     static const std::string MAPLIGHT_NAME_PREFIX;
@@ -85,7 +85,7 @@ public:
     void setPosition(Ogre::Real nX, Ogre::Real nY, Ogre::Real nZ);
     void setPosition(const Ogre::Vector3& nPosition);
 
-    virtual std::string getOgreNamePrefix() { return "MapLight_"; }
+    std::string getOgreNamePrefix() const { return "MapLight_"; }
 
     const std::string& getName() const
     { return mName; }
@@ -127,6 +127,12 @@ public:
     static void loadFromLine(const std::string& line, MapLight* m);
 
 private:
+    //! \brief Convinience function to get ogreNamePrefix + name
+    //! Used for nodes. The name does not include the _node and similar postfixes which are
+    //! added in RenderManager.
+    inline std::string getNodeNameWithoutPostfix() const
+    { return getOgreNamePrefix() + getName(); }
+
     GameMap* mGameMap;
 
     Ogre::Vector3 mPosition;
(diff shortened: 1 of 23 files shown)


oyvindln:
There is also another commit, I was working on something else when I noticed this issue, the application gave errors saying it was trying to load music/music/music/music/music...etc because of too many symlinks.

Bertram25:
There is a todo to remove and some tweaking in the deletion processing. :)
Afterwards, it should be ok. :)

oyvindln:
Think it should be sorted now. I also did some work on the minimap, it's no longer being cut off due to using unsigned integers for a value that could be negative. It's still a bit unsteady, I'm unsure what's causing it, may be some rounding issue.

oyvindln:
The change to vector is mostly just about avoiding manual memory management.

Bertram25:
> The change to vector is mostly just about avoiding manual memory management.

True, and the use of a vector in a vector shouldn't hurt performance as vectors try hard to store things contiguously in memory AFAIK. So I'm all fine with this change.

source/MapLight.h
```diff
@@ -65,7 +65,7 @@ class MapLight
         setAttenuation(range, constant, linear, quadratic);
     }
 
-    virtual ~MapLight()
+    ~MapLight()
```
hwoarangmy: Why removing the virtual ? From what I've seen, it displays a warning at compilation. Is it wanted ?

oyvindln: It is not inherited by anything, though I'm not sure why I messed with it.

Bertram25: Let's readd the virtual then. :)

hwoarangmy: When I do that, I get a warning saying that non virtual destructor might cause undefined behaviour. Since no class extends MapLight, it will not be a problem but IMO, we should try to avoid such warnings as much as possible...

Bertram25: > we should try to avoid such warnings as much as possible...

Yes, please. :)

oyvindln: Already sorted :)

Bertram25:
The three latest patches looks like nice cleanups to me. I'll test tonight. :)

Bertram25:
A warning! :)

```
opendungeons/source/MiniMap.cpp: In member function ‘void MiniMap::draw()’:
opendungeons/source/MiniMap.cpp:165:80: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
     for (int ii = 0, mm = mCamera_2dPosition.x - mWidth / (2 * mGrainSize); ii < mWidth; ++mm, ii += mGrainSize)
```

hwoarangmy:
I have the same warning :)
Apart from that, it works. Good job :)
And sorry for using your patience so much

Bertram25:
@oyvindln Yeah, last night, I hadn't much time. It's working great the warning apart. :)

Imho, once you has fixed the warning, this is mergeable. :)

oyvindln:
Fixed. I found that it's possible to rotate CEGUI objects, so we could use that to get a better quality map, may look at that later, but I want to get this PR merged first.

Bertram25:
> I found that it's possible to rotate CEGUI objects

I've tried that a good amount of time ago now but at the time I got weird results. I was a noob at CEGUI at the time (and still am), but if you find how to, I'm eager to know.

I'll test asap, and merge if everything is ok. :) Thanks!

Bertram25:
Clicking on minimap is once again broken! :S Sorry, this one won't go in, yet.

You'll have to rotate a quarter of turn to test.

Bertram25:
Works like a charm and no warnings. Merging :D

PR #222, opened 2014-10-05
Added the Hammock for Adventurers.

### LINKED ISSUE #218: Add new hammock bed for the Adventurer creature
Model available here:
http://forum.freegamedev.net/viewtopic.php?p=59983#p59983

Bertram25:
Fixes #218

I also updated the credits file and fixed a few things there.


akien-mga:
Looks fine. To see the hammock live, the easiest way is to start the Stonekeep Siege level, there is an AI hero team with dormitories there.

BTW I've noticed that the light effect produced by the mouse cursor does not apply on the dormitory texture (nor on the hammock), but that's something for later.

Merging :-)

Bertram25:
Thanks! :)

PR #223, opened 2014-10-07
Some work on GUI

### LINKED ISSUE #203: Adapt GUI to lower resolutions
As seen on #201, the GUI is particularly out of scale when running the lowest windowed resolutions that are 640x480 and 800x600. The game in itself runs fine, so if CEGUI permits it, it would be nice to downscale the whole GUI (both the main menu and the in-game UI) to fit those lower resolutions.

Bertram25:
I adapted the main menu and all the loading menus to resolutions up to 640x480.
Working on less wouldn't be wise without hurting higher resolutions IMHO.

In game I also moved the destroy room/traps to the left and separated it from other buttons to hint the user about its specialty.

Fixes #203

### DIFF
--- a/gui/OpenDungeonsEditorMenu.layout
+++ b/gui/OpenDungeonsEditorMenu.layout
@@ -8,17 +8,32 @@
         <Property name="FrameEnabled" value="False" />
         <Property name="BackgroundEnabled" value="False" />
         <Window type="OD/StaticText" name="LoadingText" >
-            <Property name="Area" value="{{0.5,-301},{0,672},{0.5,299},{0,692}}" />
+            <Property name="Area" value="{{0.5,-301},{0.5,240},{0.5,299},{0.5,260}}" />
             <Property name="Text" value="Loading..." />
             <Property name="MaxSize" value="{{1,0},{1,0}}" />
             <Property name="FrameEnabled" value="False" />
             <Property name="HorzFormatting" value="CentreAligned" />
             <Property name="BackgroundEnabled" value="False" />
         </Window>
+        <Window type="OD/StaticImage" name="WelcomeBanner" >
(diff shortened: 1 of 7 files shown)


akien-mga:
Great! Testing now.
+1 for moving the destroy room/traps button :-)

Bertram25:
Thanks :)

akien-mga:
It works fine, I'll let @hwoarangmy or @oyvindln have a look and merge.

It's not as great as I would envision for a long-term solution (I'd like the GUI to scale itself down when resolution < 800x600 or so, especially ingame where the minimap and goals take a lot of place at low resolutions), but it's already just fine for 0.4.9 IMO :-)
If we want something better, I guess some deep knowledge of CEGUI will be needed. I'm looking forward to playing around with CEED, but for some reason it doesn't start because some symbols are missing from my PyCEGUIOpenGLRenderer.so.

Bertram25:
> It's not as great as I would envision for a long-term solution (I'd like the GUI to scale itself down when resolution < 800x600 or so, especially ingame where the minimap and goals take a lot of place at low resolutions), but it's already just fine for 0.4.9 IMO :-)
> If we want something better, I guess some deep knowledge of CEGUI will be needed. I'm looking forward to playing around with CEED, but for some reason it doesn't start because some symbols are missing from my PyCEGUIOpenGLRenderer.so.

Feasible but not easy, indeed. Note that most windows should already resize well as I did the job there. But I didn't set a relative size to resolution as I had undesired side-effects last time.

Bertram25:
Is it ok @hwoarangmy @oyvindln ?

hwoarangmy:
Sorry cannot test again tonight. I will trust @akien-mga and merge ^^

PR #232, opened 2014-10-11
Trap/Room creation in a more objet way (avoiding void* pointers)

### LINKED ISSUE #228: Remove void* pointer when Traps are created and do the same for rooms
As discussed with @Bertram25 :

We should handle Trap like RoomObjects are handled here to avoid the need of using void\* pointers in the Trap creation :
- Allow the Trap static function that creates Traps to use the stream to allow specialized Traps to get the information it needs
- Do not use a generic function on server side when creating a trap. Create the good trap instead with a new (since it is the place where we have all the information)
- On client side, use a generic function with the ODPacket to create the correct trap like when reading the level file.

See PR #227 to see how RoomObjects are used

hwoarangmy:
- Trap/Room creation in a more objet way (avoiding void\* pointers)
- Removed ODPacket.h from header files where it is not needed
- Prepared Traps to allow trap specific parameters
- In level files, traps/rooms are not anymore identified by their mesh but by their type
- In level files, traps/rooms are separated by [Room][/Room] [Trap][/Trap]
- In the level files, rooms/traps with 0 tiles are not saved anymore
  fixes #228


hwoarangmy:
@Bertram25 BTW, I've removed the bool + string problem we have discussed about for rooms/traps creation in the process ^^

Bertram25:
Cool, there is a lot of logic change here but I guess for the better. :)

I didn't see anything wrong with the code, I'll test asap.

Bertram25:
Is fine, I'm merging :)

PR #236, opened 2014-10-13
Added the Lizardman Bed while on it.

Bertram25:


Bertram25:
I'll add the two other new beds as @Danimal696 got crazy last night ;)

Danimal696:
The chess pattern is only intended for treasury, but since previously all rooms shared the same texture (yuck :P) this one took the place of that texture. Its only temporary, i have a half finished new portal; the dungeon heart will need to wait until everything is done and for respect to elvano; but if at that time he still hasnt delivered ill make a new one.
By the way i wont touch the portal until i finish with the core rooms and beds.

hwoarangmy:
Ok. Then I've tested it and it works. Can be merged.

Issue #240, opened 2014-10-16
Empty corners on the inner-side of dirt wall tiles


akien-mga:
As shown on the following screenshots, there are empty corners in the inner-part of some wall tiles. It seems the mesh used for a dirt wall tiles (i.e. a dirt tile with tile fullness 100% and next to a non-dirt tile or tile with tile fullness 0%) has smooth corners (since it looks better), but might not be placed correctly in some cases (here a "north-south" wall relative to the original position of the camera).

![od-wall-tiles](https://cloud.githubusercontent.com/assets/4701338/4661164/4ca0f2ec-5520-11e4-9fa5-4ef67311a3a2.jpg)
![od-wall-tiles1](https://cloud.githubusercontent.com/assets/4701338/4661165/4e35ae86-5520-11e4-9f22-25b29d236792.jpg)

This also raises another question (not directly linked to this issue): should we use "smoothed dirt wall" tiles next to claimed walls? Wouldn't it make more sense (and look better) to use normal "inner wall" tiles, i.e. with no empty space between the dirt wall and the claimed wall?

Bertram25:
This bug is coming from the hard-coded way to select and rotate a tile not handling every cases properly.
It is feasible but will require some tests as breaking something else will be easy here. :)

Danimal696:
Yes, a replacement for walls and its code is on the todo list for the future

PR #244, opened 2014-10-17
New configurable creature stats  + new beds & small fixes

Bertram25:


akien-mga:
I don't think this is related to this PR, but OD is looking different for me today! The treasury tiles are a lot brighter, see this screenshot compared to the ones in #240.
od-light

Was there some recent change in OD that could explain it, or is it just because I updated to mesa 10.3.1 from mesa 10.3.0?

akien-mga:
Does your PR contain all commits from @hwoarangmy for the boulder trap?
I placed a boulder trap in the enemy base on the 2v2 test level where there is a claimed tile of the player's colour, but it did not attack the enemies. I have yet to try the trap in a more conventional way (i.e. having it wait for enemies and not drop it in the middle of 10 targets).

hwoarangmy:
> Was there some recent change in OD that could explain it

It is due to normal desactivation in #236 

> Does your PR contain all commits from @hwoarangmy for the boulder trap?

If you see the boulder trap button, yes ^^

> I placed a boulder trap in the enemy base on the 2v2 test level where there is a claimed tile of the player's colour, but it did not attack the enemies

It has a 100 turns cooldown so you have to wait more or less 1 minute before it can trigger. And it targets only on the 4 tiles around. To help you trigger it, you can use console command "setdest" to force an enemy creature to go play around :-)

Bertram25:
I've finished the base stats parameters so I've pushed it. :D 

Feel free to have a look and a try. :)

What's left is the XP table and handling for an attack warmup parameter.
@hwoarangmy Have you got any advice concerning the attack warmup parameter before I bite that one?

I won't likely be available tomorrow, so have a nice sunday. :D

hwoarangmy:
Ok, gonna try it right now. @Bertram25 nice sunday ^^
Concerning the warmup, I don't see anything special to say concerning architecture.

hwoarangmy:
I've tried to change the config so that a Kreatur had a range = 5 and dropped one on the battlefield but it attacks like a melee unit. It probably comes from GameMap::pathToBestFightingPosition which uses the weapons to find the best position to attack (at the greatest range). But since weapons are always created even when there are none (which is probably not the best thing to do), the new weaponless attack range is not used

Bertram25:
> I've tried to change the config so that a Kreatur had a range = 5 and dropped one on the battlefield but it attacks like a melee unit. It probably comes from GameMap::pathToBestFightingPosition which uses the weapons to find the best position to attack (at the greatest range).

Oops, was on my check list and somehow I forgot it. I'll have a look. Thanks for pointing out. :)

> But since weapons are always created even when there are none (which is probably not the best thing to do), the new weaponless attack range is not used

Not the best practice indeed. In one of my next PR, I'll work toward having NULL reference weapon pointer when no weapon equipped and fixed the code to behave correctly in that case. (I'll have to add missing parameters for creatures saved in level file anyway so it will be a good time to handle it)
Would this be ok for you?

hwoarangmy:
> Would this be ok for you?

Yes, it would be better. Moreover, I've seen a memory leak there (will be fixed in my next PR).
I guess it has been done because of :
https://github.com/OpenDungeons/OpenDungeons/blob/development/source/Creature.cpp#L254
But it could be better to add a boolean to know if there is a weapon. If true, we can read the weapon. If false, we know there is no weapon in the stream.

Bertram25:
The attack range bug should be fixed now. :)

> But it could be better to add a boolean to know if there is a weapon. If true, we can read the weapon. If false, we know there is no weapon in the stream.

The null check should suffice IMHO, if we sanely handle our weapon lifecycle, of course. ;) I'll do that next.

I'll be going (for real this time) ;)

hwoarangmy:
I've tried and it works as expected in code. As I said in a comment, when I equip a weapon, I would expect the creature to use it. For example, if I have a magma creature that throws magma at range 3 and I equip a sword, I would expect it to become a melee unit.
But for now, creatures always have weapons range 1 so it would mean that all units are melee unit until we fix the weapon system.
@Bertram25 what do you think ?

Bertram25:
> I've tried and it works as expected in code. As I said in a comment, when I equip a weapon, I would expect the creature to use it. For example, if I have a magma creature that throws magma at range 3 and I equip a sword, I would expect it to become a melee unit.
> But for now, creatures always have weapons range 1 so it would mean that all units are melee unit until we fix the weapon system.
> @Bertram25 what do you think ?

I think it's also a fair point to expect one creature to use the weapon, even at the cost of range.
Does this need a change?
- Use safest range when fighting at the cost of damage or equipment use? use equipment range only? use the best damages at the cost of range? ...
  What do you think guys?

@Danimal696 @akien-mga @oyvindln ?

Danimal696:
> use equipment range only

My opinion is that only naturally ranged creatures can do ranged damage, and melee remains melee always: Ex: 
warrior Rg:1 Dmg:5 -> Warrior with Uber sword Dmg:+15 Rg:1.5
Elf        Rg:5 dmg:3 -> elf with uber bow  Dmg:+15   Rg: 8
- This implicates armed warriors spawning with basic weapons (+0 to modifiers, just for look) and only improving their weapons to a more advanced of the same kind (Axe +0-> War Axe +5-> Steel axe +15 -> Epic Axe +30) so an elf remains always a bowman and a knigth a shield and sword user. Their ranges migth improve sligthly with upgrades but it will always be better than their basic attack. A value on creature.def to define what weapon they start/spawn with would be nice (knigth swordLv1 ShieldLv1; King Sword Lv5 Shield Lv5).

While weaponless monsters basic attack can have a range of 1( 2-3 if they are big) if they are melee or more if they are ranged (like the magma creature example) and never equip weapons (same range always) or have weapons just for show (non upgradable, exported in the model just like a claw or pincer is, no weapon bone or related code, just an extra mesh) in which case it is still an unarmed attack.

IMHO its more of a design choice rather than code choice to make creatures melee/ranged   weapon user/non weapon from the start, why?
- units have defined roles (knigth is a tanker, elf ranged, mage ranged-magical); 
- it makes it easier to animate, if i know a knigth will be always a sword and shield user a good animation can be done, if it can use any weapon it needs an animation for each weapon type, and for each different unit, and their moves will end up looking the same for all weapon users since the work would be huge and the skeleton forcefully shared to everyone. We currently have this problem, animations need way more character.
- If the weapon upgrade system is implemented, we will be halfway already from the start.
- The range remains the same or just improves with weapon upgrades, un-armed is always the same
- not to forget that when skills are introduced some units migth have an skill with bigger range than their basic attack, in this case a leveled up tentacle migth decide its better to stay in the back spitting acid rather than bitting in the front with the low level tentacles. But i think thats a problem for another day.

hwoarangmy:
> it makes it easier to animate, if i know a knigth will be always a sword and shield user a good animation can be done, if it can use any weapon it needs an animation for each weapon type, and for each different unit, and their moves will end up looking the same for all weapon users since the work would be huge and the skeleton forcefully shared to everyone. We currently have this problem, animations need way more character.

IMHO, that's the main problem. As you say having melee units that can use ranged weapons will give a bunch of work to the artists.

@Bertram25 Concerning your PR, it seems mergeable to me even if we might want to change the way range/attack is computed. That can be done later and we will see while playing how things can be improved. So please tell me if you want to change something or not.

Bertram25:
Hey :)

From what I understand, and without speaking about special skills, weapons should be configured to have at least the same range than the weaponless attack, and only possibly improve it, so the current behaviour should be ok, even if not yet optimal, as it won't prevent bad config.

> @Bertram25 Concerning your PR, it seems mergeable to me even if we might want to change the way range/attack is computed. That can be done later and we will see while playing how things can be improved.

Fine by me. :)

hwoarangmy:
ok. Then I merge

PR #249, opened 2014-10-20
Entities should be deleted by the gamemap on client side instead of the render manager

### LINKED ISSUE #247: Entities should be deleted by the gamemap on client side (like on the server side)

hwoarangmy:
- Entities should be deleted by the gamemap on client side instead of the render manager
- Creature reading/writing is the same as rooms/traps
- Small fix for creature sound
- Fixed a memory leak on Creature weapons
- Added a log for entity creation on client side as well as on server side
- Added [Creature][/Creature] in creature list in levels
  fixes #247

### DIFF
--- a/source/Creature.cpp
+++ b/source/Creature.cpp
@@ -71,7 +71,7 @@ static const int NB_TURN_FLEE_MAX = 5;
 
 const std::string Creature::CREATURE_PREFIX = "Creature_";
 
-Creature::Creature(GameMap* gameMap, CreatureDefinition* definition, bool forceName, const std::string& name) :
+Creature::Creature(GameMap* gameMap, CreatureDefinition* definition) :
     MovableGameEntity        (gameMap),
     mTracingCullingQuad      (NULL),
     mPhysicalAttack          (1.0),
@@ -107,14 +107,10 @@ Creature::Creature(GameMap* gameMap, CreatureDefinition* definition, bool forceN
     mStatsWindow             (NULL),
     mAttackedTile            (NULL),
     mAttackedObject          (NULL),
-    mSound                   (SoundEffectsManager::getSingleton().getCreatureClassSounds(getName())),
+    mSound                   (SoundEffectsManager::getSingleton().getCreatureClassSounds(definition->getClassName())),
     mForceAction             (forcedActionNone)
 {
-    assert(definition);
-    if(forceName)
-        setName(name);
-    else
-        setName(getGameMap()->nextUniqueNameCreature(definition->getClassName()));
+    setName(getGameMap()->nextUniqueNameCreature(definition->getClassName()));
 
     setIsOnMap(false);
 
@@ -183,7 +179,15 @@ Creature::Creature(GameMap* gameMap) :
     mSound                   (NULL),
     mForceAction             (forcedActionNone)
 {
+    setIsOnMap(false);
+
     setObjectType(GameEntity::creature);
+
+    pushAction(CreatureAction::idle);
+
+    // Note: We reset the creature to level 1 in that case.
+    setLevel(1);
+    mExp = 0.0;
 }
 
 /* Destructor is needed when removing from Quadtree */
@@ -194,6 +198,11 @@ Creature::~Creature()
         mTracingCullingQuad->entry->creature_list.remove(this);
         mTracingCullingQuad->mortuaryInsert(this);
     }
+
+    if(mWeaponL != nullptr)
+        delete mWeaponL;
+    if(mWeaponR != nullptr)
+        delete mWeaponR;
 }
 
 void Creature::createMeshLocal()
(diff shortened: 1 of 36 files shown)


source/GameEntity.cpp
```diff
@@ -57,12 +57,10 @@ void GameEntity::deleteYourself()
...
 
     mIsDeleteRequested = true;
+
     deleteYourselfLocal();
 
-    if(getGameMap()->isServerGameMap())
-    {
-        getGameMap()->queueEntityForDeletion(this);
-    }
+    getGameMap()->queueEntityForDeletion(this);
```
Bertram25, line 63: ahh now I see.

source/Creature.cpp
```diff
@@ -258,209 +265,196 @@ std::string Creature::getFormat()
...
-    is >> tempString;
+    OD_ASSERT_TRUE(is >> tempString);
+    mDefinition = getGameMap()->getClassDescription(tempString);
+    OD_ASSERT_TRUE_MSG(mDefinition != nullptr, "Definition=" + tempString);
+
+    mMaxHP = mDefinition->getMinHp();
+    setHP(mMaxHP);
+    mDigRate = mDefinition->getDigRate();
+    mClaimRate = mDefinition->getClaimRate();
+    mSound = SoundEffectsManager::getSingleton().getCreatureClassSounds(tempString);
```
Bertram25, line 311: Is that the fix for sound?

source/Creature.cpp
```diff
@@ -107,14 +107,10 @@ Creature::Creature(GameMap* gameMap, CreatureDefinition* definition, bool forceN
     mStatsWindow             (NULL),
     mAttackedTile            (NULL),
     mAttackedObject          (NULL),
-    mSound                   (SoundEffectsManager::getSingleton().getCreatureClassSounds(getName())),
+    mSound                   (SoundEffectsManager::getSingleton().getCreatureClassSounds(definition->getClassName())),
```
Bertram25, line 110: Ah yes, the definition instead of the name. Nice catch!

hwoarangmy, line 110: yes, name was not the best here. Moreover, even if it was the name, it is not set in the constructor...

Bertram25:
I've got an assert failure when quitting a map:
EDIT: Nope, I get it also while in game.

```
ERROR: Assert failed file opendungeons/source/Creature.cpp line 2432 info : 1 creature=Kobold96,tile=[49,62],Fullness=100
ERROR: Assert failed file opendungeons/source/Creature.cpp line 2432 info : 1 creature=Kobold96,tile=[49,62],Fullness=100
ERROR: Assert failed file opendungeons/source/Creature.cpp line 2432 info : 1 creature=Kobold96,tile=[49,62],Fullness=100
ERROR: Assert failed file opendungeons/source/Creature.cpp line 2432 info : 1 creature=Kobold96,tile=[49,61],Fullness=100
ERROR: Assert failed file opendungeons/source/Creature.cpp line 2432 info : 1 creature=Kobold96,tile=[49,61],Fullness=100
ERROR: Assert failed file opendungeons/source/Creature.cpp line 2432 info : 1 creature=Kobold96,tile=[49,61],Fullness=100
```

Seems harmless, though.

The creature is a random kobold, and this is triggered on a random tile position because of the fullness not being 0.

Bertram25:
I don't know whether it is your PR introducing it, though.

Bertram25:
The assert failure apart, this works fine :)

hwoarangmy:
That's not the first time I see this assert. It means that on client side, i creature (in your case, a kobold), is going through a wall. it happens when synchronisation with server is not good enough (we receive the server command too early and we start going to the new position before arriving where we were expected. Usually, the creature is not far from where it should be. But sometimes, it is enough to make it go through a wall.

If you try the empty map, force a creature to go to the top right and display the stat windows, you will see synchronisation is not perfect with server (on server side, you will see the creature arrives before on client - not huge but perceptible). That's why I tried to slow down the server. But if it is too slow, the creatures will sometimes stay idle on client side (waiting for the server) and that would be even worst...

To summarize, if you think this PR is ok, you can merge ^^

Bertram25:
Ok, thanks for the thorough explanation :)

I think I can merge. Since you're the boss in that area, maybe we might have to resync creatures when they are not in sync with the server since too long, or is it already done somehow?

hwoarangmy:
> I think I can merge. Since you're the boss in that area, maybe we might have to resync creatures when they are not in sync with the server since too long, or is it already done somehow?

Not really but IMHO, we should wait to see if there is really a need to resync all as it will produce glitches and might be complex, I think we should not go there until there is a real need.

Bertram25:
ack

hwoarangmy:
@Bertram25 FYI, I've found the reason behind the ASSERT failure:
https://github.com/OpenDungeons/OpenDungeons/blob/development/source/gamemap/GameMap.cpp#L1525

While computing the path through diggable tiles, we try to find the best way considering the creature (Kobold in this case) move speed. But on a diggable tile, fullness can be > 0, which causes the assert fail. I guess it wasn't seen before because we used the creature definition. I will change this in my next PR so that we use ground speed if the tile fullness > 0.

Bertram25:
Great! :D

PR #252, opened 2014-10-21
Added missing parameters to creatures in level files and reordered them.

### LINKED ISSUE #245: Creatures stored in level file have missing parameters
Now that #151 and #152 is almost bitten as well, the creatures stored in the level file and loaded at game start up need to be better in sync with the rest of the given stats:
This will also permit to move forward using level files as save games.

I'll thus propose to use the following parameters (already existing or not):
name
className
Current Level
Current XP
Current HP
Current Awakeness
Current Hunger
GoldToDeposit
posX
posY
posZ
seatId
weaponLname,
damage
defense
range 
weaponRname
damage
defense
range

If there is no weapon equipped, the "none" parameter will be put, and 0.0 values will be set in the corresponding parameters.
Later, the equipment extra parameters should be dropped when the equipment.cfg file is supported.
The level parameter should permit to correctly setup a creature with the corresponding stats.

Bertram25:
Fixes #245

### DIFF
--- a/config/creatures.cfg
+++ b/config/creatures.cfg
@@ -39,6 +39,7 @@
 # MagicalDef/Level    The magical defense earned at each level.
 # AttackRange         The attack range when attacking without equipment.
 # AtkRange/Level      The attack range earned at each level.
+# AttackWarmupTime    The time the creature will wait before triggering an attack, in seconds.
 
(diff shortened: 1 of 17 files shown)


hwoarangmy:
ok

PR #256, opened 2014-10-22
Added XP table support in the creatures.cfg file.

### LINKED ISSUE #152: Add lacking creatures stats
As discussed here: http://forum.freegamedev.net/viewtopic.php?p=59460#p59460

Once issue #151 is done, the creature's following stats will have to be added:
- [x] **XP table** 
  --> 30 values per creatures. Note that the creature's XP should go back to 0 at each level up as it isn't the case ATM. The XP to reach for next level should be displayed in the info window.
  (XP needed to level up, we should have 3 tables, weak creature, normal creature, strong creature; so weak monsters level up faster and stronger ones slower)
  The different XP tables should be found in the spreadsheet file. See #153 

**Stats Added:**
- [x] **Base HP** (at level 1)
- [x] **HP gain per level**
- [x] **Base physical attack**
- [x] **Physical attack gain / level**
- [x] **Base magical attack**
- [x] **Magical attack gain / level**
- [x] **Base physical defense**
- [x] **Physical defense gain / level**
- [x] **Base magical defense**
- [x] **Magical defense gain / level**
- [x] **Base attack range (in tile)**
- [x] **Attack range gain / level**

(Walk speed (on different terrains, 0 speed on one means it cant pass that terrain) -> Done)
- [x] **Walk speed gain per level** (% over base, should never go over 1.3 of base speed)
- [x] **Attack warmup.** (The time before a creature can trigger its attack.)
- [x] **HP Heal rate per turn** (if not drowsy, hungry)
- [x] **Awakeness lost per turn** (if not sleeping)
- [x] **Hunger growth per turn** (if not sleeping or eating)

**Worker's stats:**

Zero (0) means it can't work on related room/task, All have Base value and Gain per level.
- [x] Dig -> Done.
- [x] Claim -> Done.
- [x] **Dig gain/level**
- [x] **Claim gain/level**

Bertram25:
Final part and fixes #152 Whoo!

### DIFF
--- a/config/creatures.cfg
+++ b/config/creatures.cfg
@@ -2,6 +2,8 @@
 # Each Creature definition is defined with [Creatures] [/Creatures]
 # within a [Creature] [/Creature] tag couple.
 # The creatures initial stats are parsed within the [Stats] [/Stats] tag couple.
+# The creatures XP table is parsed within the [XP] [/XP] tag couple.
+# It represents the amount of Experience points needed to reach the next level.
 # Later, the [SpawnConditions], [MoodModifiers], and [Skills] tags will have to added when supported.
 #
 # Supported stats:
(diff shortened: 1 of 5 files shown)


hwoarangmy:
cool ^^
I will test tonight

Bertram25:
What do you think of this one?

hwoarangmy:
Good

PR #263, opened 2014-10-24
Add support to put equipment in a config file

### LINKED ISSUE #138: Create an equipment catalog support
This file will tell what mesh file to use (from the entity catalog file), the given attack/defense and range given by the equipment name.

Then, other files will be able to reference only the equipment name and get the rest through that.
The entity catalog file is a prerequisite to this issue: #60

hwoarangmy:
In this patch :
- Changed the damage computation so it is not needed to compute for every entity that can attack
- In the level file, we can only add the weapon name, not its definition
- Weapon is not a GameEntity anymore, just a Creature attribute
- In the level file, creature classes and equipments can be tuned (or even created)
  Fixes #138

To test it, I've added in the skirmish 2v2 test map a knight with a tuned LongSword, an adventurer with a level created staff.
I've also set the Kobold base attack to 4 instead of 1 and there is a level specific creature : BigDragon.
Note that there is a BigDragon near the player dungeon and BigDragon can spawn in the top right player portal.


akien-mga:
@Danimal696 I think you'll love what @hwoarangmy made possible :-)

od3
Adventurer with a custom staff and huge dragon :-D

Danimal696:
Having a look at it is better than just reading ;)

akien-mga:
Here you can have a better look ;-)
od4

Disclaimer: I edited the level file so that it spawns only BigDragons, don't worry :-D

akien-mga:
Nothing to say about the last two commits (except that I might try to design a real 1v1 map for 0.4.9 and our next LAN :-)

Bertram25:
> Nothing to say about the last two commits (except that I might try to design a real 1v1 map for 0.4.9 and our next LAN :-)

+1 :)

Bertram25:
@hwoarangmy Nice work. I've got a question about the vectors of pairs and a few small requests. Once this is sorted out, I think this will be mergeable. :)

hwoarangmy:
> I might try to design a real 1v1 map for 0.4.9 and our next LAN

A 2v2 real map would also be nice (or a 2v1 if you want to have a chance against me)

Bertram25:
Cool addition. :)
I've got the usual remaining question and design comments. Do you want to fix the function doc now or in another PR? If I'm correct about my assumption concerning the definition override that can be done only once, then we'll talk about whether it is sufficient afterwards.

Best regards,

hwoarangmy:
Concerning the comments, I will add them in my next PR if you don't mind

Bertram25:
> Concerning the comments, I will add them in my next PR if you don't mind

Np at all. :)

Bertram25:
The beast is fine. I'll merge now. The things to fix in the next PR (for help):
- Comment the two functions about damage and writing definitions diffs.
- Remove the useless GameMap::creatureDefinitionFilename member.

Thanks a lot for the big work on this. :)

PR #270, opened 2014-10-30
Angelscript update

oyvindln:
WARNING: Do not click on the files changed tab unless you have a computer that can handle it, it made my browser freeze for like 5 minutes.


akien-mga:
Do we really want the angelscript documentation and samples in our source tree? I wonder too whether the project files are needed?

I think we should get rid of everything that we won't make use of directly (I mean developers should check the AS documentation upstream and not in our source tree).

akien-mga:
There is an issue with the libname as you defined it:
`Linking CXX static library liblibangelscript.a`

oyvindln:
I put everything in for now, to make sure we didn't miss anything important, but I suppose some of it could be removed if we want.

oyvindln:
We might use the projects to compile it as a library later, though the manual can probably be removed.

hwoarangmy:
IMHO, 674 files are too much. We should only embed the needed (source) and let everything else on the web (AS website or OD website) but in any case, not in the source tree. Sources are already long enough to download...

Bertram25:
Thanks @oyvindln for all the work done. I'd request both a compile on Linux and Windows to work before merging that one.

I'll test this evening normally. :)

akien-mga:
It compiles fine on Linux, though with quite a few `-Wstrict-aliasing` warnings with GCC 4.9.2, and a couple of other warnings: `-Wswitch`, `-Wint-to-pointer-cast`, `-Wunused-variable`, `-Wunused-but-set-variable`

I don't slow down the compilation too much though, and if we want those fixed, I guess we'll have to work with upstream.

Bertram25:
> I don't slow down the compilation too much though, and if we want those fixed, I guess we'll have to work with upstream.

Yeah, as long it doesn't touch too much our own compilation process, I'd rather open an issue on their bug tracker instead but it seems it's impossible to add one there.
I've seen his creator answer on forums though:
http://www.gamedev.net/forum/49-angelcode/

I've thus created a topic but could you give me the compilation warning you have (even as a gist) and I'll put them here: http://www.gamedev.net/topic/662413-compiler-warnings-with-latest-stable-version/

akien-mga:
Here is my compilation log for angelscript: http://paste.fedoraproject.org/146507/41467991/

Bertram25:
@akien-mga Note that strict aliasing warnings cannot be avoided AFAIK:
http://www.angelcode.com/angelscript/sdk/docs/manual/doc_compile_lib.html

I'll add the other warnings there.

Bertram25:
I've added them.

Bertram25:
Answer from Andreas Jonsson;

```
Hi Bertram,

Glad to see you're still using AngelScript. I visit the OpenDungeons page once in a while to check on the progress.

I'll look into the compiler warnings, thanks for letting me know. None of them are any cause of errors though.

Regards,
Andreas
```

hwoarangmy:
As @akien-mga , I confirm a few more warnings but it compiles with MinGW and the game launches. Is there a way to test something ?

Bertram25:
> Is there a way to test something ?

Atm, only certain console commands could be linked to AS AFAIK. As seen in the scripts, the simplest function you can try is 'fps'.

Bertram25:
To me, it's fully working. :)
@hwoarangmy If at least the fps command is working for you, I think you can merge.

oyvindln:
Yeah, fps command is currently implemented in the angelscript file. Worked fine for me as well. 

I think we do have to rework the scripting somewhat though, right now the wrapper just gives access to most of the singleton classes and let's the script do whatever with it. This predates the change to using the server-client model for everything, so doing changes in game is not going to work. I think for map scripting we should rather send in some context variable to the script and have a limited set of functions for doing things. As for the console functions, I still don't understand the point of having the functions in angelscript as opposed to in the code, the console doesn't actually allow any scripting, it just have some functions implemented in angelscript as opposed to native code.

Bertram25:
> I think we do have to rework the scripting somewhat though, right now the wrapper just gives access to most of the singleton classes and let's the script do whatever with it. This predates the change to using the server-client model for everything, so doing changes in game is not going to work.

I agree.

> I think for map scripting we should rather send in some context variable to the script and have a limited set of functions for doing things.

I agree as well, while I didn't seriously think of any use case so far.

>  I still don't understand the point of having the functions in angelscript as opposed to in the code, the console doesn't actually allow any scripting, it just have some functions implemented in angelscript as opposed to native code.

IIRC, we already discussed that and agreed AS wasn't fitting console commands, at least as a primary way to implement them. I wouldn't remove the commands right now, though. As they are our only way to test AS atm.
Once we've added a few actually useful scripting functions, yeah, let's remove the scripted commands.

hwoarangmy:
IMHO, scripts and console are used in different cases. console is good to debug by forcing something for example.
Scripts are good to add effects in a levels.

hwoarangmy:
I have a segfault when I enter an invalid command. I've looked and it seems that:
https://github.com/oyvindln/OpenDungeons/blob/development/source/scripting/ASWrapper.cpp#L407

mStringArray is not valid when this function is called. From what I've seen, this variable is not used anywhere else in OD code so I would guess it is AS engine that released it.

Bertram25:
IMHO, this is another issue not linked to the current upgrade. We will want to get rid of most AS command anyway. If you don't see anything else, IMHO you can merge. :)

oyvindln:
That's weird, I don't see why it would be released before the angescript object was destroyed.

What command did you try? I am not getting a segfault on invalid commands.

Bertram25:
> What command did you try? I am not getting a segfault on invalid commands.

Me neither. ?

hwoarangmy:
First, I made a typo on "help". I think it was "hel"
And I've tried "pop" and I get a segfault, too.

Concerning merging, this problem is probably linked to this PR (might be a change in the way AS is supposed to be used) so I think we should try to go deeper to make sure there is no bigger problem hidden.

oyvindln:
I will look at it more tomorrow, maybe there is some condition causing the function to be run without the ASWrapper object being initialised properly.

hwoarangmy:
Since I seem to be the only one to get this segfault, I will try to get more information...

akien-mga:
I can't reproduce the segfault either.

hwoarangmy:
I confirm the problem. I've added a breakpoint to see what happens to ASWrapper::mStringArray and, as supspected, it is released when calling mBuilder->BuildModule();

The callstack is:
0  0x00573742 in asIObjectType::~asIObjectType (this=0x20509ab0, __in_chrg=<optimized out>) at C:/devMinGW/od-gho-src/dependencies/angelscript/angelscript/source/../include/angelscript.h:1048
1  0x004db78e in asCObjectType::~asCObjectType (this=0x20509ab0, __in_chrg=<optimized out>) at C:\devMinGW\od-gho-src\dependencies\angelscript\angelscript\source\as_objecttype.cpp:331
2  0x004b69cf in asCScriptEngine::RemoveTemplateInstanceType (this=0x204f6778, t=0x20509ab0) at C:\devMinGW\od-gho-src\dependencies\angelscript\angelscript\source\as_scriptengine.cpp:3462
3  0x004ad3ed in asCScriptEngine::ClearUnusedTypes (this=0x204f6778) at C:\devMinGW\od-gho-src\dependencies\angelscript\angelscript\source\as_scriptengine.cpp:1290
4  0x004df398 in asCGarbageCollector::GarbageCollect (this=0x204f73f4, flags=1, iterations=1) at C:\devMinGW\od-gho-src\dependencies\angelscript\angelscript\source\as_gc.cpp:228
5  0x004b992e in asCScriptEngine::GarbageCollect (this=0x204f6778, flags=1, iterations=1) at C:\devMinGW\od-gho-src\dependencies\angelscript\angelscript\source\as_scriptengine.cpp:4563
6  0x004ac1c3 in asCScriptEngine::CleanupAfterDiscardModule (this=0x204f6778) at C:\devMinGW\od-gho-src\dependencies\angelscript\angelscript\source\as_scriptengine.cpp:825
7  0x00506e6a in asCModule::InternalReset (this=0x2050b970) at C:\devMinGW\od-gho-src\dependencies\angelscript\angelscript\source\as_module.cpp:517
8  0x00506313 in asCModule::Build (this=0x2050b970) at C:\devMinGW\od-gho-src\dependencies\angelscript\angelscript\source\as_module.cpp:266
9  0x004bfc49 in CScriptBuilder::Build (this=0x204fc770) at C:\devMinGW\od-gho-src\dependencies\angelscript\add_on\scriptbuilder\scriptbuilder.cpp:501
10 0x004be227 in CScriptBuilder::BuildModule (this=0x204fc770) at C:\devMinGW\od-gho-src\dependencies\angelscript\add_on\scriptbuilder\scriptbuilder.cpp:105
11 0x00498422 in ASWrapper::ASWrapper (this=0x20570590) at C:\devMinGW\od-gho-src\source\scripting\ASWrapper.cpp:90
12 0x0045c3fc in ModeManager::ModeManager (this=0x20570350, renderWindow=0x1493d7b0) at C:\devMinGW\od-gho-src\source\modes\ModeManager.cpp:57
13 0x00482fba in ODFrameListener::ODFrameListener (this=0x1eb6a948, renderWindow=0x1493d7b0, overLaySystem=0x15031370) at C:\devMinGW\od-gho-src\source\render\ODFrameListener.cpp:99
14 0x004a88cc in _fu364___ZN4Ogre11ColourValue5WhiteE () at C:\devMinGW\od-gho-src\source\ODApplication.cpp:122
15 0x004a98a3 in WinMain@16 (hInst=0x400000, strCmdLine=strCmdLine@entry=0x99551ee "") at C:\devMinGW\od-gho-src\source\main.cpp:82

So, it is not valid when we call ASWrapper::executeConsoleCommand.

As a naive try, I've moved:
mStringArray = mEngine->GetObjectTypeById(mEngine->GetTypeIdByDecl("string[]"));
after
mBuilder->BuildModule();

and It seems to fix the segfault. But I've not checked in AS documentation if it is the way it should be used or not ^^
You are more aware than me about AS so I will trust your judgement.

oyvindln:
Seems fair to me, I've added the change.

Bertram25:
Still no crash after change, and the commands are working. So I'm ok for a merge if you are @hwoarangmy 
Thanks @oyvindln :)

hwoarangmy:
I confirm there is no more segfault. But when I try "fps 10", the console displays:
AS (0, 0) : ERR :
Failed in call to function prepare with 'null' (Code: '-6')

So I guess that for some reason, mBuilder->GetModule()->GetFunctionByDecl("void executeConsoleCommand(string &in, string[] &in)") returns null.

Since it is the first time I try AS, it might not be related to this PR, though.
I will test on an older version to see if it is the same

EDIT: I confirm it is the same as before. So, I merge as it is ok for you. But it seems like it is broken with MinGW/Windows...

PR #281, opened 2014-11-03
Improved the way creatures see and seek claimable and diggable tiles.

Bertram25:
This fixed the small pathfinding glitches for me and improves already a lot the efficiency of workers. :)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -286,6 +286,7 @@ set(OD_SOURCEFILES
 #### Find packages ###############
 ##################################
 
+find_package(Threads REQUIRED)
 find_package(OIS REQUIRED)
 find_package(OGRE REQUIRED)
 find_package(CEGUI REQUIRED)
@@ -396,6 +397,9 @@ add_executable(${PROJECT_BINARY_NAME} WIN32 ${OD_SOURCEFILES})
 ##################################
 #### Link libraries ##############
 ##################################
+
+# Add threads when linking AngelScript
+target_link_libraries (${AS_LIBRARY_NAME} ${CMAKE_THREAD_LIBS_INIT})
 
 # Link angelscript
 target_link_libraries(${PROJECT_BINARY_NAME} ${AS_LIBRARY_NAME})
--- a/README
+++ b/README
@@ -13,13 +13,16 @@ Data files :
 ------------
 
 config/          - Several game config files
+dist/            - Icons and linux desktop entry file.
 gui/             - CEGUI files + corresponding Gui images
+levels/          - Game levels
+licenses/        - License files used for game data and code
 materials/       - Materials (models texturing scripts and textures)
 models/          - Model files
 music/           - Music files
-sounds/          - Game Sounds
-levels/          - Game levels
+packs/           - Ogre files useful to run the starting pop-up
 scripts/         - Our AngelScript code files
+sounds/          - Game Sounds
 
 Code files:
 -----------
@@ -29,7 +32,6 @@ cmake/           - Helper files for CMake
  |- modules/     - Addon scripts for CMake to find dependencies
 dependencies/    - All external projects
  |- angelscript/ - AngelScript SDK, our scripting language
-documentation/   - Misc. files (incl. a Doxyfile and sketches)
 sources/         - All our own .cpp and .h files of the game
 .gitignore       - The files and folders that are ignored by git locally
 CMakeLists.txt   - CMake script for generating the Makefile and IDE projects
--- a/source/ODApplication.cpp
+++ b/source/ODApplication.cpp
@@ -51,8 +51,9 @@
 #include <fstream>
 
 ODApplication::ODApplication() :
-    mRoot(NULL),
-    mWindow(NULL)
+    mRoot(nullptr),
+    mWindow(nullptr),
+    mInitialized(false)
 {
     try
     {
@@ -123,6 +124,8 @@ ODApplication::ODApplication() :
         mRoot->addFrameListener(mFrameListener);
 
         mRoot->startRendering();
+
+        mInitialized = true;
     }
     catch(const Ogre::Exception& e)
     {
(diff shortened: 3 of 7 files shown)


akien-mga:
I'm getting a linking issue with your PR, though I can't find a reason why:

```
Linking CXX executable opendungeons
/usr/bin/ld: libangelscript.a(as_thread.cpp.o): undefined reference to symbol 'pthread_rwlock_wrlock@@GLIBC_2.2.5'
/lib64/libpthread.so.0: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
CMakeFiles/opendungeons.dir/build.make:2642: recipe for target 'opendungeons' failed
make[2]: *** [opendungeons] Error 1
CMakeFiles/Makefile2:95: recipe for target 'CMakeFiles/opendungeons.dir/all' failed
make[1]: *** [CMakeFiles/opendungeons.dir/all] Error 2
Makefile:137: recipe for target 'all' failed
make: *** [all] Error 2
```

I've tried to build again @oyvindln's branch, and it works fine. Will try to build development.

oyvindln:
Hm, sounds like angelscript may require linking with the thread libraries.

Bertram25:
> Hm, sounds like angelscript may require linking with the thread libraries.

Then, why on the same machine, your branch is fine while mine isn't? I thought both our branches had the AS update and pthread removal?

akien-mga:
Same issue in OpenDungeons/development

Bertram25:
@akien-mga Sorry for the question but can you clear both your cmake cache and CMakeFiles/ temp folder and try again? (Or maybe you already did it?)

akien-mga:
I did it for @Bertram25's branch and for OpenDungeons/development. I don't remember for @oyvindln's, I'll try.

akien-mga:
Ah it seems I wasn't fully up-to-date on @oyvindln's branch, so one of the his later commits might explain the issue.

akien-mga:
Maybe this? https://github.com/OpenDungeons/OpenDungeons/commit/be5a2db3a086566cc76a07af09f4f5eac06878d4#diff-af3b638bc2a3e6c650974192a53c7291L475

Bertram25:
Taken from the CMakeLists.txt from the library latest version download, those lines clearly shows a need for thread:

``` cmake
find_package(Threads)
target_link_libraries(Angelscript ${CMAKE_THREAD_LIBS_INIT})
```

Maybe we can use boost::thread instead of pthread?

Bertram25:
Update:
http://stackoverflow.com/questions/5395309/cmake-and-threads
I wonder whether liangelscript.a would compile if we force the cmake module not to search for pthread and invoke the find_package statement, such as:

``` cmake
set(CMAKE_THREAD_PREFER_PTHREAD FALSE)
find_package(Threads REQUIRED)
target_link_libraries (${AS_LIBRARY_NAME} ${CMAKE_THREAD_LIBS_INIT})
```

Maybe boost must have to readded in that case?

oyvindln:
Well we could just readd linking to pthreads on linux.

Bertram25:
@oyvindln Well, true. I just wondered whether we could still get rid of it.
Btw, the same issue will probably arise on windows, right?
I guess pthreads should be reinstated in that case.
@hwoarangmy Can you tell us whether AS still compile fine for you on development?

oyvindln:
It won't be an issue on windows, it's not used by anything on windows any more.

Bertram25:
@oyvindln ? Then, there is something I don't get. Angel Script is requiring threads on linux but not on windows?

oyvindln:
It uses the native thread libraries on each system, pthreads is the native thread library on linux, but not on windows. The extra library we used was just an implementation of pthreads on top of the windows thread libraries.

Bertram25:
> It uses the native thread libraries on each system, pthreads is the native thread library on linux, but not on windows. The extra library we used was just an implementation of pthreads on top of the windows thread libraries. 

Ah of course, silly me, pthreads mean posix threads. It makes sense now. Thanks for helping an old fool lurking in the middle of his own-made desert. XD
(I don't know why, I was truely thinking pthreads were some custom made thread library, without remembering about the posix ones.)
Then, the issue is about simply readding a few lines in the CMakeLists.txt file.

@akien-mga Could you try readding those lines (where relevant) and test?

``` cmake
set(CMAKE_THREAD_PREFER_PTHREAD TRUE)
find_package(Threads REQUIRED)
target_link_libraries (${AS_LIBRARY_NAME} ${CMAKE_THREAD_LIBS_INIT})
```

akien-mga:
@Bertram25 Adding you code snippet fixes the issue. I'll let you see where it is most relevant to add it, and maybe if it should be UNIX-only (I guess the find_package(Threads REQUIRED) might be relevant for Windows too, but I wonder about the CMAKE_THREAD_PREFER_PTHREAD.

oyvindln:
CMAKE_THREAD_PREFER_PTHREAD is at least not relevant on windows. Not sure if it should be used on linux either.

Bertram25:
@akien-mga If checking for the existence of thread support is also relevant for windows. Then, indeed we shouldn't make cmake prefer pthreads in that case.

Bertram25:
@ everyone: I'm making a patch. If both you and @hwoarangmy tell me it is fixing the issue (or at least not breakig anything), then we'll be out of trouble. :)

oyvindln:
CMAKE_THREAD_PREFER_PTHREAD seems to be only relevant on IRIX and HP-UX, so I don't think it's needed.

Bertram25:
Ok, I've added the change. @oyvindln Thanks for bearing with my sometimes surprising stupidity. ;)

oyvindln:
@Bertram25 No problem

hwoarangmy:
I will try that tonight

hwoarangmy:
I've tested the PR with MinGW (before the last commit) and it works (I could launch a level successfully).

hwoarangmy:
The crash problem I was talking about happens when the plugin.cfg file is not good (or the renderer is missing). That should not happen that often and is much less important than cancelling ogre window.
I merge

Bertram25:
I may handle it the next pr if you want. That won't be difficult to add. :)

PR #297, opened 2014-11-06
Unit test stuff

oyvindln:
Added some components used to set up unit testing. BoostTestTargets is from here: https://github.com/rpavlik/cmake-modules
It made setting it up a little easier, but we'll see if we keep using it or not. The test case for the random function was just to see if it worked, used it because it didn't have any other dependencies elsewhere in the code.

### DIFF
--- a/cmake/modules/BoostTestTargets/BoostTestTargets.cmake
+++ b/cmake/modules/BoostTestTargets/BoostTestTargets.cmake
@@ -1,0 +1,274 @@
+# - Add tests using boost::test
+#
+# Add this line to your test files in place of including a basic boost test header:
+#  #include <BoostTestTargetConfig.h>
+#
+# If you cannot do that and must use the included form for a given test,
+# include the line
+#  // OVERRIDE_BOOST_TEST_INCLUDED_WARNING
+# in the same file with the boost test include.
+#
+#  include(BoostTestTargets)
+#  add_boost_test(<testdriver_name> SOURCES <source1> [<more sources...>]
+#   [FAIL_REGULAR_EXPRESSION <additional fail regex>]
+#   [LAUNCHER <generic launcher script>]
+#   [LIBRARIES <library> [<library>...]]
+#   [RESOURCES <resource> [<resource>...]]
+#   [TESTS <testcasename> [<testcasename>...]])
+#
+#  If for some reason you need access to the executable target created,
+#  it can be found in ${${testdriver_name}_TARGET_NAME} as specified when
+#  you called add_boost_test
+#
+# Requires CMake 2.6 or newer (uses the 'function' command)
+#
+# Requires:
+# 	GetForceIncludeDefinitions
+# 	CopyResourcesToBuildTree
+#
+# Original Author:
+# 2009-2010 Ryan Pavlik <rpavlik@iastate.edu> <abiryan@ryand.net>
+# http://academic.cleardefinition.com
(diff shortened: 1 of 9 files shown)


Bertram25:
I didn't know Ryan Pavlik had made cmake modules while I'm not surprised. ;)

What kind of unit tests have you got in mind, for instance?
I also wondered: Wouldn't you want to add a new cmake target, like say, `unit_tests`, so we can do a `make unit_tests` to specifically make them (and not each time we compile)?

oyvindln:
Well my plan was to use them for new code I make.

`make test` runs unit tests. I would assume it builds if they aren't already built too.
If you don't want make them you can simply turn off BUILD_TESTING in the cmake, though the whole point of unit tests is to make sure the tested functionality does what it's supposed to do, so it makes sense to have them automatically built.

Bertram25:
> Well my plan was to use them for new code I make.

ok.

> If you don't want make them you can simply turn off BUILD_TESTING in the cmake, though the whole point of unit tests is to make sure the tested functionality does what it's supposed to do, so it makes sense to have them automatically built.

My point is that the default behaviour will have to disable those, since the default build will be for casual testers and fresh-meat seekers.
I mean as developers, we should be the ones enabling options, and not make the "standard" users disabling them. :)

oyvindln:
Then I think the easiest would be to add a custom option that is off by default.

Bertram25:
IMHO, nothing to add. The feat is ready. :)
I'll test asap and merge if there is nothing else.
Maybe @hwoarangmy will want to try on windows just in case?

hwoarangmy:
Yes, I will give a try tonight and I will also have a look at the boost thread linking with MinGW.

cmake/modules/BoostTestTargets/BoostTestTargets.cmake
```diff
@@ -0,0 +1,274 @@
...
+			target_link_libraries(${_target_name} ${LIBRARIES})
+		endif()
+
+		if(RESOURCES)
+			set_property(TARGET ${_target_name} PROPERTY RESOURCE ${RESOURCES})
+			copy_resources_to_build_tree(${_target_name})
+		endif()
+
+		if(NOT Boost_TEST_FLAGS)
+#			set(Boost_TEST_FLAGS --catch_system_error=yes --output_format=XML)
```
hwoarangmy, line 227: Why did you comment this line ? Could you please delete it if not relevant or add some comment if it can be of some use ?

cmake/modules/BoostTestTargets/GetForceIncludeDefinitions.cmake
```diff
@@ -0,0 +1,44 @@
+# - Get the platform-appropriate flags to add to force inclusion of a file
```
hwoarangmy, line 1: Interesting. Could re-used to make memory tests and many stuff...

cmake/modules/BoostTestTargets/BoostTestTargets.cmake
oyvindln, line 227: It's not my code, the boostTestTargets stuff is from https://github.com/rpavlik/cmake-modules .

hwoarangmy, line 227: ok. Nevermind then

hwoarangmy:
I've compiled with and without the flag set and it works well. When enabling test, I get the executable as expected. And when I launch it, the test is ok ^^

Issue #230, opened 2014-10-09
Add support for default spawn pools and support to optionally override them in levels per seat.


Bertram25:
Prerequisite: #150 

The default spawn pools will be defined per faction in a config/factions.cfg file:

Example:

```
[Factions]
    [Faction]
        Name Keeper
        [Spawn]
        Dragon
        Wyvern
       ...
        [/Spawn]
    [/Faction]
    [Faction]
        Name Hero
        [Spawn]
        Gnome
        Adventurer
       ...
        [/Spawn]
    [/Faction]
[/Factions]
```

~~This will permit to define default spawn pools per faction, that can still optionally be overridden in the level file within SpawnPool tags.~~
=> We will see later if we want to open an issue for allowing editing spawn pools and how.

Thus, the current Faction in the level files will have to be split into two: AI type (can be human) and Faction.

Also, it would be nice when choosing a level to optionally have the right to define one's faction. This could lead to interesting replayability. Later also, it could be cool to let the server define the human/AI slots also. :)

Feel free to comment. :)

hwoarangmy:
> it would be nice when choosing a level to optionally have the right to define one's faction

In the seat definition, we could have "Keeper", "Hero", "Choice". And, for now, if it is "Choice", we assume it is Keeper. This way, when we will have the GUI that allows to choose, it would be easy ^^

> it could be cool to let the server define the human/AI slots also

Right ^^ Again, it is just a mater of GUI :)

hwoarangmy:
While doing the faction stuff, I'm trying to set up the menu to configure seats (to associate seats to players). But I'm loosing patience while trying to set up cegui combo boxes. has anyone ever tried it ? The list is empty no matter what I've tried...

@Bertram25 an idea ?

In the layout I've added a "OD/Combobox" and in code, I used the addItem function to add items...

Bertram25:
> In the layout I've added a "OD/Combobox" and in code, I used the addItem function to add items...

IMHO, you doing nothing wrong so far. The combo boxes might be broken without anyone aware of it.
That's a bummer really. What I've found so far but without any actual clues:
http://delta3d.org/forum/viewtopic.php?showtopic=19760
For self reference: http://cegui.org.uk/wiki/Sample_code_for_all_Widgets#Combobox

I don't know what code you've used to create the combo box maybe but have a go at using the one in the second link, just in case.
[...]

PR #298, opened 2014-11-06
Allow to choose faction before launching a game + configurable spawn pool per faction

hwoarangmy:
I added a gui to configure seats before launching a game. You can choose faction and set the players on any available seat. Only the server can setup the seats.

In the level, for each seat, you can choose "Choice" to allow the player to choose or you can force one of the available factions (defined in factions.cfg).
For player type, you can choose between "Choice", "Human", "AI" or "Inactive" (pretty explicit)

Note that since the spawn pool is now defined in factions.cfg, I've removed them from all the level files.

For @akien-mga , since I know you like to hunt bugs, I've tried to make the seat configuration as final as I could. That means that during seat configuration, players can be added/can disconnect. In this case they should be added/removed from the combos.
Moreover, I've added network communication so that the clients see in real time how the server is configuring the seats.

Note that there can be more players connected than seats available or used. In any case, connected players that were not chosen for a game will get disconnected and return to main menu.

Note that in the skirmish empty level, you can launch the game with only AI players. In this case, like any not chosen player, you will get disconnected and return to main menu ^^

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -222,6 +222,7 @@ set(OD_SOURCEFILES
     ${SRC}/modes/GameMode.cpp
     ${SRC}/modes/InputManager.cpp
     ${SRC}/modes/MenuMode.cpp
+    ${SRC}/modes/MenuModeConfigureSeats.cpp
     ${SRC}/modes/MenuModeEditor.cpp
     ${SRC}/modes/MenuModeMultiplayerClient.cpp
     ${SRC}/modes/MenuModeMultiplayerServer.cpp
--- a/config/factions.cfg
+++ b/config/factions.cfg
@@ -1,0 +1,40 @@
+# Factions definition file.
+# It defines which creatures can spawn through the portal per faction.
+# Each Faction is defined with [Factions] [/Factions]
+# within a [Faction] [/Faction] tag couple.
+
+[Factions]
+[Faction]
+    Name	Keeper
+    [SpawnPool]
+    ConstructWorker
+    CaveHornet
+    Wyvern
+    Dragon
+    Orc
+    Kreatur
+    Goblin
+    PitDemon
+    Spider
+    Tentacle1
+    Tentacle2
+    Troll
+    LizardMan
+    Scarab
+    Slime
+    [/SpawnPool]
+[/Faction]
+[Faction]
+    Name	Hero
+    [SpawnPool]
+    Gnome
+    Knight
+    BigKnight
+    Adventurer
+    Dwarf1
+    Dwarf2
+    Dwarf3
+    Wizard
+    [/SpawnPool]
+[/Faction]
+[/Factions]
--- a/config/global.cfg
+++ b/config/global.cfg
@@ -64,6 +64,10 @@
     Type	SpawnConditions
     Filename	spawnconditions.cfg
 [/ConfigFile]
+[ConfigFile]
+    Type	Factions
+    Filename	factions.cfg
+[/ConfigFile]
 [/ConfigFiles]
 [GameConfig]
     NetworkPort	31222
--- a/gui/OpenDungeonsMenuConfigureSeats.layout
+++ b/gui/OpenDungeonsMenuConfigureSeats.layout
@@ -1,0 +1,58 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<GUILayout version="4" >
+    <Window type="OD/StaticImage" name="Background" >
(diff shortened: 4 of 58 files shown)


Bertram25:
Well, ... wow. I just can't wait to test this change.
Not only for the features, but the extra care about code polishing can be really felt here.
This must have been quite some work. So congrats already. :)
I'll test asap.

hwoarangmy:
FYI, I've seen a bug in player faction choice. I've used the index in the combo instead of the index in the faction list (stupid me). That means that if you force a faction to hero, it will be a keeper ^^
I will fix it tonight with the other things you've seen.

Bertram25:
Cool :)
Some friends will come this week-end and they are both DK lovers. Time for a show... and some creature's killing. ;)

hwoarangmy:
Please also note that I've not implemented anything to allow changing the spawn pool in the level file. We should decide how do we want to do that before because from now on, we cannot assume what the spawn pool will be (Hero or Keeper). For this reason, we can group hero and keeper creature (like PitDemon = BigKnight) and allow disabling them by adding a tag [DisableTier4a] for example. Or we might want say that disabling rooms will be enough to automatically disable creatures (as long as we consider that hero and keeper do have the same rooms - even if one day they should have different graphics) ...

Bertram25:
@hwoarangmy About what you just say. On the design level, we haven't even decided strongly to make Heroes an actual faction with proper rooms and all. IMHO, the Hero side must be discussed from the ground up. Or we could drift badly in term of design.
That said, having a true hero faction with custom rooms looks cool, while I wouldn't like it to be 1:1 in term of skills/rooms/spawn conditions, ...

Let's discuss that part on the forums?
EDIT: The question of whether spawn pools should be editable per level will have to be discussed as well, indeed.

hwoarangmy:
I talked about spawn pool edition because it is part of #230. But IMHO (and seems like yours), we should decide this on the forum before. In any case, it is not part of this PR and, I think we should close the issue when this PR will be integrated. If it is decided we want to allow spawn pool changes within level file, we can always open a new one.

Bertram25:
> I think we should close the issue when this PR will be integrated. If it is decided we want to allow spawn pool changes within level file, we can always open a new one.

Agreed.

akien-mga:
Guys, OD becomes gorgeous! Can't wait to try this out live instead of with four local instances of the game :-P Should we try a LAN sometime this week-end?

od-lobby

hwoarangmy:
Everything should be ok now ^^

> Should we try a LAN sometime this week-end?

Usually, I can be available after 22h (when the childs sleep and I've eaten) ^^

akien-mga:
I've noticed that one can't set an AI for "Sead id 1" (at least in the 2v2 map).

Bertram25:
> Should we try a LAN sometime this week-end?

I'll focus on fixing my PR, test the other ones and why not fix the creature behaviour tonight.
Hopefully, we can merge this before playing a bit. ;)
I'll definitely won't be available saturday night, but I will be there on sunday at 22pm, for a good fight. ;P

Bertram25:
@akien-mga I definitely need to test myself, but what happens if you select the same player nickname several times?
What happens when two players have the same nickname?

akien-mga:
>  I'll definitely won't be available saturday night, but I will be there on sunday at 22pm, for a good fight. ;P

Sunday at 22 pm should be fine for me too.

> I definitely need to test myself, but what happens if you select the same player nickname several times?
> What happens when two players have the same nickname?

I haven't done extensive testing yet, but it seems solid. You can select the same player twice, it will be unassigned from its first slot when you add it into another slot.

hwoarangmy:
> I've noticed that one can't set an AI for "Sead id 1" (at least in the 2v2 map).

This is normal. It is because Seat id=1 player type is defined as "Human" where the others are defined as "Choice"

@Bertram25 I don't want to talk for @akien-mga but the expected behaviour is:

> what happens if you select the same player nickname several times?

it is removed from the combo where it was (the old one becomes empty)

> What happens when two players have the same nickname?

They will be displayed twice with same nick but it should work (players are identified by their id, not their nick)

akien-mga:
From my first tests, and since @Bertram25 reviewed the code, I'd say this is good to merge. There might probably be some issues, but they can be fixed a posteriori :-)

hwoarangmy:
> There might probably be some issues

No issues on my PR !!!! ;-)

Bertram25:
Works just great, I didn't see any bugs while testing both sides and using multiplayer or single player. I'm merging...

PR #316, opened 2014-11-12
When a player is rejected from the lobby, game should start normally for the seated players + allow clicking fighter button more than once per turn

hwoarangmy:


Bertram25:
I tested both features, both are working fine. :) 

Works fine for the button and rejecting a player in the lobby works fine. I even tested when a player disconnected during the game and everything still worked fine for the clients left.

Also, nothing to say about the code, it's quite straight-forward. Can be merged IMHO.

akien-mga:
Both features work fine locally (tested a game with three clients, one disconnected, I had to select an AI in place of the disconnected client so that I could launch the game).

PR #317, opened 2014-11-13
CMake linux option not to grab keyboard + Added support to build a crypt

Bertram25:


akien-mga:
@Danimal696 The KnightCoffin.png and KnightStatus.png textures are exactly the same pictures. I wondered whether it would be possible to use the same texture file for both meshes?
I know it's nit-picking and the file is only 200 kB, but we do have some textures that are quite heavy (~15 MB for the lava golem), so if there's a way to avoid duplicates, which should go for it :-)

Danimal696:
yes,ill merge textures: its the lava golem even used?

akien-mga:
> its the lava golem even used?

No idea about the lava golem, I don't think we're using it yet. I'm not a pro in texturing but it looks like there's something funny with them, the alpha layer (usually shown as a checkboard) seems to be _part of the picture_ and it changes in the different lighting components (diffuse, emit, normal, spec). This is definitely worth checking out because those pics are 2048x2048 and are quite heavy (and probably broken if the alpha layer is not an alpha layer but a real checkboard texture).

akien-mga:
> Moreover, this doesn't seem to work, the output of message(STATUS ${OD_NO_KEYBOARD_GRAB}) is empty after this declaration. It seems it should be explicitly set to ON, but again that's already the value of OD_LINUX_NO_KEYBOARD_GRAB, so let's use this one.

Actually shouldn't the OD_NO_KEYBOARD_GRAB (or OD_LINUX_NO_KEYBOARD_GRAB if we drop the intermediate assignment) variable be passed as an argument in CMAKE_CXX_FLAGS?
It doesn't work here if I set OD_NO_KEYBOARD_GRAB to ON in CMakeLists.txt, nor when I remove the intermediate variable and use OD_LINUX_NO_KEYBOARD_GRAB directly in InputManager.cpp.

But it does work if I add it to the CMAKE_CXX_FLAG myself.

Danimal696:
there is neither skeleton or mesh for the lava golem; this is the kind of stuff the old team loved to do, half done things laying everywhere; please erase all things related to lava golem, in due time we will reskin the current troll to be fire or ice flavour

akien-mga:
About the texture stuff (lava golem and co.), I guess it would be better to discuss this in a separate issue: #318.

Bertram25:
> please erase all things related to lava golem, in due time we will reskin the current troll to be fire or ice flavour

So we'd remove the lava golem completely? Your last word, M. Foucault?
EDIT: For the record, this above is a reference to the French "Who wants to be a billionaire?" ;)

akien-mga:
IMO we should keep such stuff in the media repo, but remove everything that we haven't implemented yet from the main repo (apart maybe from stuff that we plan on implementing, such as the keeper hand).

Bertram25:
> IMO we should keep such stuff in the media repo, but remove everything that we haven't implemented yet from the main repo (apart maybe from stuff that we plan on implementing, such as the keeper hand).

Sounds relevant, indeed. I'd like to get @Danimal696 point of view to know the final plan, though. Because if we never add this creature, I wouldn't keep it at all, to be honest.

Bertram25:
> But it does work if I add it to the CMAKE_CXX_FLAG myself.

I am on it, and indeed I fell on the same conclusion. Weird, the previous worked when I made my tests...

akien-mga:
Tested and it works fine, in my opinion this can be merged. As for the duplicate Knight\* texture, we can change that later when @Danimal696 has reexported the models.

Bertram25:
Ok, I merge it, so we've got a Crypt. ;)

PR #321, opened 2014-11-14
Some quick-fixes

Bertram25:
On the knight coffin model and on using the Crypt in the editor.


hwoarangmy:
Works well. I had not seen the king on the WC for top claimed walls ^^

PR #324, opened 2014-11-16
Some more quick-fixes

Bertram25:
Line-endings + removed the golem worker from the keeper spawn pool :)


akien-mga:
Looking good. I just wondered, what is this resources.zip.cfg stuff anyway? Is it something that we use, or should we drop it?

Bertram25:
Good question. For now, we're not zipping resources but will we do that in the future? @oyvindln Do you remember what was the original plan?

hwoarangmy:
resource.zip.cfg is a problem not concerned by this PR. I merge

oyvindln:
@Bertram25 I think it's just a leftover from ogre example code, it should be safe to remove.

Bertram25:
@oyvindln Ok, I'll make sure to remove it in one of my next commits. Thanks for the info. :)

PR #325, opened 2014-11-18
Add a new AI that build every useful rooms + small fixes

hwoarangmy:
Fixes:
- Added a building type to buildiong class (to prepare for tile.getCoveringBuilding instead of room/trap)
- Fixed a bug in tile claiming where it is not displayed claimed if the kobold has a certain dance rate
- Replaced Player.getHasAI per Player.getIsHuman since it is what we want to know (and because inactive players have no ai assigned and should be handled like AI players)
- Added support to allow AI to pickup/drop creatures
- Fixed the map info reading in levels to display configurable players

Concerning AI:
- To place rooms, it uses a point system (about distance from dungeon heart and active spots). That can allow to easily change how the computer places rooms and maybe even change between different AI levels.
- If the AI has no treasury, it builds 1.
- If the an allied unit gets attacked on a nearby allied claimed tile, the AI drops creatures to help
- If its creatures get wounded, the AI picks them up and drops them in its dungeon heart so they can recover

As always, I've changed the skirmish 2v2 small test map to allow testing. Now, players 3 and 4 start without treasury.
Player 4 has 3 available claimed tiles and should build on it. Player 3 have available claimed tiles in your dungeon only (not reachable). Thus, he should wait for his kobolds to claim before building the treasury

While testing, don't hesitate to play with the editor to see how the AI adapts when it cannot build somewhere or when certain rooms are already built
Moreover, if you build something on allied claimed tiles where a room was going to be built, the AI should start digging somewhere else.


akien-mga:
Oh yeah! :-)

hwoarangmy:
BTW, for those who loose their dungeon to 2 BigKnights, the AI is now able to get PitDemons (it's so cool to see that ^^ ). Don't mess too much with it as you could suffer the same fate...

Bertram25:
OMG, need to try that asap!

akien-mga:
This is great so far, the AI does its job pretty well!

I've experienced one segfault when expanding an existing dormitory (which had been partly destroyed by BigKnights :-O), but it might not be related to this PR (maybe we never tried expanding a partly-destroyed room?).

```
During this turn there were 2 calls to GameMap::path().miscUpkeepTime=4930
Server sends newturn=1239
Client received turnStarted=1239
SERVER - Adding room Dormitory47, nbTiles=5, seatId=1
signal 11 (Segmentation fault), address is 0 from 0x7f635d68fecc

CLIENT - Adding room Dormitory47, nbTiles=5, seatId=1
[bt]: (1) /lib64/libstdc++.so.6 : std::string::assign(std::string const&)+0xc [0x7f635d68fecc]
[bt]: (2) /lib64/libstdc++.so.6 : std::string::assign(std::string const&)+0xc [0x7f635d68fecc]
[bt]: (3) ./opendungeons : RenderedMovableEntity::RenderedMovableEntity(GameMap*, std::string const&, std::string const&, float)+0x50 [0x4b6140]
[bt]: (4) ./opendungeons : Building::loadBuildingObject(GameMap*, std::string const&, Tile*, double, double, double)+0x124 [0x486134]
[bt]: (5) ./opendungeons : RoomDormitory::createMeshLocal()+0x220 [0x5600c0]
[bt]: (6) ./opendungeons : ODServer::processClientNotifications(ODSocketClient*)+0x6dbf [0x529d5f]
[bt]: (7) ./opendungeons : ODServer::notifyClientMessage(ODSocketClient*)+0x19 [0x536e99]
[bt]: (8) ./opendungeons : ODSocketServer::doTask(int)+0x10c [0x53ae1c]
[bt]: (9) ./opendungeons : ODServer::serverThread()+0x56 [0x538cd6]
[bt]: (10) /lib64/libsfml-system.so.2 : +0x7c28 [0x7f635dd1ac28]
[bt]: (11) /lib64/libpthread.so.0 : +0x75bd [0x7f635d8e25bd]
[bt]: (12) /lib64/libc.so.6 : clone+0x6d [0x7f635cdf578d]
```

hwoarangmy:
> I've experienced one segfault when expanding an existing dormitory

Well, I would bet it is not related to this PR. But I will have a look.

Bertram25:
I tested it and I love it already :D 

I've dug the main logic, but I still haven't fully understand how exactly points are computed so far, while it seems to be a nice hat-trick :tophat: :)
I get how you obtain the first tile before computing points, but I'll need to dig more to get how the rest of the room position is computed.

That said, the AI is really amazing now and it seems very good in term of performance as well...

I'll dig a bit more and merge if no body else finds something fishy. (We can always improve later on, anyway)

hwoarangmy:
I've tried to explain a bit what is done in the file source. If you feel it is not enough, please tell me what's the problem and I will try to explain more. When you will have understood what it is all about, that will mean that the comments are enough ^^

hwoarangmy:
BTW, when this one will be integrated, I might have a surprise that I've not talked much about and that might help us a lot to debug. But that's a secret ;-)

Bertram25:
@akien-mga I can't see why the segfault you've got would be linked with the current PR. Worth a new bug though. Wanna open it?

> I've tried to explain a bit what is done in the file source. If you feel it is not enough, please tell me what's the problem and I will try to explain more. When you will have understood what it is all about, that will mean that the comments are enough ^^

@hwoarangmy No, it's alright, I believe I figured out. :D I had to take the time to do a full test case by myself with a paper notes. It's cool and rather straight-forward once you get it. :) Computing the points given if you start digging a square of the given size at the given bottom left location. And if you're going too far away, the handicap value is forcing you to choose. It's well done! :)
Somehow, I also needed time to get you checked only the left or right wall if I understood correctly.

Di you get inspired by an existing algorithm or something like that, maybe?

> BTW, when this one will be integrated, I might have a surprise that I've not talked much about and that might help us a lot to debug. But that's a secret ;-)

Let's see, I'm merging. ;)

hwoarangmy:
> Computing the points given if you start digging a square of the given size at the given bottom left location. And if you're going too far away, the handicap value is forcing you to choose. It's well done! :)

Exactly ^^ Moreover, top and right squares are computed from bottom left to top right. bottom and left squares are computed from top right to bottom left. That's a remain of my original algorithm that was supposed to find the best available size (to allow building 3x3 to 5x5 depending on availability and distance handicap). But it was too complex and sometimes, I was missing the best place. So, I decided to force the width. If we find no 5x5 width room available, we will retry with 3x3.

> Somehow, I also needed time to get you checked only the left or right wall if I understood correctly.

Hmmm, that's not what I expected. If I made no mistake, we consider the four walls. BTW, you may notice that the tiles we will dig to reach the room are not taken into account. That's why, if you build a trap where an allied AI is going to build, it will choose a new place. If you sell the trap, The AI might not try to build in the place you have freed because 1 wall will have only 4 tiles and thus, will not get as many points as a new one.

> Di you get inspired by an existing algorithm or something like that, maybe?

No, it is a self made algorithm (but I had to play with paper before getting how to do it) ^^

PR #329, opened 2014-11-19
And even more quick-fixes :)

Bertram25:


akien-mga:
> Added spawn conditions to BigKnight equivalent to the PitDemon. … 

That sounds personal... :-P

Bertram25:
> That sounds personal... :-P

Eh eh. Nope, look:

> ok. Like you, if I play a game and the dwarfs kill my dungeon, I will add some strong spawn conditions

Now this is personal. ;P 
More seriously speaking, as BigKnight was a tier3 creature, I think adding spawn conditions equivalent to the pitdemon is the way to go.
Btw, we will have to add some spawn conditions to at least tier2 hero creatures, such as Dwarves, right? ;)

> Have you seen that all creatures did not had the same heal / turn ?

Yep, but I believe that certain of those healing values were too high, or even perfectly overpowered. (The troll had a heal rate of 1 HP per turn!). That's why I think we should start by aligning the heal / turn values and tweak slightly were we think it's relevant in a second time.

hwoarangmy:
> Btw, we will have to add some spawn conditions to at least tier2 hero creatures, such as Dwarves, right?

I bet the new AI killed your dungeon with 2 dwarves...

> That's why I think we should start by aligning the heal / turn values and tweak slightly were we think it's relevant in a second time

Seems good ^^

I will try that tonight. It should not break anything but I prefer to launch it before merging ^^

Bertram25:
> I bet the new AI killed your dungeon with 2 dwarves...

Pff, I can't wait to see your dungeons destroyed by <insert-some-random-creature-without-spawn-conditions> :P

akien-mga:
Looks good in game, let's merge.

Bertram25:
> Do you want to add the new Cavehornet material from @Danimal696 in this PR?

Danimal said he had also made other cleanups. So let's do all that in another PR. It will be easier for me to follow. (And you took the decision already, anyway. ;])

Danimal696:
i did already but not on a PR but directly, nothing was compomised be sure of that. I just removed  ramdom useless models

Issue #54, opened 2014-08-09
Game overall too dark with Nvidia on Linux (optimus/bumblebee)


akien-mga:
I was not mistaken yesterday when I said the game was a bit too dark, even for a "damp, dark and dangerous" dungeon :-D

Here are two comparative screenshots, one taken with my Intel GPU (where we know GLSL shaders don't work as per issue #41), and one taken with my Nvidia GPU. Note that I'm accessing the Nvidia GPU through the bumblebee daemon on Linux, which is kind of a hack, so the issue might be related (though it works fine in most games).

Intel (open source) driver
![odscreenshot_1](https://cloud.githubusercontent.com/assets/4701338/3865775/733dec36-1fa9-11e4-8181-da48e831d0d3.png)

Nvidia (proprietary) driver
![opendungeons-bug-nvidia-too-dark](https://cloud.githubusercontent.com/assets/4701338/3865777/b5f447aa-1fa9-11e4-81c6-8710d79ea8d5.png)

As a side note, I noticed that the screenshot feature in OD works fine with my Intel GPU, but does not work with the Nvidia GPU. My guess would be that the culprit is bumblebee, could someone with Linux and a non-optimus Nvidia GPU confirm that?

Bertram25:
@akien-mageia Ah indeed, it's dark ;)
I wondered: Did change the gamma in the nvidia settings, for instance? We don't have any configurable video options, so this might be a cause. ;)

Also, I'm sure the actual goal will later be to create a source light per rooms, for instance. As it seems some code was prepared for this. This would help a with you issue, IMHO.

hwoarangmy:
yes, I've seen some code for it too. But it has been deactivated so there might be a reason (too laggy?)... We could ask oln. Maybe he remembers why...

akien-mga:
I'll try to change the gamma in the nvidia settings to see if it's better.

Bertram25:
> yes, I've seen some code for it too. But it has been deactivated so there might be a reason (too laggy?)... We could ask oln. Maybe he remembers why...

For info, on Ogre 1.7, when adding too much light sources, the RT shader was crashing from what I read, and I guess that's the main reason lights code was removed.
Note that before, the code was adding a light source per room tile, which was way too much.
Also, light handling has improved (a bit) with Ogre 1.9.
@oyvindln may indeed have more info.

[...]

PR #330, opened 2014-11-20
Added a replay menu

hwoarangmy:
Here comes the surprise ^^


Bertram25:
Ok, I've looked at the code. It looks straight-forward and doesn't seem to contain any flaws so far. :)
And indeed, it's a real cool feature to permit debugging. So I, for one, support it.

About the file "spamming", I'd still request we have an option to disable them per default before next release (which is soon hopefully :]), since, sad to say but, the casual users don't care about helping the project with replays and we shouldn't add ~10Mb files in their user data folder per game.

The next thing to add, IMHO, is changing the speed of streamed packet reading, so we can quickly reach a specific point to debug, and maybe display the replay game time while watching one. :)

I'll test all that asap.

hwoarangmy:
> About the file "spamming", I'd still request we have an option to disable them per default before next release (which is soon hopefully :]), since, sad to say but, the casual users don't care about helping the project with replays and we shouldn't add ~10Mb files in their user data folder per game.

Yes but because we don't have an option gui, it might not be that easy. That's why I added in the replay menu the possibility to delete replay. IMHO, it is better to let it as default while there are no players (which will still be the case, IMHO, until we get FOW - when we will really have a playable game)

> The next thing to add, IMHO, is changing the speed of streamed packet reading, so we can quickly reach a specific point to debug

As you can guess, it is not that easy ;-) Even commercial games like Starcraft 2 have big problems when you try to jump within a replay ^^

> maybe display the replay game time while watching one. :)

Actually, displaying time might be interesting even during the game. We will see later if we need a specific mode for watching replays.

Bertram25:
> Yes but because we don't have an option gui, it might not be that easy. That's why I added in the replay menu the possibility to delete replay. IMHO, it is better to let it as default while there are no players (which will still be the case, IMHO, until we get FOW - when we will really have a playable game)

Ah, I didn't see the delete button. Makes things better, indeed. As for players, believe me, a few people I know were already having lans without me even knowing. :D So, IMHO, even if it's a bit of a wild guess, I think we'll have players already with the 0.4.9 version. :)
EDIT: If the others are ok with the replay menu as is, then nevermind my request to disable it, then.

> As you can guess, it is not that easy ;-) Even commercial games like Starcraft 2 have big problems when you try to jump within a replay ^^

Sure, that's why I didn't speak about a jump slider or the possibility to go back, only a way to speed up turns forward. But you're right, this can wait. :)

> Actually, displaying time might be interesting even during the game.

Yep, why not add this within the F11 text for now?

hwoarangmy:
> Yep, why not add this within the F11 text for now?

Yes, good idea. I will have a look

akien-mga:
The replay feature works great!

I noticed the replays can't start from time to time, and OD stops (it does not segfault, it seems to encounter a critical error and to shut down properly). So far, I had it happening:
- After playing "Big Test Map" for a while, then leaving the game (but not OD) and trying to see the replay.
- After playing "Scream in the dark" for a while, I could see the replay once (in the same OD session), but when I tried to start it a second time, the issue happened.

For this second time I reproduced the issue, here is what I have in the logs:

```
CLIENT - Adding Creature Dragon1
CLIENT - Adding Creature Dragon2
CLIENT - Adding Creature Dragon3
CLIENT - Adding Creature Dragon4
Loaded 4 creatures in level
ERROR: Assert failed file /home/akien/Projects/opendungeons/OpenDungeons/source/network/ODClient.cpp line 158 info : Wrong mode 7
CLIENT - Added player: Player
CLIENT - Added player: Keeper AI 2
CLIENT - Added player: Keeper AI 3
CLIENT - Added player: Keeper AI 4
CLIENT - Added player: Keeper AI 5
Client received turnStarted=0
```

```
Client received turnStarted=1
CLIENT - Adding Creature Kobold1
CLIENT - Adding Creature Kobold2
CLIENT - Adding Creature Kobold3
CLIENT - Adding Creature Kobold4
CLIENT - Adding Creature Kobold5
Client received turnStarted=2
CLIENT - Adding Creature Spider6
CLIENT - Adding Creature Wyvern7
CLIENT - Adding rendered object RenderedMovableEntity_Treasury8_[91,5]_130,MeshName=GoldstackLv1
CLIENT - Adding Creature Wyvern8
CLIENT - Adding rendered object RenderedMovableEntity_Treasury17_[4,91]_131,MeshName=GoldstackLv1
CLIENT - Adding Creature Wyvern9
CLIENT - Adding rendered object RenderedMovableEntity_Treasury24_[86,92]_132,MeshName=GoldstackLv1
CLIENT - Adding Creature Wyvern10
CLIENT - Adding rendered object RenderedMovableEntity_Treasury39_[38,54]_133,MeshName=GoldstackLv1
Client received turnStarted=3
Client received turnStarted=4
Client received turnStarted=5
Client received turnStarted=6
An internal Ogre3D error ocurred: OGRE EXCEPTION(5:ItemIdentityException): Object named 'Creature_Dragon1' does not exist. in SceneManager::getMovableObject at /home/iurt/rpmbuild/BUILD/ogre_src_v1-9-0/OgreMain/src/OgreSceneManager.cpp (line 6949)
Internal Ogre3D exception: OGRE EXCEPTION(5:ItemIdentityException): Object named 'Creature_Dragon1' does not exist. in SceneManager::getMovableObject at /home/iurt/rpmbuild/BUILD/ogre_src_v1-9-0/OgreMain/src/OgreSceneManager.cpp (line 6949)
*** ERROR: Internal Ogre3D exception: OGRE EXCEPTION(5:ItemIdentityException): Object named 'Creature_Dragon1' does not exist. in SceneManager::getMovableObject at /home/iurt/rpmbuild/BUILD/ogre_src_v1-9-0/OgreMain/src/OgreSceneManager.cpp (line 6949)
Quitting main application...
```

It looks like it starts processing turns even if all creatures have not been loaded yet.

akien-mga:
It seems I can always reproduce by loading two replays in the same OD session.
Here's a log from such a test case: http://paste.fedoraproject.org/152442/90821141/

akien-mga:
Apart from this, I had this feature request: maybe we could add a checked box in the in-game exit popup (when you press escape in a game), that would let you choose whether you want to save the replay or not. This way players could decide whether the game was interesting enough to be saved and replayed later.

I guess the checkbox would be opt-in, so that you only save the replays that you chose to save.

hwoarangmy:
The assert you got is about the mode not being the expected one. In fact, you are adding players while in replay menu but you should be in the configure seats when doing that. The problem is synchronization as I was talking about in the description. I will add a fix for that to make sure that in replay mode, we don't process more than one message at a time. I will see if it breaks replay or not.
But for this bug, we only need a workaround because as there is no server, there can't be synchronization. And, if we can avoid, I would prefer to not add workaround where they are not needed.

hwoarangmy:
@akien-mga Could you please try with this latest commits to launch your replay ?

Moreover, the dormitory has had several fixes (pretty buggy because it was not updated during successive improvements and had several problems due to different handling since first attempts at fixing multiplayer). BTW, that might also fix the bugs seen within dormitory.

akien-mga:
@hwoarangmy There's a case issue in BaseAI.cpp and KeeperAI.cpp:

```
/home/akien/Projects/opendungeons/OpenDungeons/source/ai/BaseAI.cpp:24:29: fatal error: gameMap/GameMap.h: No such file or directory
 #include "gameMap/GameMap.h"
                             ^
/home/akien/Projects/opendungeons/OpenDungeons/source/ai/KeeperAI.cpp:21:29: fatal error: gameMap/GameMap.h: No such file or directory
 #include "gameMap/GameMap.h"
                             ^
```

The folder is called "gamemap".

hwoarangmy:
@akien-mga Thanks, it is fixed now. Seems like my find and replace was a little bit too strong ^^

akien-mga:
I still get the same issue when loading a second replay after having watched a first one (partly, not until the end).

I confirm that the "wrong mode" issue has been fixed by 713aa98, but this issue remains and seems to be fatal:

```
An internal Ogre3D error ocurred: OGRE EXCEPTION(5:ItemIdentityException): Object named 'Creature_Dragon1' does not exist. in SceneManager::getMovableObject at /home/iurt/rpmbuild/BUILD/ogre_src_v1-9-0/OgreMain/src/OgreSceneManager.cpp (line 6949)
Internal Ogre3D exception: OGRE EXCEPTION(5:ItemIdentityException): Object named 'Creature_Dragon1' does not exist. in SceneManager::getMovableObject at /home/iurt/rpmbuild/BUILD/ogre_src_v1-9-0/OgreMain/src/OgreSceneManager.cpp (line 6949)
*** ERROR: Internal Ogre3D exception: OGRE EXCEPTION(5:ItemIdentityException): Object named 'Creature_Dragon1' does not exist. in SceneManager::getMovableObject at /home/iurt/rpmbuild/BUILD/ogre_src_v1-9-0/OgreMain/src/OgreSceneManager.cpp (line 6949)
```

akien-mga:
For the record, I've also checked what happens when you reach the end of a replay: the turns stop incrementing and the creatures stop moving, only the animations (portal) keep playing. That's good enough for now IMO :-)

Bertram25:
> For the record, I've also checked what happens when you reach the end of a replay: the turns stop incrementing and the creatures stop moving, only the animations (portal) keep playing. That's good enough for now IMO :-)

Yeah, this is good enough, indeed. @hwoarangmy is in Chuck Norris mode!!! :godmode: Take cover!
EDIT: Testing atm...

hwoarangmy:
Indeed, I am. And prepare because the floodfill patch is also coming :godmode:
BTW, sorry to mix commits like that but I don't like branches ^^

Bertram25:
:heavy_multiplication_x: . :heavy_multiplication_x:  :anguished:

hwoarangmy:
Here it is ^^

Bertram25:
The latest one deserve some good testing to make sure there are no regression, even if the code looks quite logical when you look at it. @hwoarangmy is in bug killing frenzy: You shouldn't come on his way or you will... Argh!

Bertram25:
> BTW, sorry to mix commits like that but I don't like branches ^^

As long as you cleanly separate commits, it's ok to me. In the end, they all go to a better (and common) place ;)

akien-mga:
I'll test the last commit and let @Bertram25 do some extensive testing of the beast before he merges. As for the part I already tested, it's good to be merged IMO, the bug when loading a second replay in a given OD instance can be looked at in another PR if @hwoarangmy doesn't send us another Chuck Norris bugfix :-)

hwoarangmy:
> if @hwoarangmy doesn't send us another Chuck Norris bugfix :-)

Sorry, I will fix at least the bug you found about loading replays. From what I've seen, it seems like a stupid error from me. I should post soon.

hwoarangmy:
@akien-mga I let you test ^^

akien-mga:
Commit d0cca66 does fix the bug in Stonekeep siege with digging gold surrounded with water.

There seems to be a small glitch in the digging animation, but we can have a look at it later on: my kobold was a bit close to the gold tile and not using is pickaxe when digging the first gold tile (I could reproduce it when digging a first dirt tile on the other side of the moat).

I did not see any regression, kobolds don't walk on lava or on walls, and cave hornets still fly over lava.

akien-mga:
And 16d24f3 does the trick! :-)

It's good to be merged as far as I'm concerned, but I'll let at least @Bertram25 have a look at it in game.

Bertram25:
I've found two things, the first is minor and shouldn't prevent a merge:
- When launching a replay, the host is configuring window is shown the time the replay is being loaded to me. Is it desired? (Not hurting that much, I was just wondering...)
- The second is this:
  odscreenshot_2014-11-21_005747_1
  On of the AI, the one on the right in stonekeep has got shades of colors when using the dormitory sometimes. ?? I'll try to invesigate more tomorrow. Does someone has the same bug?

In any case, if it's not related to this PR, I think you can merge. :)

akien-mga:
> On of the AI, the one on the right in stonekeep has got shades of colors when using the dormitory sometimes. [...] Does someone has the same bug?

I did not see that, but I'll let stonekeep run for a while to see if I can reproduce.

akien-mga:
Scratch that, I can confirm the issue with the same seat in Stonekeep siege.
od-bug

akien-mga:
I'd like to see whether it changes when creatures install their beds, but the AI doesn't seem that greedy yet, so it doesn't clear the gold path to the portal :-p

hwoarangmy:
> When launching a replay, the host is configuring window is shown the time the replay is being loaded to me. Is it desired? (Not hurting that much, I was just wondering...)

I wouldn't say it is desired but it is the easiest way so I just let it as it was (because its means just doing nothing). From my point of view, this replay feature is more for debugging than a production feature.
If you want, you can also pause the game during playtime or during the replay and you will see weird things ^^
IMHO, if we want to have a real replay player, it will need to be improved. But for me, as long as we see no critical bug (crash or something really stopping), that shouldn't prevent the merge.

hwoarangmy:
> Scratch that, I can confirm the issue with the same seat in Stonekeep siege.

Was it during game or during replay ?

akien-mga:
> > Scratch that, I can confirm the issue with the same seat in Stonekeep siege.
> 
> Was it during game or during replay ?

During game. I've also confirmed that the colours do not change when creatures drop beds in this dormitory.

I'll try to reproduce with @Bertram25's branch to check whether it's linked to this commit.

hwoarangmy:
ok. Then, I don't see how it would be linked to this PR. And I've just tried and I see nothing similar. Might be linux specific. Could you confirm if it is systematic and if it happens without this PR ?
Thanks

akien-mga:
I confirm that it happens with @Bertram25's branch, so it is not linked to this PR.
BTW with this other branch the bug looks exactly the same as it does in my above screenshot, so maybe the visual manifestation of the bug is computer-specific :-)

akien-mga:
Merging then.

hwoarangmy:
> BTW with this other branch the bug looks exactly the same as it does in my above screenshot

so, is it systematic ?

And thanks for the tests

akien-mga:
> so, is it systematic ?

I can't tell for sure, I reproduced it twice in a row, but now I've tried to run OD with my nvidia GPU (instead of my intel chipset as I do usually), and I don't see this separation (but the lighting is awful, already signaled in #54).

od-bug1

I'll open another issue about this dormitory bug.

hwoarangmy:
> but the lighting is awful

lol. In my computer, it looks much more like that ^^

Bertram25:
> I can't tell for sure, I reproduced it twice in a row, but now I've tried to run OD with my nvidia GPU (instead of my intel chipset as I do usually), and I don't see this separation (but the lighting is awful, already signaled in #54).

For the record, there is a possibility that the lighting seen with nvidia graphics is actually the correct one. Explaining why effects such as fog or skyboxes don't work/appear black. Again, the intel drivers have improved, but they are still not perfect, and Ogre 1.x was clearly not developed using them.
So, if, for instance, one of the next versions of the intel drivers make the game look more like on nvidia drivers, I wouldn't be surprised. ;) Hopefully, by the time we switch to Ogre 2.x, those kind of problems will have improved. :)

> I wouldn't say it is desired but it is the easiest way so I just let it as it was (because its means just doing nothing). From my point of view, this replay feature is more for debugging than a production feature.
> If you want, you can also pause the game during playtime or during the replay and you will see weird things ^^
> IMHO, if we want to have a real replay player, it will need to be improved. But for me, as long as we see no critical bug (crash or something really stopping), that shouldn't prevent the merge.

Sure :) My only point would then not to hide the "Loading..." text in that window in that case, so that we know it is loading and we can wait. Later, when the option menu has been added, we'll add an option to permit enabling/disabling such a feature, so we don't have too much questions by confused players. :)

Issue #322, opened 2014-11-14
Implement the Crypt logic


Bertram25:
From here: http://forum.freegamedev.net/viewtopic.php?p=60796#p60796
and here: http://forum.freegamedev.net/viewtopic.php?p=60821#p60821

First proposal from @Danimal696 :
- Creatures corpses are carried here by kobolds
- only one creature per coffin(active spot) allowed
- each coffin has a random rotting counter (say 30 to 50 seconds) that activates when a creature is deposited in it.
- when the counter reachs zero the coffin is "empty" ready for another corpse and some corpse points are added to the room.
- when enough corpse points are reached/stored by the room an undead creature appears with a 50/50 chance to be a ghoul or a ghost. (at least one of them for the issue. The other one can be done later.)
- the corpse points reset after an undead creature has appeared, so you cant get more than one undead at the same time if you rot an specially meaty (hp) creature like the troll, dragon or modified creature. (There should be no limit according to @hwoarangmy)

Unclear points:
- Should corpse points added per rotten creature be equal to his max Hp? the creatures rewarded by the Cript will be specialists and very strong, so you will have to rot, say, 40 spiders to get one or instead just 10 knights. I think its better than just using total quantity of corpses rotted.
  @hwoarangmy is ok with this.
- Note: I still want to use the prison as a source of cheap undeads skeletons as opposed to the stronger undeads given by the cript.
- A small chance to get a really powerful undead? 47.5% - ghoul / 47.5% ghost / 5% Sarcophag <- or just make this creature available througth sacrifies combinations at Temple?
  @bertram25 @hwoarangmy: No chance here if possible.
- @hwoarangmy : IMHO, it would be better to make the rats attack the prisoners. It's an evil game after all. And the number of active spots would change the number of rats (and, thus, the time it takes to turn them into skeletons) ^^
- @hwoarangmy : Concerning the crypt room, as you said, it could be a good thing to be able to deposit 1 corpse per central active spot. Concerning the wall active spots, they could be used to increase the amount of points you get per corpse (IMHO, it is good to always have a bonus when you manage well active spots).

@Danimal696 proposed spiders atm:
Spider model link: http://forum.freegamedev.net/viewtopic.php?p=60887#p60887

hwoarangmy:
> IMHO, it would be better to make the rats attack the prisoners

For this, we were talking about prison, not crypt ^^ As far as I've understood, spiders will be there just for fun

Danimal696:
[...]

PR #336, opened 2014-11-22
Added small spiders to the crypt

hwoarangmy:
part of #322

They are as simple as possible:
- They live for 10 turns
- Each central activespot from the crypt has 10% chances of spawning 1 per turn (which makes an average of 1 spider per central active spot for every turn)
- When they are outside the crypt (if the room is sold or destroyed), they die

With this logic, we don't have to handle anything like room absorbtion or active spot change


akien-mga:
I'll test this asap, since @hwoarangmy was in Chuck Norris mode lately he's probably urging to push tons of new commits :-)

akien-mga:
Works great, let's merge :-)

PR #340, opened 2014-11-24
Several stuff toward the 0.4.9 release

Bertram25:
Not many things left :)


akien-mga:
I'm testing it live.

hwoarangmy:
I will test that tonight. If everything goes well, I should also be able to do a PR implementing the crypt logic ^^

Bertram25:
> I will test that tonight. If everything goes well, I should also be able to do a PR implementing the crypt logic ^^

Cool :D Do you think we can still put it for 0.4.9?

hwoarangmy:
> Do you think we can still put it for 0.4.9

Yep, I think we can. It needs a little bit more testing but it should be good. And as you will see, it is done in a pretty nice OO way and with this, carrying objects to enable traps or bringing gold to the treasury is 95% done ^^

akien-mga:
This is not related to this PR but would be part of the "nice to have" fixes for 0.4.9: there are still a handful of sounds that might not be normalised and produce some glitches/cracking sounds.

@Bertram25 Do you have an efficient way to spot those?

Bertram25:
> @Bertram25 Do you have an efficient way to spot those?

Not AFAIK, I normalized everyone of them already. But maybe the bug happens also before reaching a sound value of 1.0f? Feel free to open them in Audacity, though. We never know...

Bertram25:
> Fix confirmed, now the dormitory is a bit dark (or more precisely, claimed tiles and walls are way too bright comparatively to the rest), but lighting improvements are for another milestone ;-)

Yep, the walls are bright especially as there are light sources around. Hopefully, we'll improve this by applying normals to each and every models as a first step. The second will be to adapt the ambient lighting to the new models once we made sure every normals are applied correctly.
The third step will be more fancy: Add distance fog and why not a skybox. ;)

akien-mga:
Looks good overall. I had troubles getting strong creatures on Scream in the dark, but I don't think it's related to this PR. Either I don't understand the spawn conditions, or I found a small bug in the logic that I couldn't analyse yet (my portal just stopped working after a while, regardless of all the requirements I had reached).

hwoarangmy:
It stops working when max creature number has been reached. Currently 15 per seat

akien-mga:
> It stops working when max creature number has been reached. Currently 15 per seat

Ah that makes sense, I had been pretty lazy so I had mostly tier 1 and tier 2 creatures :-)

hwoarangmy:
It works also for me. I merge

Bertram25:
I'll fix the typos in the next PR.

PR #341, opened 2014-11-25
Implement crypt logic

hwoarangmy:
If you rot 10 hp in the crypt, you might spawn.... a gnome ^^
It is configurable in config files and can be changed when @Danimal696 will have finished the proper creature. I've modified the 2v2 skirmish to allow test more easily

### DIFF
--- a/config/global.cfg
+++ b/config/global.cfg
@@ -79,4 +79,6 @@
 [/ConfigFiles]
 [GameConfig]
     NetworkPort	31222
+    CreatureDeathCounter	30
+    MaxCreaturesPerSeat	15
 [/GameConfig]
--- a/config/rooms.cfg
+++ b/config/rooms.cfg
@@ -13,6 +13,9 @@
 # TrainHallAwaknessPerAttack        (double) How many awakness a creature looses for each blow
 # TrainHallCooldownHitMin           (uint)   How many turns a creature will stay idle after hitting a dummy minimum
 # TrainHallCooldownHitMax           (uint)   How many turns a creature will stay idle after hitting a dummy maximum
+# CryptRotNbTurns                   (uint)   How many turns a creature will rot before its max hp are added to the crypt
+# CryptPointsForSpawn               (uint)   How many points are needed for the crypt creature to spawn
+# CryptSpawnClass                   (string) Class of the creature that will spawn when the given points are gathered
 
 [Rooms]
     HatcheryHungerPerChicken	10
@@ -24,4 +27,7 @@
     TrainHallAwaknessPerAttack	5.0
     TrainHallCooldownHitMin	3
     TrainHallCooldownHitMax	6
+    CryptRotNbTurns	10
+    CryptPointsForSpawn	10
+    CryptSpawnClass	Gnome
 [/Rooms]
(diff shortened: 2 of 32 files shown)


akien-mga:
I guess the commit log for 8b5a317 does not completely reflect everything that you did, no?

hwoarangmy:
> I guess the commit log for 8b5a317 does not completely reflect everything that you did, no?

Yes, it is my mistake. I thought I had made a new commit but I was editing the oldest one. I've seen that only after pushing to my repo. But the fixes are there so...

A best suited tile would be:
Implemented crypt logic ^^
BTW, I've added the logic to allow taking any entity to  some building. That means that when we will want to implement forge logic, it will be easy. The forge logic will be a copy of the training hall. When an object will be created, everything is ready so it can be carried to the needing trap ^^

Bertram25:
Aww, you're evil. There is no uglier, kinder, saner and happier creature than a gnome. Erk.

Danimal696:
:D great news!! but a gnome?! such aberration, you are insane!!

Bertram25:
The code looks fine AFAICT. I'll test it now. My newest PR will have to be re-rebased once this one is merged. ;)

Bertram25:
I managed to trigger the treasury seeking behaviour. It's working fine, while it seems not very high in the action priority list of the worker. Should it be raised (at least for gold) or should we rather do that for next release or never?

What do you think, guys?

Bertram25:
@Danimal696 Seems you have no more excuse to add the necrophag. ;)

Bertram25:
@akien-mga If you see no obvious bugs, you can merge. :)

hwoarangmy:
> while it seems not very high in the action priority list of the worker

Yep, it is after claiming walls (when dropping on a dead creature). We could increase the probability to check for carryable entities. In any case, I've added a boolean to avoid looking several times for that. And from what I've seen, it could be a good thing to do the same for tile claiming. But anyway, the idle action should be re-written IMHO. We will see later ^^

Bertram25:
> We will see later ^^

Sure. Just wanted to know what you thought about it. :)

hwoarangmy:
> Just wanted to know what you thought about it. :)

It's just a feeling but I think it shouldn't be too high. We will see during our next lan ;-)

hwoarangmy:
> :D great news!! but a gnome?! such aberration, you are insane!!

I wanted to add a BigKnight but if we use it against @Bertram25, I feared he would change their stats so they would not be a threat anymore...

Bertram25:
> I wanted to add a BigKnight but if we use it against @Bertram25, I feared he would change their stats so they would not be a threat anymore...

As long as I've got pitdemons ;P

akien-mga:
I tested the crypt logic, it works fine.

I think that the corpse carrying action should have a higher priority than claiming tiles/walls when dropping a worker on a corpse. Currently it can be pretty hard to time things correctly since when you drop a kobold on a corpse, it will start by claiming the tile or the wall next to it (if it needs to be claimed of course), so you often have to drop it on the corpse twice or three times so that it grabs it. And since the corpse disappear quickly, this can be tricky.

hwoarangmy:
> And since the corpse disappear quickly

This can be changed in the config file. IIRC, currently, it is set to 30 turns ~20 seconds. But we can increase if needed.

> Currently it can be pretty hard to time things correctly since when you drop a kobold on a corpse, it will start by claiming the tile or the wall next to it

Yes, but I believe that right now, we want to drop them on the corpe because we want to test the crypt logic. I'm not sure that during a game, it will be such a need to carry the corpse so fast. Anyway, this can be changed easily so we will see ingame ^^

PR #342, opened 2014-11-25
Added support to make an entity translucent (For the boulder trap) + game entity class cleanup.

Bertram25:


Bertram25:
I'll rebase asap :)

Bertram25:
Allright, rebase done. :)

Bertram25:
To be rebased once the crypt PR is merged. Please don't merge this one before. Thanks :)

Bertram25:
Ok, I'll rebase and apply the requested fixes tonight if all goes well.

Bertram25:
Alright, all good. You can now review the code. :)

hwoarangmy:
I've stolen your code about opacity to test on the keeper hand and it works ^^

Bertram25:
> I've stolen your code about opacity to test on the keeper hand and it works ^^

Then, I wouldn't like to be in your place, all my code is made under the Evil 3.0 license ;)

hwoarangmy:
The patch seems good but there is a problem with the way you change traps opacity. IMHO, it would be easier to use activate(Tile*) because I guess most of the traps will trigger one after the other so there is probably not much interest in allowing changing opacity for every tile in a trap.

Bertram25:
> The patch seems good but there is a problem with the way you change traps opacity. IMHO, it would be easier to use activate(Tile*) because I guess most of the traps will trigger one after the other so there is probably not much interest in allowing changing opacity for every tile in a trap.

Then, here, there is clearly something in the concept I didn't get. I thought every traps were made of one and only one tile per instance? Can you tell me more?
EDIT: Plus, the activate/deactivate mechanism is here to tell the trap has triggered as a whole, not a part of it. If I'm not mistaken traps aren't working like rooms aka absorbing other trap's tiles.

hwoarangmy:
Well, when I first asked, that's what I had proposed but you all told me it was better to have the same behaviour as rooms. Thus, a trap can have several tiles. Each being triggered independently. But they don't get absorbed like rooms.

But it's not that complex. A trap (class Trap) is just composed of several tiles each having a trap (rendered object) on it

hwoarangmy:
Here is the discussion:
http://forum.freegamedev.net/viewtopic.php?f=35&t=5858

Now that I read it again, it is not that clear that you wanted the room like behaviour. But back then, I thought it was ^^
Anyway, that doesn't really matter. You just have to use the Tile for activation/deactivation and that will do. BTW, that's why Trap::shoot needs the tile as parameter ;-)

Bertram25:
> Anyway, that doesn't really matter. You just have to use the Tile for activation/deactivation and that will do. BTW, that's why Trap::shoot needs the tile as parameter ;-)

By re-reading it, I think I understand why you went that way, and I guess it's fine, as long as it doesn't get in the way for the needed features. I'll do the fix :)

hwoarangmy:
In any case, it might be usefull in the futur because if we want to, we can move the room absorbtion logic to building to allow traps to absorb traps and to make some that are more powerfull when bigger. Even if we said it wasn't needed back then, I think we might find a use for it now that it is the case. Moreover, the fact that trap is a building will help us much when the forge will be used for traps (bu changing a little bit the code like I did for treasuries)

Bertram25:
ok, should be done now :)

hwoarangmy:
I will try that tonight but from what I've seen, it should be ok. Anyway, we shouldn't put too much energy on that as when the forge will be ready, the activation parameter will not be used anymore (at least if we keep the DK behaviour)

BTW, I've just been thinking that it could be good to check if the opacity change happens only for the concerned player. From what I've seen, every players will receive the notification with this patch. But we can do that later.

Bertram25:
> BTW, I've just been thinking that it could be good to check if the opacity change happens only for the concerned player. From what I've seen, every players will receive the notification with this patch. But we can do that later.

You're right. Every player will receive the info. While this seems a faulty behaviour, the trap should not even be visible when not discovered by enemy players, so I guess the whole thing has to later be rethought, anyway.

The good point is that we have now some code capable to make an entity translucent. :)
Btw, as for the Keeper hand, the translucency can be done directly in the material file, but is it desired?

Bertram25:
> as when the forge will be ready, the activation parameter will not be used anymore (at least if we keep the DK behaviour)

Yes, I confirm traps/doors should be forged and activated by carriers. It will be much more fun that way and will prevent battle front trap spamming. And they will deactivate once used a few times (to decide for each...). In any case, they'll probably keep their activated state so AFAIK there is no code effort waste, IMO.

While I think about it, I do think the room and traps price per tile should be made a configurable value, even if the current value looks fine so far. What do you think?

akien-mga:
> While I think about it, I do think the room and traps price per tile should be made a configurable value, even if the current value looks fine so far. What do you think?

I agree.

hwoarangmy:
> Btw, as for the Keeper hand, the translucency can be done directly in the material file, but is it desired?

For having tested it, yes, it is desirable because the hand has more or less the size of the creatures and it hides them. As for every things, it can be changed if we think it is bad after all. Concerning the mesh itself, it will need to be changed because:
- Animations have wrong names: walk and idle
- There are only 2 animations (nothing for slapping)
- It is opaque (and IMHO, to should not - that's why I do that by code)

> While this seems a faulty behaviour

Not really, as you said it will have to be changed to make enemy players not see traps. So, IMHO, it is not the right time to do that. It was more like a reminder.

> While I think about it, I do think the room and traps price per tile should be made a configurable value, even if the current value looks fine so far. What do you think?

Yep. Actually, I don't know why we haven't already done that as it is pretty obvious

Bertram25:
> For having tested it, yes, it is desirable because the hand has more or less the size of the creatures and it hides them. As for every things, it can be changed if we think it is bad after all.

Ok, you can always your PR, and I can change that afterwards in the material file. In any case, @Danimal696 should come into this playground to fix the mesh issues, before I move on with this. And as now he knows to apply translucency, maybe he won't even need me. ;)

EDIT: I'll open an issue for 0.5.0 to make the room/trap price configurable. If it's done before, then all the better. :)

Danimal696:
i can do alphas but no translucencies, i would need a code sample to manually edit the file. 
So the hand has two animation badly named? what should they be called?

Bertram25:
> i can do alphas but no translucencies, i would need a code sample to manually edit the file. 

That's exactly the same thing, don't worry. As a proof, you successfully made the cavehornet's wings translucent using the alpha channel and blending options.

hwoarangmy:
> So the hand has two animation badly named? what should they be called?

There is one "Walk" that just moves a little bit the fingers. I would call it "Idle" and that would be the default looped animation.
Then, there is a "Idle" animation that shows the hand grasp something. I would call it "pickup".

So, there is no "Slap" nor "Drop" animation. I think that Slap would be the more urgent. For transparency, we will see which feedback we will get when released. If wanted, I can add something in the config manager to allow changing opacity from 0 to 1

Bertram25:
> If wanted, I can add something in the config manager to allow changing opacity from 0 to 1

Instead, I would change the material to make it support the translucency, and remove it form the code. I'll have a look at that once I've seen your PR. In meanwhile, you can try that if you want (based on the cavehornet's one):

``` script
material Keeperhand
{
    receive_shadows on
    technique
    {
        pass
        {
            ambient 0.800000011920929 0.800000011920929 0.800000011920929 1.0
            diffuse 0.6400000190734865 0.6400000190734865 0.6400000190734865 1.0
            specular 0.1551724523305893 0.1551724523305893 0.1551724523305893 1.0 37.5
            emissive 0.0 0.0 0.0 1.0
            scene_blend alpha_blend
            depth_write off
            cull_hardware none
            cull_software none
            texture_unit
            {
                texture Keeperhand_colC.png
                tex_address_mode wrap
                filtering trilinear
                colour_op alpha_blend
            }
        }
    }
}
```

You can also change all the '1.0' values to less (but the same one) if needed.

Danimal696:
ok, i will try

hwoarangmy:
ok. I will try tonight

hwoarangmy:
Good, I merge

PR #348, opened 2014-11-27
Added the keeper hand + slap

### LINKED ISSUE #65: Add support for the Keeper's Hand cursor.
There is a model ready for that:
https://github.com/OpenDungeons/OpenDungeons/raw/development/models/Keeperhand.mesh

Later, adding support to slap own creature's to make them go quicker may be considered. ;)

hwoarangmy:
- several improvments about GameEntity handling
- removed maplight from the gamemode/editormode
  fixes #65 
- The most wanted feature we were all missing and dreaming about: you can now slap creatures/chickens/crypt smallspiders ^^


source/modes/EditorMode.cpp
```diff
@@ -434,65 +459,21 @@ bool EditorMode::mousePressed(const OIS::MouseEvent &arg, OIS::MouseButtonID id)
...
         std::string resultName = itr->movable->getName();
 
-        if (resultName.find(Creature::CREATURE_PREFIX) != std::string::npos)
-        {
-            // Pick the creature up and put it in our hand
-            if(inputManager->mExpectCreatureClick)
-            {
-                // TODO : switch to FPP mode
-#if 0 // FPP mode do not exist yet but we keep track of what was done.
-                mModeManager->requestFppMode();
```
hwoarangmy, line 444: I've removed that since it will probably not be used for a while and this code is becoming more and more obsolete

Bertram25:
The code looks fine, has got interesting cleanups and is fun to see in action. ;)
As a first improvement, I'd rotate the hand 45° counter clock-wise to make it more naturally placed.
Secondly, I'd trigger the creatures hurt animation when slapped, and I'd even make them flee if low on HP like if they were attacked. ;)
The other improvements are on @Danimal696 's shoulder. The slap animation should be rather dynamic, and the pickup animation should be kept, IMHO, since we can reuse it.

It's starting to look cool :)

Bertram25:
IMHO, can be merged if nobody sees a bug. :)

hwoarangmy:
> As a first improvement, I'd rotate the hand 45° counter clock-wise to make it more naturally placed.

If you want, we can create an ussue for that

> Secondly, I'd trigger the creatures hurt animation when slapped, and I'd even make them flee if low on HP like if they were attacked.

Yep, hurt animation is a good idea (and it is a good place to use it). Concerning fleeing, we should decide how we do that. Currently, they would flee if they have no bed. But if they have one, they will go sleeping (walking, not fleeing). It may be interesting to let them flee when slapped for a few turns. But what to do if there are ennemies around?

Bertram25:
> Concerning fleeing, we should decide how we do that. Currently, they would flee if they have no bed. But if they have one, they will go sleeping (walking, not fleeing). It may be interesting to let them flee when slapped for a few turns. But what to do if there are ennemies around?

IMO, each time a creature is slapped and his HP are below something like 50% (later, why not decrease their mood for the same test?), we could trigger the same behaviour as if attacked by enemies. This way, they would flee as if there were enemies around. And if there actually are enemies around, well, that won't change anyrthing, right? ;)

As for all the rest, we can create an issue for 0.5.0 for that. :)

akien-mga:
> Secondly, I'd trigger the creatures hurt animation when slapped, and I'd even make them flee if low on HP like if they were attacked.

It would be nice if slapping a creature would force it to drop the first item in its priority list and run to do something else :-) i.e. if you kobold is currently claiming tiles but you want it to go dig further, you slap it and it will stop claiming tile and search for another task assigned to it. Same with other creatures, you don't want your BigKnight sleeping when it could be training, slap it! :-P

akien-mga:
Looks good, I merge :-)

akien-mga:
I guess some balancing is needed, I one-shot-slapped my BigKnight in the 2v2 test map, but IIRC he had a low HP per design :-)

hwoarangmy:
> I guess some balancing is needed, I one-shot-slapped my BigKnight in the 2v2 test map, but IIRC he had a low HP per design :-)

Yes. Slapping takes15% of the creature max HP (allowing to kill them if you want). Because the BigKnight starts with 4 hp and max hp = 60, you can one shot kill him ^^
And because 2 BigKnights had the bad idea to kill @Bertram25, now, they heal really slowly so you have to be patient to not be able to kill him

> It would be nice if slapping a creature would force it to drop the first item in its priority list and run to do something else

They should do the most prioritary tasks if you don't force them by dropping to do something. We should keep this design because AI will probably not use that to change their behaviour. So, if we rely too much on slapping, AI would need to be improved...

> we could trigger the same behaviour as if attacked by enemies

Yes but if there are no enemies around, it would stop fleeing. That's why I was talking about having a minimum turns of fleeing.

> And if there actually are enemies around, well, that won't change anyrthing, right? ;)

Yes, the problem is if there are no enemies around ;)

hwoarangmy:
For @Danimal696 I forgot to say that if you slap a creature on the editor, it is removed from the gamemap.

Danimal696:
> For @Danimal696 I forgot to say that if you slap a creature on the editor, it is removed from the gamemap.

Thats pretty interesting ;) . What state is the hand in? i can tell you did some modification on it but its not clear enougth from your post which ones. did you already a slap animation or just the code?

hwoarangmy:
I've used the "Idle" animation to slap (which looks like grabing) and the "Walké animation (that looks like idle) when doing nothing special.

Bertram25:
> And because 2 BigKnights had the bad idea to kill @Bertram25, now, they heal really slowly so you have to be patient to not be able to kill him

XD

> Yes but if there are no enemies around, it would stop fleeing. That's why I was talking about having a minimum turns of fleeing.

Ah, indeed.

> For @Danimal696 I forgot to say that if you slap a creature on the editor, it is removed from the gamemap.

Ah, useful. :)

Bertram25:
@hwoarangmy Btw, I forgot to test that, but it seems now you can't open status window of enemy creatures, right?

akien-mga:
> @hwoarangmy Btw, I forgot to test that, but it seems now you can't open status window of enemy creatures, right?

I can still do that here.

Bertram25:
@akien-mga ok, thanks for the test. :)
I've re-read the code, and there is indeed a canOpenStatusWindow() function but it only checks the entity type. Nevermind then,

Danimal696:
> I've used the "Idle" animation to slap (which looks like grabing) and the "Walké animation (that looks like idle) when doing nothing special.

thats pretty messed up

hwoarangmy:
> thats pretty messed up

Yep, that's what I said in the issue ^^

Bertram25:
Funny, isn't it? Was the hand planned to be a fingertip-walking monster?

PR #351, opened 2014-11-28
When the XP table is overriden in a level file, it should be written by the editor when the level is saved

hwoarangmy:
To test it, you can open a level where the XP table is overriden for a creature and save it. In the saved level, the xp table should be the same


hwoarangmy:
@Bertram25 It should be fixed IMO

hwoarangmy:
> we'll have to check why def1->mXPTable[MAX_LEVEL] is valid while undefined

Seems like vector do not really care about index being >= vector.size. BTW, this patch assumes that every creature has the same max level (which is better IMHO as it can be pretty hard to understand from a player's point of view that the max level is different for each class). But if, for some reason, we change that, we should keep it in mind.

PR #352, opened 2014-11-30
Added all GameEntities that can be in a tile to one list only to make code more generic

hwoarangmy:


Bertram25:
I'll test tonight along with the removal of the rendermanager. :)

hwoarangmy:
I will post tonight a fix with the requested changes and a little bit more cleaning. I'm almost done with replacing Tile::getCoveringRoom/Tile::getCoveringTrap by Tile::getCoveringBuilding

I've also seen that I've not applied code conventions on all the Tile members (because I was working on vectors only). But as it is pretty straight forward, I might also have a look there.

Bertram25:
Cool :)

hwoarangmy:
Is it more readable like that ?

Bertram25:
Much better :D I'll test now.

Bertram25:
Ouch, infinite loop bug detected or so it seems. I'm not sure if this patch is actually concerned, so I'll try without.
To reproduce create a dormitory (I used 3x4) then create an extension to the dormitory: A new dormitory with adjacent tiles to the first one. The game should freeze and the memory will quickly grow. So be prepared for a quick app kill as well. ;)

I'll try without the PR now.

Bertram25:
No freeze without the PR. The bug is hiding here, somewhere. +.+

source/entities/ChickenEntity.cpp
```diff
@@ -84,8 +84,17 @@ void ChickenEntity::doUpkeep()
     }
 
 
-    Room* currentHatchery = tile->getCoveringRoom();
-    if((currentHatchery != nullptr) && (currentHatchery->getType() == Room::RoomType::hatchery))
+    Room* currentHatchery = nullptr;
+    if(tile->getCoveringRoom() != nullptr)
+    {
+        Room* room = currentHatchery = tile->getCoveringRoom();
```
Bertram25: Typo here.

hwoarangmy: Nice catch

hwoarangmy:
It is indeed in the room absorbtion code. I'm currently having a look

source/entities/Tile.cpp
```diff
@@ -1120,12 +1097,7 @@ double Tile::scaleDigRate(double digRate)
 
 Tile* Tile::getNeighbor(unsigned int index)
 {
-    return neighbors[index];
-}
-
-std::vector<Tile*> Tile::getAllNeighbors()
-{
-    return neighbors;
+    return mNeighbors[index];
```
Bertram25, line 1104: As seen in the Experience commit bug when you requested mExp[TOTAL_EXP] instead of mExp[TOTAL_EXP - 1]
Calling the [] operator for the vector does create an undefined values for the given index if it is invalid.
So, note for later, add a check here. :)

source/rooms/Room.cpp
```diff
@@ -497,15 +497,14 @@ void Room::setupRoom(const std::string& name, Seat* seat, const std::vector<Tile
...
 {
     bool isRoomAbsorbed = false;
-    std::vector<Tile*> borderTiles = getGameMap()->tilesBorderedByRegion(getCoveredTiles());
-    for (std::vector<Tile*>::iterator it = borderTiles.begin(); it != borderTiles.end(); ++it)
+    for (Tile* tile : getGameMap()->tilesBorderedByRegion(getCoveredTiles()))
     {
-        Tile* tile = *it;
-        Room* borderingRoom = tile->getCoveringRoom();
-        if (borderingRoom != nullptr && borderingRoom->getType() == getType()
-            && borderingRoom != this && borderingRoom->getSeat() == getSeat())
```
hwoarangmy, line 506: seems like i forgot the "borderingRoom != this"
What a stupid mistake ^^

Bertram25, line 506: Don't be too hard on you, the good is much better than the bad. And luckily, I've hunted bugs like @akien-mga this time. ;)

hwoarangmy:
It should be good now

source/rooms/Room.cpp
hwoarangmy, line 506: lol. That's why this PR should not be in the 0.4.9 unless we test it pretty well ^^

Bertram25, line 506: IMHO, nothing unless extremely needed should be part of 0.4.9 from now on.
0.4.9 is already in stable and if I manage to sort out some balancing improvements this week, we should have a testable balance in a few days. :) Afterwards, it will be release time! :D

Bertram25:
I didn't have time to compile yesterday. Probably tonight.

Bertram25:
The latest version is ok. I'd say it's good to merge. So I'll merge and we'll see whether other bugs appear but I think this should be fine.

PR #359, opened 2014-12-03
Drop resources.zip.cfg

akien-mga:


akien-mga:
@Bertram25 @hwoarangmy 
How do you manage your own fork and the main OpenDungeons repo + all pull requests?

I've tried to configure remotes and branches to get everything in the same git clone (I build each branch in a separate build-<name> folder), and it works great to test your pull requests.

I've spent some time trying to configure my own "akien-dev" branch pointing to my fork, and I can pull without any problem, but I don't manage to push my changes to github. `git status` tells me that I have commits to push, but `git push` says that everything is already up to date.

So for this PR I had to fall back to my old system of having a git clone of my fork separate from the main clone where I test your PRs.

Bertram25:
@akien-mga 
Well, since one or two git versions, 'git push' permits you push to the same named remote branch only, and only if it is set as upstream.
This means two things:
1. When you try 'git push' alone, it will try to push local default branch to the same named remote branch.
2. To push to something else, you'll have to be in the branch you want to push from, and do: git push origin <branch-name> . Branch-name will have to be the same as the branch you're in.

Did I answer your question?

Bertram25:
Thanks, looks fine and is working. I'll merge.

akien-mga:
So IIUC, I can't `git push` to github.com/akien-mga if I cloned github.com/OpenDungeons and set the latter as upstream? That makes sense and might make it easier not to break things in the upstream repo, but I wondered if there was a way to make it "all in one" :-)

_Edit:_ Or actually, it could work since the upstream can be set on a per-branch basis. The remaining issue in my case might be that my local branch was named "akien-dev" with akien-mga/OpenDungeons/development as upstream branch. So the issue would be the fact that my local branch has a different name than the upstream branch? But obviously I can't name my local branch "development" since the name is already used by OpenDungeons/OpenDungeons/development.

Bertram25:
> So the issue would be the fact that my local branch has a different name than the upstream branch?

In brief, yes. In that case, you can either `git push origin akien-dev` which will create or update the corresponding remote akien-dev branch and manage your PR from that one, or `git checkout development` and cherry-pick your commits from akien-dev, or `git merge akien-dev` when in development.

In any case, you should never clone the main repo directly, but rather clone your own fork and do a `git pull git://gihub.com/OpenDungeons/Opendungeons.git development` for instance, if you need to refresh it before working on it.
You can also refresh your commits using `git pull --rebase git://gihub.com/OpenDungeons/Opendungeons.git development` for instance, when your PR indicates merge conflicts.

I might have over-answered. I hope you don't mind. ;)

akien-mga:
Thanks, I'll give it another try using your input. I think my mistake was that I tried `git push akien-mga development` when it should have been `git push origin akien-dev` (provided `origin` is defined properly I suppose).

I'll also have a look at migrating my PR testing environment into my own fork instead of the main repo :-)

Bertram25:
> I'll also have a look at migrating my PR testing environment into my own fork instead of the main repo :-)

Thanks. I say that because it helps avoiding bad things to happen when a push went in an unexpected way. We're all human after all. ;)

As for origin:
http://stackoverflow.com/questions/9529497/what-is-origin-in-git

You can of course change 'origin' to another repo URL but it should be rather rare in a normal workflow here.

PR #360, opened 2014-12-03
Fixed MSVC compilation + fix for 358

Bertram25:
I assume the MSVC compiler tries to optimize the char array creation, triggering an error when compiling.

### DIFF
--- a/source/ai/BaseAI.h
+++ b/source/ai/BaseAI.h
@@ -20,6 +20,7 @@
 
 #include <string>
 #include <vector>
+#include <cstdint>
 
 class GameMap;
 class Player;
--- a/source/modes/MenuModeConfigureSeats.h
+++ b/source/modes/MenuModeConfigureSeats.h
@@ -20,6 +20,8 @@
 
 #include "AbstractApplicationMode.h"
 
+#include <cstdint>
+
 class Seat;
 class Player;
 class ODPacket;
--- a/source/network/ODPacket.cpp
+++ b/source/network/ODPacket.cpp
@@ -249,25 +249,30 @@ void ODPacket::writePacket(int32_t timestamp, std::ofstream& os)
     os.write(buffer, bufferSize);
 }
 
+const int32_t BUFFER_SIZE = 1024;
+
 int32_t ODPacket::readPacket(std::ifstream& is)
 {
     int32_t timestamp;
-    int32_t bufferSize;
+    int32_t packetSize;
 
     is.read(reinterpret_cast<char*>(&timestamp), sizeof(int32_t));
     if(is.eof())
         return -1;
 
-    is.read(reinterpret_cast<char*>(&bufferSize), sizeof(int32_t));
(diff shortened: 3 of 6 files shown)


oyvindln:
I think it's because MSVC doesn't support [VLAs](https://en.wikipedia.org/wiki/Variable-length_array). I don't think using it together with an input-defined size is a good idea anyhow.

Bertram25:
Yeah, sounds true. Some sanity check may have to be added, maybe. @hwoarangmy will know more since he's the one creating the packet handling in the first place.

hwoarangmy:
yep, a size check might be a good idea at least. I'm gonna have a look if there is something we can do with a predefined size buffer. IIRC, sf::Packet almost gives access to its data array but I think it is a vector. I will tell more soon

hwoarangmy:
I've add a look at sf::Packet code and I think there is no problem with using a sized buffer. I'm gonna try right now and post the code if it works

hwoarangmy:
ok, I've tested and it works with this:

```

const int32_t BUFFER_SIZE = 1024;

int32_t ODPacket::readPacket(std::ifstream& is)
{
    int32_t timestamp;
    int32_t packetSize;

    is.read(reinterpret_cast<char*>(&timestamp), sizeof(int32_t));
    if(is.eof())
        return -1;

    is.read(reinterpret_cast<char*>(&packetSize), sizeof(int32_t));
    if(is.eof())
        return -1;

    mPacket.clear();
    char buffer[BUFFER_SIZE];
    while(packetSize > 0)
    {
        int32_t sizeToRead = std::min(packetSize, BUFFER_SIZE);
        is.read(buffer, sizeToRead);
        mPacket.append(buffer, sizeToRead);
        packetSize -= sizeToRead;
    }

    return timestamp;
}
```

Do you want to go for it ? I've tested with BUFFER_SIZE = 10 and it is ok.

Bertram25:
Looks good. Thanks for the fix.
I'll test asap and upgrade my PR with your change if it's ok. :)

Bertram25:
Btw:

> Remember to use delete[]. Or even better, use unique_ptr.

@oyvindln  Thanks for the reminder for the delete. It's such a common mistake. ;) I won't use a unique_ptr for such a simple case if it is possible though but I understand the proposal.

oyvindln:
Well @hwoarangmy has a better suggestion anyhow.

Bertram25:
Ok, I cleaned up the main repo history and added the commit fixing compilation.

hwoarangmy:
Nothing to say with the code, I'm currently testing

hwoarangmy:
That's good, i merge

akien-mga:
Currently learning some basics of C++, I'll soon™ be able to review the code with a better-trained eye ;-)

For the reference I'm learning with _C++ Primer_, I read good reviews of it. The 5th edition is updated for C++11 too, so for OD it should be fine.

Bertram25:
Cool :)

PR #362, opened 2014-12-04
Reworked room absorbtion code

### LINKED ISSUE #361: Some active spots are lost when a room is absorbed
When a room is absorbed, if a creature is working, the active spot it was on will not be used anymore. This is a regression from an unknow patch.
Anyway, the room absorbtion code needs to be refreshed since many things have changed with buildings

hwoarangmy:
fixes #361

### DIFF
--- a/source/ai/BaseAI.cpp
+++ b/source/ai/BaseAI.cpp
@@ -53,8 +53,8 @@ bool BaseAI::buildRoom(Room* room, const std::vector<Tile*>& tiles)
 {
     room->setupRoom(mGameMap.nextUniqueNameRoom(room->getMeshName()), mPlayer.getSeat(), tiles);
     mGameMap.addRoom(room);
-    room->checkForRoomAbsorbtion();
     room->createMesh();
+    room->checkForRoomAbsorbtion();
     room->updateActiveSpots();
     mGameMap.refreshBorderingTilesOf(tiles);
 
--- a/source/entities/Building.cpp
+++ b/source/entities/Building.cpp
@@ -275,7 +275,7 @@ double Building::takeDamage(GameEntity* attacker, double physicalDamage, double 
 
 std::string Building::getNameTile(Tile* tile)
 {
-    return getName() + "_tile_" + Ogre::StringConverter::toString(tile->x)
+    return getMeshName() + "_tile_" + Ogre::StringConverter::toString(tile->x)
         + "_" + Ogre::StringConverter::toString(tile->y);
 }
 
--- a/source/entities/Building.cpp
+++ b/source/entities/Building.cpp
@@ -27,6 +27,36 @@
 
 const double Building::DEFAULT_TILE_HP = 10.0;
 
+void Building::createMeshLocal()
+{
+    GameEntity::createMeshLocal();
+
+    if(getGameMap()->isServerGameMap())
+        return;
+
+    std::vector<Tile*> coveredTiles = getCoveredTiles();
+    for (Tile* tile : coveredTiles)
+    {
+        RenderRequest* request = new RenderRequestCreateBuilding(this, tile);
+        RenderManager::queueRenderRequest(request);
+    }
+}
+
+void Building::destroyMeshLocal()
+{
+    GameEntity::destroyMeshLocal();
+
+    if(getGameMap()->isServerGameMap())
+        return;
+
+    std::vector<Tile*> coveredTiles = getCoveredTiles();
+    for (Tile* tile : coveredTiles)
+    {
+        RenderRequest* request = new RenderRequestDestroyBuilding(this, tile);
+        RenderManager::queueRenderRequest(request);
+    }
+}
+
 void Building::addBuildingObject(Tile* targetTile, RenderedMovableEntity* obj)
 {
     if(obj == NULL)
@@ -152,6 +182,7 @@ void Building::addCoveredTile(Tile* t, double nHP)
 {
     mCoveredTiles.push_back(t);
     mTileHP[t] = nHP;
+    t->setCoveringBuilding(this);
 }
 
 bool Building::removeCoveredTile(Tile* t)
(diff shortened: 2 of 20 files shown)


Bertram25:
Heya @hwoarangmy :)

Seems you did some good code streamlining here. The new code is much more readable. I'll test asap.
(And normally, tonight I'll be finally able to test the removal of the render manager. XD)

akien-mga:
Hm, I'm missing a lot of email notifications from github lately :-/
Do you also have issues with it?

akien-mga:
> Hm, I'm missing a lot of email notifications from github lately :-/

I've changed my notification email (was using an alias that redirects to my actual email account, that might cause some issues).

Bertram25:
@akien-mga Just in case, all is fine on my own. I hope this will be ok for you.

hwoarangmy:
> Hm, I'm missing a lot of email notifications from github lately :-/

Well, I've disabled notifications because I was having too many (I had to delete 100 mails once). So I might not be a good example ^^

> Seems you did some good code streamlining here

I will also have a look if it is possible to unlink the ground tile name from the room name. That would allow to not have to destroy every tile from the oldroom and add the same ones from the new one when there is a room absorbtion.

Bertram25:
This will be for tomorrow evening for me. :) I need some sleep and tomorrow is very busy. See ya tomorrow.

akien-mga:
While you are working on rooms, could you have a look at this warning?

```
[ 84%] Building CXX object CMakeFiles/opendungeons.dir/source/rooms/RoomTrainingHall.cpp.o
/home/iurt/rpmbuild/BUILD/opendungeons-0.4.9/source/rooms/RoomTrainingHall.cpp: In member function 'virtual void RoomTrainingHall::doUpkeep()':
/home/iurt/rpmbuild/BUILD/opendungeons-0.4.9/source/rooms/RoomTrainingHall.cpp:253:42: warning: 'wantedY' may be used uninitialized in this function [-Wmaybe-uninitialized]
         if(creaturePosition.x == wantedX &&
                                          ^
/home/iurt/rpmbuild/BUILD/opendungeons-0.4.9/source/rooms/RoomTrainingHall.cpp:253:9: warning: 'wantedX' may be used uninitialized in this function [-Wmaybe-uninitialized]
         if(creaturePosition.x == wantedX &&
         ^
```

Bertram25:
I'll test now as well.

akien-mga:
This seems to work flawlessly, I did not spot any regression so far.

hwoarangmy:
@akien-mga ok, i will fix this. When you see stuff like that, please don't hesitate to give a word as my compiler do not complain for that. I will fix it in this PR if something is found and the next one if it is merged
BTW, the forge implementation is there and waiting for this PR to be integrated ^^

Bertram25:
Woohoo. Same here, had a full game without problems. Checked every room for tile addition and destruction. IMHO, it's good to go. Thus, I'll merge and the warning fix can get integrated in the next PR.

akien-mga:
> When you see stuff like that, please don't hesitate to give a word as my compiler do not complain for that.

Sure, usually I report warnings as soon as I see them. This one was introduced in a previous PR, but it's probably one of the few that I did not test before they were merged :-P

BTW I imported OD into Mageia as a 0.4.9 RC:
http://mageia.madb.org/package/show/release/cauldron/application/0/name/opendungeons
http://svnweb.mageia.org/packages/cauldron/opendungeons/current/SPECS/opendungeons.spec?view=markup
(based on the development branch from just before the merge of this PR). I'll update it as soon as the definitive 0.4.9 is out :-)

Note to self: I should split the data files into a "opendungeons-data" arch-independent subpackage, and specify the correct licenses for the data.

Bertram25:
> Sure, usually I report warnings as soon as I see them. This one was introduced in a previous PR, but it's probably one of the few that I did not test before they were merged :-P

Funny, since I've got warnings enabled, but it seems they weren't triggered. I'll recheck my flags...

Bertram25:
@akien-mga So cool to see OD in a distribution. Great work! :)

akien-mga:
> Funny, since I've got warnings enabled, but it seems they weren't triggered. I'll recheck my flags...

Note that I'm relatively "bleeding edge" when it comes to the compiler as I have gcc 4.9.2.

Bertram25:
> Note that I'm relatively "bleeding edge" when it comes to the compiler as I have gcc 4.9.2.

Even funnier, since me too. Let me do a few other things and I'll check that. :)

akien-mga:
Oh, I didn't think Debian would provide gcc 4.9.2 so soon :-)

akien-mga:
> Note to self: I should split the data files into a "opendungeons-data" arch-independent subpackage, and specify the correct licenses for the data.

Done. I also unbundled the FreeMono.ttf font. It would be nice to provide an "upstream" solution for that (i.e. if the font is present on the system, use it and do not install the one in gui/FreeMono.ttf), but for now I can't imagine an easy solution to do this (the font seems to be in different locations depending on the distro, font packaging seems to be a mess, at least in Mageia...).

Bertram25:
Ok, I've got my warnings back.

As fonts management, per experience, I'd say this should be handled per distributions and we should simply make sure to provide a font for easy windows packaging, and let the font be configurable in the data path so that packagers can change it when building their package (which is already the case).
Imho, there is nothing more to do on our side and if some font is provided by mageia, you should probably update your specs file to get rid of the font once installed and change the paths in the gui files to look for the correct system font.
Feel free to prove me wrong if you know about better practices. You're the true packager after all. :)

akien-mga:
> I'd say this should be handled per distributions and we should simply make sure to provide a font for easy windows packaging, and let the font be configurable in the data path so that packagers can change it when building their package (which is already the case).

Indeed, this is evident but I hadn't though about it :-)

Should I patch this file to link to the system wide font?
https://github.com/OpenDungeons/OpenDungeons/blob/development/gui/ogrefonts.fontdef

Or do you mean something else with "which is already the case"? Ideally we would have an opt-in CMake variable to which the font path could be passed (the default behaviour staying the current one of course, with the font bundled in gui/).

akien-mga:
Actually the (relative) path to the font dir has to be given to the resource manager through: https://github.com/OpenDungeons/OpenDungeons/blob/development/cmake/config/resources.cfg.in

I'll stop hijacking this PR and open an issue for this, I think we can manage a nice solution for packagers here (i.e. a cmake option that would add the path to the font dir to the resources.cfg when defined).

Issue #354, opened 2014-12-02
Implement fighter logic when not in a battle


hwoarangmy:
See http://forum.freegamedev.net/viewtopic.php?f=35&t=6013&sid=d77d962287d1146877e3351cf66df5f6

To summarize:
To have something easy to configure, I would propose to allow to add in the creature definition file a list of usable rooms and a value. The higher the value is the most likely the creature will go there if the room has empty space. For example, a troll would have something like:
Forge 5 5
TrainingHall 2 3

The first value is the creature likeness to the room. A value of 0 means that the creature won't go there alone.
The second value is the creature efficiency. A value of 0 means that the creature is not good in this room. A value > 0 will be handled by the room itself.

A value of 0 in efficiency will prevent a creature to work in a room (whatever its likeness is) even if dropped. If a creature has a likeness = 0 and an efficiency > 0 for a room, it won't go there by itself but it will agree to work if dropped.
By default (if a room is not present in the creature definition), the likeness will be 0 and efficiency will be 0.
If a creature has a likeness > 0 and efficiency = 0, he will go to the room to wander but won't use it (even if dropped)

PR #365, opened 2014-12-06
Implemented forge logic

hwoarangmy:
As discussed, the changes are:
- Every created trap is now deactivated by default. It needs a crafted trap from the forge to be activated
- Once activated the trap can shoot immediately
- After a few shoots (configurable per trap in the config files), the trap is deactivated and needs another crafted trap
- Between 2 shoots, there is a reload time
- Attackable traps can be attacked by creatures even when deactivated
- When created, crafted traps use one active spot in the forge until they are taken by a kobold

EDIT:
- I've added support in the level file to choose if a trap is activated or not at load time
- For now, no creature will go on the forge to work by themselves. However, if the randomly go there, they might consider to work. But unlike training hall, they won't go there for that purpose
- For now every creature is as efficient as the others at forge.

Note that the 2 last points will be fixed when #354 will be done


akien-mga:
Currently testing, it seems to work great. The small 2v2 map is a nice way to test that traps have a cooldown and get deactivated after a while. To check the forge crafting logic, you need to wait a bit more for the blue player to build a forge :-)

akien-mga:
It simply works flawlessly :-)

Adventurer and goblin, hard working at the forge to power up my boulders :-D
od-forge1

akien-mga:
I even saw a kobold drag a trap from the forge to its location :-D

akien-mga:
And it stacks:
od-forge2

akien-mga:
The only issue I see so far is the behaviour when dropping creatures on active spots: whether I drop a creature on the first active spot (link one on my screenshots) or the second (right one), they work on the first one, or leave if it's already taken. I can't forcefully assign a creature to the second active spot (but some creatures worked on it of their own will as shown in the first screenshot).

akien-mga:
I got a segfault at turn 1488. I was doing anything particular, I had just placed a new boulder trap a few turns before next to the enemy lines, and I was looking whether a kobold would go fetch my available boulder to activate the trap.

```
Server sends newturn=1488
CLIENT - Adding room Treasury40, nbTiles=1, seatId=2
Client received turnStarted=1488
CLIENT - Adding Creature Dwarf1164
CLIENT - Removing rendered object RenderedMovableEntity_Treasury6_[12,38]_374,MeshName=GoldstackLv3
CLIENT - Adding rendered object RenderedMovableEntity_Treasury__478,MeshName=GoldstackLv3
CLIENT - Removing rendered object RenderedMovableEntity_Crypt39_470,MeshName=SmallSpider
An internal Ogre3D error ocurred: OGRE EXCEPTION(4:ItemIdentityException): An object of type 'Entity' with name 'Room_Treasury_tile_12_38' already exists. in SceneManager::createMovableObject at /home/iurt/rpmbuild/BUILD/ogre_src_v1-9-0/OgreMain/src/OgreSceneManager.cpp (line 6832)
Internal Ogre3D exception: OGRE EXCEPTION(4:ItemIdentityException): An object of type 'Entity' with name 'Room_Treasury_tile_12_38' already exists. in SceneManager::createMovableObject at /home/iurt/rpmbuild/BUILD/ogre_src_v1-9-0/OgreMain/src/OgreSceneManager.cpp (line 6832)
*** ERROR: Internal Ogre3D exception: OGRE EXCEPTION(4:ItemIdentityException): An object of type 'Entity' with name 'Room_Treasury_tile_12_38' already exists. in SceneManager::createMovableObject at /home/iurt/rpmbuild/BUILD/ogre_src_v1-9-0/OgreMain/src/OgreSceneManager.cpp (line 6832)
Quitting main application...
```

source/rooms/RoomForge.cpp
```diff
@@ -52,6 +73,310 @@ RenderedMovableEntity* RoomForge::notifyActiveSpotCreated(ActiveSpotPlace place,
...
+        Ogre::Real wantedY = -1;
+        getCreatureWantedPos(creature, tileSpot, wantedX, wantedY);
+
+        RenderedMovableEntity* ro = getBuildingObjectFromTile(tileSpot);
+        OD_ASSERT_TRUE(ro != NULL);
+        if(ro == NULL)
+            continue;
+        Ogre::Vector3 creaturePosition = creature->getPosition();
+        if(creaturePosition.x == wantedX &&
+           creaturePosition.y == wantedY)
```
oyvindln: Using == with floating point values is generally not a good idea.

hwoarangmy: That's right. In this case, it's to check if the creature is in the destination we set. And there is some kind of computation while the creature moves so I guess there might be a problem if we are unlucky. I will have a look if we can use the tile only.

Bertram25:
@akien-mga @hwoarangmy About the crash: Weird, seems quite unrelated to the given change. ?
I'll let a few games run without the PR and we'll see.

hwoarangmy:
@akien-mga Do you have the complete log and the replay ?

hwoarangmy:
> I can't forcefully assign a creature to the second active spot

Well, by design, active spots are used in the creation order (or when released) so the same one will be used wherever you drop a creature.

akien-mga:
> Do you have the complete log and the replay ?

Here is the log: http://remi.verschelde.fr/files/misc/opendungeons.log
And the replay: http://remi.verschelde.fr/files/misc/replay_20141206_234911.odr

hwoarangmy:
@akien-mga Thanks for the log. I will have a look
BTW, it's the occasion to see if a replay from a computer works on another ^^

@Bertram25 I would think that the crash is more related to my previous PR about room absorbtion. But I will see on the log/replay ^^

source/rooms/RoomForge.cpp
Bertram25: True, nice catch.

Bertram25:
I'll try as well tomorrow, just for the fun or seeing @akien-mga in action. :)

hwoarangmy:
First, I've seen that @akien-mga is a cheater that gets free gold
Secondly, his dungeon is a mess ^^

More seriously, the good point is that I was able to watch the replay ^^

Concerning the bug itself, I've not made sure yet but my guess is that the player 2 (your ally, on your north) has got his treasury destroyed. Because the AI only had a 2 tiles treasury, it wanted to build another tile. And I think that there is a problem in the server that made the tile creation be sent before the tile destruction. And because of my last PR, the room name is not in the tile name anymore and ogre crashed.

The other interesting thing is that we can clearly see many creatures stuck. Especially from your AI ally. We should deal with that earlier than expected...

hwoarangmy:
I've changed what was requested. Concerning the bug, I confirm that the room tile creation is asynchronous while room destruction is synchronous.
So I think my guess was correct. I will see if I can find something after sleeping a bit ^^

But as @Bertram25 guessed, the bug is not related to this PR

BTW, the good thing is that when we will have a fix, we could use your replay to test it because it reproduces the crash ^^

Danimal696:
should i create a model for unbuilt traps or do we stick to smaller traps versions? i like the small versions, but i would made them 1/3 of what is hown.

hwoarangmy:
The crash is fixed. As expecred it was because the AI gets a treasury destroyed and tries to rebuild it within the same turn. Because the fix is on server side, @akien-mga 's replay should still crash.

akien-mga:
> First, I've seen that @akien-mga is a cheater that gets free gold
> Secondly, his dungeon is a mess ^^

Hehe, when I test new features, I tend to make shortcuts :-p

> The crash is fixed.

Great! :-) So to check I guess one would have to destroy a treasury tile of any AI?

hwoarangmy:
> Hehe, when I test new features, I tend to make shortcuts :-p

I tend to do the same ;-)

> Great! :-) So to check I guess one would have to destroy a treasury tile of any AI?

Yes. And the easiest is to disable treasury check cooldown so that the bug is systematic:
https://github.com/OpenDungeons/OpenDungeons/blob/development/source/ai/KeeperAI.cpp#L74

replace:
if(mCooldownCheckTreasury > 0)
by
if(false)

akien-mga:
Hm, I made the change you suggested but it does not seem to reproduce the bug: I rushed the blue AI on the 2v2 map to crush some treasury tiles (pro tip: kill the dragon before ;-) ), and it waited some turns before it built them again. Either way, the bug seems to be fixed :-)

akien-mga:
Hm, with the unmodified PR, it seems the AI might forget to rebuild its treasury tiles. It is not systematic though, so I think your fix is good, but maybe there are some conditions in the AI check for treasury tiles that are not always met? e.g. if the dungeon temple is partly destroyed, does the AI resign and stop trying to complete its stock of treasury tiles?

_Edit:_ Overall it works just fine, there might just be some corner cases where the logic stalls; i.e. if a treasury tiles is destroyed, rebuilt and destroyed again in a matter of seconds, it seems the AI doesn't try to rebuild it again.

hwoarangmy:
@akien-mga FYI, the logic is to check if the AI has at least 3 treasury tiles. If yes, it proceeds next actions. If no, it builds 1 tile (if possible next to 1 treasury). Then, it will not check for treasury for 10 to 30 turns (to avoid constantly building the missing tiles especially since the first one is free).

And concerning the dungeon temple, if it is destroyed, the AI is dead and won't do anything.

Bertram25:
I'll be testing now :)

Bertram25:
Can't see any flaws in the code, I'm recompiling and retesting right now! :D

hwoarangmy:
Just in case, if you ask yourself, the bug seen by @akien-mga cannot happen for a human player because human players messages are processed before the gamemap upkeep. That means that if a tile is going to be destroyed, no player commands are processed until messages are sent. So, if a player asks to build something on a tile that will be destroyed during the same turn, it will not be allowed (because at this time, the tile is already built). And if he asks the next turn, the server messages will already have been sent so even if the message is sent asynchronously, it will be received after the tile destruction.

Bertram25:
> Just in case, if you ask yourself, the bug seen by @akien-mga cannot happen for a human player because human players messages are processed before the gamemap upkeep. 

Indeed, I was watching that. Thanks for the explanation. :)

Bertram25:
Ok, I've taken all the time to confortably test that game changer. :D
First things first, it makes things much more strategical now. And you have to much more carefully choose how and when you make traps now.

Here a the few things I've noticed so far: If you confirm it isn't a bug I guess we can merge:
- As discussed, creatures won't forge automatically. You have to drop them here.
- When dropping several creatures there, they all try to forge the first needed trap. I mean, if you placed a cannon, a spike and a boulder, every creature will try to create a cannon, until you actually place the cannon trap by one of your workers. This is resulting with a lot of cannon ready usually.
- When moving a trap to the location, the trap appears one tile top-right of the worker. I wonder whether it is done on purpose. I would instead make it appear on top of it, and a little smaller than that. But this can be fine-tuned later on. :)
- I noticed that it is the same at least for corpse.

Ah, and this is a rare and really bad bug, I was beaten by the dark/light blue team while attempting to sneak in the light blue lair while they both were attacking the grey one in "Scream in the Dark..." ;)
I need to fix that one.

hwoarangmy:
> As discussed, creatures won't forge automatically. You have to drop them here.

Yep, that will be fixed when #354 will be done

> When dropping several creatures there, they all try to forge the first needed trap

Actually, there is no random at all. To know what to forge, we take all the traps and break at the first one in need. That means that while the first one has not got its trap, forge will produce for it; Moreover, the trap consider it doesn't need a crafted trap anymore when a kobold starts to bring one. That means that right after a crafted trap is done, because no kobold is carrying a trap at this moment, the creatures will start to craft another one of the same type.
IMHO, we should add some random there so it is not as predictable. But that's for another PR ^^

> When moving a trap to the location, the trap appears one tile top-right of the worker

I've just attached the carried node to the kobold's node. I've just seen that it was not too messy and let it as is. But if you have a better way to do that, please don't hesitate.

> I noticed that it is the same at least for corpse

Yep, that's the same code ^^

So, IMHO, this can be merged ^^

akien-mga:
Merging then :-)

Bertram25:
Yep, it's perfectly working and we can improve later anyway :)

PR #368, opened 2014-12-09
Removed render manager queue + code conventions on gamemap + creature vision debug

hwoarangmy:
To test the creature vision debug, open the console and write "visdebug creatureName"
That will display every tile the given creature sees. Note that it clearly shows that there is a problem on how visible tiles are computed (especially on a 1 tile wide corridor).
Note also that this command is available on the server only. And if you use it, every client will see the visible tiles.

If the creature whose vision is displayed is pickedup, the tiles won't be displayed until it is dropped.

### DIFF
--- a/source/entities/Building.cpp
+++ b/source/entities/Building.cpp
@@ -37,8 +37,8 @@ void Building::createMeshLocal()
     std::vector<Tile*> coveredTiles = getCoveredTiles();
     for (Tile* tile : coveredTiles)
     {
-        RenderRequest* request = new RenderRequestCreateBuilding(this, tile);
-        RenderManager::queueRenderRequest(request);
+        RenderRequestCreateBuilding request(this, tile);
+        RenderManager::executeRenderRequest(request);
     }
 }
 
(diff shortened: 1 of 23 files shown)


akien-mga:
The vision debug feature is awesome :-)

I notice that the display of the vision debug is not reset when starting a new map (with a debug feature activated in the previous map).

Apart from that, everything runs fine :-)

Bertram25:
Visual debugging FTW :) I'll test asap.

hwoarangmy:
> RenderManager::RequestCreateBuilding(parameter...) and so on instead of creating render requests and pass them to the manager

If we do that, every function from the render manager would have to be public. But IMHO, it would be better and if we want to use a queue again, we can wrap directly within the render manager (which is not worst IMHO).

> I notice that the display of the vision debug is not reset when starting a new map (with a debug feature activated in the previous map).

Yes, it's more like a debug feature. I've not tested but I think that if the creature dies, it won't be removed either. I've just made it work because I've seen strange behaviours when creature were fighting (sometimes, they don't see each other when in a corridor)

Bertram25:
It's working fine and I saw nothing in the code, I'll merge.

PR #373, opened 2014-12-13
Added console commands to see visible tiles for a seat and to allow creatures to spawn at levels > 1

hwoarangmy:
To display visible tiles for seat 1:
seatvisdebug 1

To change level of creature BigKnight1 to 15:
setlevel BigKnight1 15

Note that it will also allow to downgrade creature level. This command might help us to balance, too.

### DIFF
--- a/config/console_commands.txt
+++ b/config/console_commands.txt
@@ -35,10 +35,12 @@ possescreature
 refreshmesh
 rotatespeed
 save
+seatvisdebug
 setcamerafovx
 setcamerafovy
 setcolor
 setdest
+setlevel
 startcreatureculling
 starttileculling
 switchpolygonmode
--- a/levels/skirmish/TestSingleplayerSmall.level
+++ b/levels/skirmish/TestSingleplayerSmall.level
@@ -944,10 +944,10 @@ ProtectDungeonTemple	NULL
 3	Tentacle1	Tentacle11	1	0	max	100	0	0	5	15	0	none	none
 [/Creature]
 [Creature]
-3	Tentacle1	Tentacle12	1	0	max	100	0	0	5	11	0	none	none
+3	Tentacle1	Tentacle12	10	0	max	100	0	0	5	11	0	none	none
 [/Creature]
 [Creature]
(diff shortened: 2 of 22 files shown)


Bertram25:
Nice features. :)
I'll test asap and I have got a few comments, as usual. ;)

akien-mga:
Fine to merge as far as I'm concerned.

hwoarangmy:
@Bertram25 Tell me if you want me to disable visible tile computation before FOW is implemented.

I am currently having a look if it can be done on server side (that would be the best since it would avoid easy cheating by disabling FOW on client side).

Bertram25:
> I am currently having a look if it can be done on server side (that would be the best since it would avoid easy cheating by disabling FOW on client side).

Ah, so here was your evil plan ;D I see why you let it computed now. My next question will be: What is the performance change because of this? If it is negligible, then it's great as it means computations for the FOW won't cost that much and then my comment can be ignored. I'll test asap and see what happens.

hwoarangmy:
> What is the performance change because of this

Well, we go through all the tiles to check which are claimed by who and add them (and their neighboors) in the visible tiles list of the concerned seat if claimed. That can eventually be improved by changing the claim list only when a tile is claimed. But my guess is that it shouldn't be too costly. ANd the good point is that is pretty simple to implement.
Concerning the tiles the creatures sees, I've splitted the creature visible tiles computation to allow filling visible tiles at the same time. So there is almost no extra cost about it (just adding the tiles to the list).

Bertram25:
> I am currently having a look if it can be done on server side (that would be the best since it would avoid easy cheating by disabling FOW on client side).

Oh yeah, I think it's a sub-goal handy to look for. :)

Bertram25:
Btw, it seems to me you didn't added the new commands to the commands text file. This prevents auto completion when typing the command name using tab.

hwoarangmy:
yep, i forgot. I will add it

Bertram25:
Works like a charm, and the consequences on the FPS goes unnoticed on my side.
I think we can merge. Wanna add the missing commands before?

hwoarangmy:
done ^^

akien-mga:
Looks good, I merge.

Issue #367, opened 2014-12-07
Improvements to resources.cfg.in


akien-mga:
Talking about the resources.cfg.in as of https://github.com/OpenDungeons/OpenDungeons/blob/2f2a9573377f589b6298525653ca44859d24af0a/cmake/config/resources.cfg.in

There might be some enhancements to do in this file, but as I don't really know OGRE's resource management, I open this issue as a RFC (request for comments).

What I see:
- `@OPENDUNGEONS_MEDIA_DIR@/` is useless as far as I can tell: we don't define it in CMake, and @Bertram25 implemented the "out of source" media dir in an other way. We should probably drop it then. If so, should we keep it as a file to configure in cmake/config?
- Due to the implementation of the data dir in the code, there doesn't seem to be an easy way to define absolute paths in this file, since all paths are prefixed with the actual data dir (e.g. `./` or `/usr/share/games/opendungeons/`). So if I want to add a system font dir as a resource, I need to use a path relative to the data dir, and that's a bit annoying to compute (since some distros might install OD in /usr/share/opendungeons and not /usr/share/games/opendungeons).
  Is there a way to achieve absolute paths in the resources.cfg syntax, or a change that could be made in the code to respect leading slashes as a mark of the root filesystem? Linked to #366.
- Is the file up to date? I see for example that there are three subdirectories linked in materials/RTShaderLib, but the repo contains more: GLSL150, GLSLES, HLSL.
- Do we want to keep materials/RTShaderLib/Cg? AFAIK we dropped the Cg plugin because it's nonfree.

Bertram25:
1. Yes, IMO we should be able to drop @OPENDUNGEONS_MEDIA_DIR@/ and related stuff.
2. Completely duscussed in #366 Relative paths will have to be used for now. Either Ogre will move or we will have to keep using relative paths or add a custom resource loader.
3. & 4. I updated the RTShaderLib/ folder some time ago to the latest version. AFAIK, the Cg, GLSL150, GLSLES and HLSL folders can be dropped and the cg path removed from the plugin(_d).cfg path.

I'll make a PR and hopefully this won't break stuff for everyone.

oyvindln:
Removing the HLSL folder will cause problems for people using Direct3D on windows. The GLSL150 folder may be used depending on OpenGL version, so I don't know if that should be removed. Cg is disabled, so we don't need that, and I would assume we don't have anyone using the game on an OpenGL ES platform so that is probably safe to remove too.

[...]

Issue #22, opened 2014-07-29
Editor: Add support to create a new level or load an existing one.


Bertram25:
- [x] At the moment, the editor can only open the `levels/Test.level` as it is hard-coded.
  A loading menu should be added so that one can choose to load an existing level.
- [x] A menu should also be added to create a new one.
- [ ] In the editor, a menu should be added to save the level and setup its common parameters:
- Level non-translated name
- Level non-translated description
- Made for multiplayer? (boolean)
- Can be used in Skirmish (if multiplayer)
- Team/AI slot setup
- Tileset used.

Bertram25:
The possibility to select a level has been added in #157 #158 
The possibility to create a new one is much less urgent as a level can be copied, renamed and reused.
What we should solve first is the fact that a user won't be able to edit levels on a packaged version of OD on linux, or on windows without administrator's rights. (For now it's already great to be able to select a level though.)

I do think making the game follow the common policies depending on the platform (#25) and then let the game list user-made levels in its custom data folder is the next step.

Bertram25:
@akien-mga Btw, when you'll package OD, the editor won't be able to save maps, as they will be in /usr/share/... 

Once #25 is done, do you have any idea on how we should deal with custom maps against bundled ones?
Or maybe we should only let the editor able to change maps in the user data dir?

akien-mga:
As stated previously, we should add support (using boost I guess) for multiple asset location, at first ~/.local/share/opendungeons/ and /usr/share/games/opendungeons/ (and maybe ./ too to cater for in-source builds?). The user diretory should be checked first, and the system directory afterwards. This way files that are found in the user directory will be loaded instead of the official system ones. Then we can make it so that the save function of the editor saves to the user directory in ~/.local/share/opendungeons/levels/

Bertram25:
Fine by me.
I'd indeed use the given order you told: 
1. ~/.local/share/opendungeons/ or ApplicationData and such on Windows (Depends on the version)
2. ./ for developpers (is the same for windows anyway.)
3. /usr/share/games/opendungeons/ of in general the installation data path.

akien-mga:
The second point in the opening comment would be a great feature to have for 0.5.0, as it would let generate arbitrary sized levels easily (for now it seems to involve a lot of manual editing unless I'm mistaken).

BTW should we keep the "Map Editor" milestone? IMO it makes it a bit harder to keep track of the editor related features that we want for a given versioned milestone.

[...]

PR #377, opened 2014-12-14
Working on fixing #367 + misc stuff

Bertram25:


oyvindln:
Are you sure about removing the possibility to use D3D on windows?

akien-mga:
Looks fine as far as I'm concerned.

hwoarangmy:
I will try with windows/mingw tonight

hwoarangmy:
I've tried and there are problems...
The tiles you can see in the screenshot are only normal claimed tiles (I've built nothing)
odscreenshot_2014-12-14_221440_1

akien-mga:
Is it better if you readd `materials/RTShaderLib/Cg` in resources.cfg?

hwoarangmy:
I've made some tries. From what I've seen, its not about materials/RTShaderLib/Cg.

If I remove materials/RTShaderLib/GLSL150, I see no more glitches.

Bertram25:
> Are you sure about removing the possibility to use D3D on windows?

@oyvindln While it is true that the HLSL shaders + the possibility to use the Direct X renderer is handled by Ogre on Windows, I wondered why we would like to keep it.
I mean if we're going cross-platform, there is a point where we'll definitely use more precise shaders to improve the visuals, and using two different renderers will be twice the cost in test and development.

That said, I'm not willing to be blunt here and if you have good reasons to keep it, I can readd it back.

> If I remove materials/RTShaderLib/GLSL150, I see no more glitches.

Ok, then the GLSL150 should be removed and disabled, I guess. It was before and it seems nobody complained about OD not starting so far.

> Is it better if you readd materials/RTShaderLib/Cg in resources.cfg?

@akien-mga As far as I've seen, the Cg is required as a dependency on Windows for some reasons. (Ogre does complain about some files from it being missing), but doesn't seem to be needed in the resource data directly somehow.

I don't want to play with fire here, just removing what's actually not used, so here is what I'll do:
- Readd Cg/ in the resources file to stay logical with what is used.
- Remove GLSL150 files and from the resources list to force the GLSL folder to be used, even when GLSL150 support exists on the hardware/driver.

The question open is: Should I readd HLSL shaders and Direct X renderer support on windows?

akien-mga:
> As far as I've seen, the Cg is required as a dependency on Windows for some reasons. (Ogre does complain about some files from it being missing), but doesn't seem to be needed in the resource data directly somehow.

It sounds like Ogre tries to load it even though we don't use it, that's probably something that we should look into in the future. For now it's no big deal, those files are not heavy and I guess they are not nonfree like Cg :-)

> The question open is: Should I readd HLSL shaders and Direct X renderer support on windows?

I don't know much about shaders technologies, so I can't really comment. But your point about being cross-platform makes me think that we can let go of DirectX and focus on the OpenGL renderer. OpenGL is well established even in the Windows gaming sphere now, and unless their is a real gain in implementing DirectX support, I'd say let's focus on the open source and cross platform one :-)

Bertram25:
Ok, the DirectX question remains. The rest is done. :)

hwoarangmy:
That's ok with me ^^

Bertram25:
Hi there, :)

I added 'King of the Hill' yesterday. It is playable but I'm still rather unhappy with it.
As is, the map will fit more a Human vs Human battle, while I should add a few more twists.

I'll also make it smaller for the single-Player version to make it more interesting and let the AI react better.

akien-mga:
Awesome, I'll try it out and make sure that I don't design a similar map (I was planning on making one today). Note that the map in your commit is named "The Bridge", is it the same as "King of the Hill"?

akien-mga:
Since yours is 1v1, I plan to go for a medium to small-sized (80x80 or something) 1v1v1 map (like a triangle).

Bertram25:
> Awesome, I'll try it out and make sure that I don't design a similar map (I was planning on making one today). Note that the map in your commit is named "The Bridge", is it the same as "King of the Hill"?

Oops, s/King of the Hill/The Bridge/

The King of the Hill map, is another idea for later. ;)

Cool to see you're on your own map. :) Once my PR is in, I guess we'll be able to go on with the RC.

akien-mga:
What is the easiest way to generate an empty level of X*Y dimensions?
I think a "Create new level" button which lets you type in X and Y dimensions and generates a level filled with dirt would be a nice feature for 0.5.0, I'll open an issue :-)

_Edit:_ Already part of #22

Bertram25:
Yeah, all that should be discussed in #22 :)

hwoarangmy:
> What is the easiest way to generate an empty level of X*Y dimensions?

IMHO, the easiest is to copy a map (the smallest), remove all the tiles (within [Tiles][/Tiles]) and change the map size (also within [Tiles][/Tiles]). Then, remove the rooms (so you can add them where you want).
Because in the editor, it is not possible to add dungeon heart/portal, I would add another room, like a crypt, and then edit the level manually to replace the crypt by dungeon heart/portal.

akien-mga:
> Because in the editor, it is not possible to add dungeon heart/portal, I would add another room, like a crypt, and then edit the level manually to replace the crypt by dungeon heart/portal.

Thanks for the tip, that saves some time :-)

akien-mga:
About the Bridge: I seem to win instantly when I start it in singleplayer.

Bertram25:
> About the Bridge: I seem to win instantly when I start it in singleplayer.

Ah? weird. I'll have a look.

hwoarangmy:
> Thanks for the tip, that saves some time :-)

Welcome ^^

hwoarangmy:
Concerning the winning problem, I confirm I also have it. I guess it's coming from the fact that the goal is to kill every creature and at game start, there are none...
Might be cool to add "Destroy all ennemy buildings"

Bertram25:
> Might be cool to add "Destroy all ennemy buildings"

Yep, was thinking the same. I'll do that. Has anyone news from @oyvindln concerning the Direct X question?

Bertram25:
> Might be cool to add "Destroy all ennemy buildings"

After thinking about this, I do believe the Kill all enemy objectives should take in account enemy rooms spawning creatures, I'll add that.

Bertram25:
N.B.: Adding the objective to destroy every room as a separate issue is a good idea. I may create it if everyone agrees.

Bertram25:
The "You won" bug is fixed.

Bertram25:
I will also take the time to tweak the level size a bit to make it more active. :)

hwoarangmy:
ok. I will test when you are done

Bertram25:
Ok, I should be done both for the map and the Objective text.

Bertram25:
May I request we merge this even if @oyvindln hasn't answered yet, since IMHO, we're going on the OpenGL way. I'll be able to move on with the RC, then.

akien-mga:
Looking good overall, merging.

Bertram25:
Good, my turn to bring the RC and corresponding git stuff now :)

PR #378, opened 2014-12-14
Add a man page and install it on Linux (fixes #357)

### LINKED ISSUE #357: Add a manpage
We should provide a manpage for the opendungeons binary (typically this would be a requirement in Debian's packaging policy if we want OD included in Debian).

BTW, are there parameters that can be used with the OD binary? `opendungeons --help` starts the game normally.

akien-mga:
This commit adds a man page to $prefix/share/man/man6/opendungeons.6 ("6" being the category for games man pages).
I added the paths to the installation and doc folders in the man page, so it has to be configured via CMake (that's debatable, I can also drop the mention and make it a static file in dist/).

I also added instructions to install CREDITS and FAQ.txt to the typical doc dir on Linux: $prefix/doc/<name>. I did not install the README since it's aimed at developers (describes the source code and how to compile) and the source is not included in system installations).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -560,6 +560,7 @@ else()
     set(OD_ICON_FULLPATH ${OD_SHARE_PATH}/icons/hicolor/scalable/apps/opendungeons.svg)
 endif()
 configure_file(${CMAKE_CONFIG_DIR}/opendungeons.desktop.in ${CMAKE_BINARY_DIR}/opendungeons.desktop)
+configure_file(${CMAKE_CONFIG_DIR}/opendungeons.6.in ${CMAKE_BINARY_DIR}/opendungeons.6)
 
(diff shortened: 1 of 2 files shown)


akien-mga:
The description is based on http://opendungeons.sourceforge.net/?q=about
We might want to improve it eventually.

akien-mga:
Man page based on this template btw: http://www.tldp.org/HOWTO/Man-Page/q3.html

akien-mga:
I'll fix the AUTHOR section, I don't think it should (only) reference the manpage writer but more the whole development team.

_Edit:_ Done.

hwoarangmy:
I've tried on windows and seen no regression ^^

Bertram25:
I think this can be merged. :)
Thanks for the work done @akien-mga

PR #384, opened 2014-12-16
New skirmish and multiplayer level Greed or Might (FFA 3 seats)

akien-mga:
Another bonus level just before the release ;-)

I also added a clean template for level creation.


Bertram25:
I'll test asap. You're right, let's use the template until we have automated level creation.

hwoarangmy:
Cool a new level ^^
I will test asap

hwoarangmy:
I've tested your map, it's pretty nice ^^
I understand the name :)
But for the AI, it is better to link the portal with the dungeon heart because it might not do it by itself...

akien-mga:
>  But for the AI, it is better to link the portal with the dungeon heart because it might not do it by itself...

You have a point, I'll do that.

hwoarangmy:
Nice map. I like the idea ^^
I will let @Bertram25 have a look and merge if he takes too long :)

Bertram25:
Yeah. I must admit Christmas does wonders with my spare time. I'll test tonight, though.

Bertram25:
Excellent Level :)

akien-mga:
Note that I haven't tested it thoroughly yet, there might be too much/too little gold, and I don't know if it's manageable to beat the boss dragons (it also means that one has to claim the way up to a dragon to be able to engage into fight with something else than kobolds).

But I think it could be a nice level to test in LAN, we might see some funny strategies :-D

Bertram25:
Unless the dragons are lvl 30, I think it's feasible. I'm in the middle of it while compiling atm ;)

PR #388, opened 2014-12-18
Fix library tooltip and balance Greed or Might

akien-mga:
Slightly increased the level of the dragons, but also nerfed their HP regen and upped necessary XP/level
Also set all tiles with fullness>0 to fullness 100
And I removed three gold tiles with fullness 0 that were a mistake and should have been plain old dirt :-)


Bertram25:
Working as expected :) I'll merge.

PR #392, opened 2014-12-19
Fix doc installation

akien-mga:
CMake doesn't understand the bash syntax to list several files in the same dir it seems, so I'm using another method :-)

The previous implementation gave this error:

> CMake Error at cmake_install.cmake:122 (file):
>  file INSTALL cannot find
>  "/home/iurt/rpmbuild/BUILD/opendungeons-0.4.9-rc1/{AUTHORS,CREDITS,FAQ.txt}".


Bertram25:
As expected, working fine, the three files are installed. I'll merge.

PR #405, opened 2014-12-29
Several fixes/improvements

hwoarangmy:
I think  6f2b8f3 fixes the segfault seen by @akien-mga during our lan (AFAIK, it's the only place calling Ogre::AnimationState::addTime from ODClient::processOneClientSocketMessage).

### DIFF
--- a/source/entities/Creature.cpp
+++ b/source/entities/Creature.cpp
@@ -629,7 +629,6 @@ void Creature::doUpkeep()
 
             // Remove the creature from the game map and into the deletion queue, it will be deleted
             // when it is safe, i.e. all other pointers to it have been wiped from the program.
-            LogManager::getSingleton().logMessage(getGameMap()->serverStr() + "Removing creature " + getName());
             getGameMap()->removeCreature(this);
             deleteYourself();
         }
--- a/source/entities/Creature.cpp
+++ b/source/entities/Creature.cpp
@@ -3021,7 +3021,9 @@ void Creature::clearActionQueue()
     mActionQueue.clear();
     stopJob();
     stopEating();
-    releaseCarriedEntity();
+    if(mCarriedEntity != nullptr)
+        releaseCarriedEntity();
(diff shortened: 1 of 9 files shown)


hwoarangmy:
@akien-mga FYI, I've tried to look at your replay but it doesn't reproduce the segfault. But it stops 3 turns earlier than the log you posted so I guess we lost the part that caused the segfault. However, as I said, I think it is fixed (even if I cannot test)

Bertram25:
df3f5b3 is interesting. I'll take more time to study it. :)
If you're reasonably sure about yourself when fixing the segfault, we should close the issue.
Let's reopen a new one if the segfault comes back.

hwoarangmy:
> If you're reasonably sure about yourself when fixing the segfault, we should close the issue.

I'm reasonabily sure it fixes the segfault @akien-mga saw during our lan. Which is not the same as the issue he reported (AFAIK, there is no issue for this one).
But actually, I will try to let some games for maybe 2 hours or more and see if I can reproduce. If you and @akien-mga could also do the same, we could close the issue if none of us can reproduce.

hwoarangmy:
BTW, I think that the segfault I got in the second game of our lan was the same as @akien-mga got.

akien-mga:
This looks like a safe changeset. I will be away for the next two days, so feel free to merge without me having tested it all :-)

Bertram25:
Looks sane for the code. I'm testing atm.

Bertram25:
I've tested a lan toggling vision on the  75x75 and 400x400 maps, it takes time to send the tile info on the big one, but once it's done, the game plays fairly well. :D 

We'll need to test that on a wider network but it's looking great so far! \O/

Ah, and I confirm chat is working again.
I'll test the kobold carrying stuff tomorrow. ;)

Bertram25:
I've tested the kobold carriers, it's looking good. :)

PR #409, opened 2014-12-30
Some fixes...

Bertram25:
Everything is in the commits messages. :)


hwoarangmy:
ok. Then I merge as I've seen no regression, only better or same ^^

PR #410, opened 2014-12-31
NULL->nullptr commit + some more work...

Bertram25:
- I only change NULL into nullptr in OD source files, neither in AS ones nor in level files, for the record.
- I also fixed a member name not following conventions in the GUI class, and removed a useless friendship
  with the minimap.


hwoarangmy:
The PR works and even if there is still some work to do, it is enough for me. I merge

Bertram25:
@hwoarangmy Ok, I'll add a PR removing the mentioned console commands. :)

@oyvindln First sorry for interfering with your own work, I do hope I'm not damaging it, but I thought your PR was more about using CEGUI for the console, right?
Secondly, ok, I'll work now on moving the commands and their description back within the code and in the console class only so we have everything at the same location.
I hope porting the console to CEGUI will be easier afterwards. :)

oyvindln:
@Bertram25 I will have to redo some of the work anyhow. (Unless someone else does it first.) It was both about using a cegui window, and cleaning up the console code. If you look at how it's currently done, it seems half-finished and is a bit of a mess. (E.g, the console prompt is a public variable in the frame listener class.)

PR #414, opened 2015-01-03
Fixed memory leak and some warnings

oyvindln:
Fixed a memory leak.
Started adding more compiler warnings ("-Wall" leaves out many, see https://stackoverflow.com/questions/5088460/flags-to-enable-thorough-and-verbose-g-warnings/9862800#9862800 ) and added SYSTEM to include directories to avoid having warnings trigger in header files. This also allow us to use "-Werror". 
New warning are disabled for now, as many of them will trigger, my though is that we can incrementally add them and fix the occuring warnins.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -349,16 +349,16 @@ find_package(OGRE REQUIRED)
 find_package(CEGUI REQUIRED)
 find_package(SFML 2 REQUIRED COMPONENTS Audio System Network)
 
-if ((OGRE_VERSION_MAJOR LESS 1) AND (OGRE_VERSION_MINOR LESS 9))
-    message(SEND_ERROR "OGRE version >= 1.9.0 required")
+if((OGRE_VERSION_MAJOR LESS 1) AND (OGRE_VERSION_MINOR LESS 9))
+    message(FATAL_ERROR "OGRE version >= 1.9.0 required")
 endif()
 
-if ((CEGUI_VERSION_MINOR LESS 8) OR (CEGUI_VERSION_MAJOR LESS 1))
-    message(SEND_ERROR "CEGUI version >= 0.8.0 required")
+if("${CEGUI_VERSION}" VERSION_LESS "0.8.0")
+    message(FATAL_ERROR "CEGUI version >= 0.8.0 required")
 endif()
 
 if (SFML_VERSION_MAJOR LESS 2)
-    message(SEND_ERROR "SFML version >= 2.0 required")
+    message(FATAL_ERROR "SFML version >= 2.0 required")
 else()
     message(STATUS "SFML include directory: ${SFML_INCLUDE_DIR}; SFML audio library: ${SFML_AUDIO_LIBRARY_DEBUG} ${SFML_AUDIO_LIBRARY_RELEASE}")
 endif()
@@ -473,8 +473,8 @@ target_link_libraries(
     ${OGRE_RTShaderSystem_LIBRARIES}
     ${OPENAL_LIBRARY}
     ${OIS_LIBRARIES}
-    ${CEGUI_LIBRARIES_RELEASE}
-    ${CEGUI_LIBRARIES_DEBUG}
+    ${CEGUI_LIBRARIES}
+    ${CEGUI_OgreRenderer_LIBRARIES}
 )
 
 # Set linker options in MSVC
@@ -496,16 +496,7 @@ endif()
 # Link sfml
 # No need to make a difference when release/debug is found because even
 # if only one is found, the other is set to the same value
-target_link_libraries(
-    ${PROJECT_BINARY_NAME}
-
-    optimized ${SFML_AUDIO_LIBRARY_RELEASE}
-    optimized ${SFML_SYSTEM_LIBRARY_RELEASE}
-    optimized ${SFML_NETWORK_LIBRARY_RELEASE}
-    debug ${SFML_AUDIO_LIBRARY_DEBUG}
-    debug ${SFML_SYSTEM_LIBRARY_DEBUG}
-    debug ${SFML_NETWORK_LIBRARY_DEBUG}
-)
+target_link_libraries(${PROJECT_BINARY_NAME} ${SFML_LIBRARIES})
 
 ##################################
 #### Unit testing ################
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -99,6 +99,8 @@ endif()
 #### ExplicitCompilerFlags #######
 ##################################
 
+# TODO: Use add_compiler_options instead of setting CMAKE_CXX_FLAGS
+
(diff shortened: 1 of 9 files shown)


hwoarangmy:
I've tested and it seems Random.cpp is missing a #include <ctime>

Apart from that, I've seen no problem with compilation. I still have the same warnings I had before. After adding the missing include, this PR is mergeable for me. I will let @akien-mga and @Bertram25 test too because we know cmakefile is pretty sensible ^^
Moreover, I already know that the included debug CEGUI library will break @Bertram25 env. But last time, he said he would try to find why ^^

oyvindln:
Ahh, removing ctime was unintentional. I will make a new commit in a bit.

akien-mga:
It builds fine here (release build), I'll be testing soon.

akien-mga:
It works fine too :-)

CMakeLists.txt
```diff
@@ -340,16 +349,16 @@ find_package(OGRE REQUIRED)
...
 
-if ((OGRE_VERSION_MAJOR LESS 1) AND (OGRE_VERSION_MINOR LESS 9))
-    message(SEND_ERROR "OGRE version >= 1.9.0 required")
+if((OGRE_VERSION_MAJOR LESS 1) AND (OGRE_VERSION_MINOR LESS 9))
+    message(FATAL_ERROR "OGRE version >= 1.9.0 required")
 endif()
 
-if ((CEGUI_VERSION_MINOR LESS 8) OR (CEGUI_VERSION_MAJOR LESS 1))
-    message(SEND_ERROR "CEGUI version >= 0.8.0 required")
+if("${CEGUI_VERSION}" VERSION_LESS "0.8.0")
```
Bertram25, line 356: Ah? I didn't know about VERSION_\* . good to know. :)

CMakeLists.txt
```diff
@@ -397,25 +406,22 @@ include_directories(
     #OpenDungeons includes
     ${CMAKE_SOURCE_DIR}/source
 
-    #General external includes
-    ${DEPENDENCIES_DIR}
-
     #Angelscript includes
-    ${ANGELSCRIPT_DIR}/angelscript/include
-    ${ANGELSCRIPT_SRC_DIR}
+    SYSTEM ${ANGELSCRIPT_DIR}/angelscript/include
```
Bertram25, line 410: What is `SYSTEM`?

CMakeLists.txt
```diff
@@ -439,6 +445,11 @@ endif()
 # Create the angelscript library file
 add_library(${AS_LIBRARY_NAME} ${AS_SOURCEFILES})
 
+# Treat warnings as errors for main build, but not angelscript, since it would fail.
```
Bertram25, line 448: Cool, it will help.

source/network/ODClient.cpp
```diff
@@ -879,6 +879,7 @@ void ODClient::processClientNotifications()
                 sendToServer(event->mPacket);
                 break;
         }
+        delete event;
```
Bertram25, line 882: Nice catch. :)

source/utils/Random.cpp
```diff
@@ -36,58 +30,35 @@
 unsigned long myRandomSeed;
 const unsigned long MAX = 32768;
 
-unsigned long randgen()
+static unsigned long randgen()
 {
     myRandomSeed = myRandomSeed * 1103515245 + 12345;
-    unsigned long returnVal = (unsigned int) (myRandomSeed / 65536) % MAX;
+    //TODO: What is the purpose of the cast?
```
Bertram25, line 36: Avoid returning decimal values?

source/utils/ResourceManager.cpp
```diff
@@ -49,7 +49,7 @@
 #include <OgreRenderTarget.h>
 
 template<> ResourceManager* Ogre::Singleton<ResourceManager>::msSingleton = 0;
-#if defined(OD_DEBUG)
+#if OGRE_PLATFORM == OGRE_PLATFORM_WIN32 && defined(OD_DEBUG)
```
Bertram25, line 52: Here, why couldn't the application be compiled in debug mode on linux?
I do think this should be the same behaviour for every platform.
Thus, I'd remove the `OGRE_PLATFORM == OGRE_PLATFORM_WIN32 &&` part again and would remove the confusing comment as well.

What do you think?

Bertram25, line 52: > Because the plugins doesn't seem have the _d suffix on linux systems. (As the debug info is separate). They do in the windows SDK.

I just saw that. Nevermind. Funnily, other's comments have moved in the PR only, so I missed them.

Bertram25:
Looks fine. :)

I'll test asap.

> Moreover, I already know that the included debug CEGUI library will break @Bertram25 env. But last time, he said he would try to find why ^^

Yeah, if the problem happens again, I'll deal with it on my own. :)

Bertram25:
Ah, also, I do think this PR should be merged before mine once tested. (And I'll rebase if needed.)

CMakeLists.txt
oyvindln, line 410: > added SYSTEM to include directories to avoid having warnings trigger in header files

source/utils/Random.cpp
oyvindln, line 36: It already does integer division without the cast.

CMakeLists.txt
Bertram25, line 410: Ah ok, just found it:
http://www.cmake.org/cmake/help/v3.0/command/include_directories.html

source/utils/Random.cpp
Bertram25, line 36: Then, IIRC, when A / B, B is promoted to the type of A. This means 65536 is promoted as an unsigned long and maybe the cast does put this back to an unsigned int afterwards to play fine with `% MAX.` or have better computation performance maybe?
Otherwise, I have no idea why.

hwoarangmy, line 36: I can see 2 possible reasons:
- To strip the high bits. Maybe is it better for the algorithm to have only 32 bits (on most architectures)
- returnVal used to be an unsigned int and there was a cast to avoid a warning. Then, when returnVal  became unsigned long, because no warning was triggered from smaller to bigger type, it stayed unseen.

akien-mga:
It seems everyone is happy with this PR :-)

source/utils/Random.cpp
Bertram25, line 36: > returnVal used to be an unsigned int and there was a cast to avoid a warning. Then, when returnVal became unsigned long, because no warning was triggered from smaller to bigger type, it stayed unseen.

It must be that then. Casts are rarely so clever without a comment, or at least I hope. ;)
I'd propose to remove that cast and see whether it shows any problem while we're in early 0.5.0 development. WDYT guys?

PR #416, opened 2015-01-03
Implemented Persistent objects

hwoarangmy:
Persistant objects are rendered objects associated with a tile that are visible on client side even when vision on the tile is lost. They stay visible even if destroyed on the server gamemap. They get refreshed only when the cient gets vision on the associated tile.
For now, only Portals and dungeon temple uses them. But soon, traps will too.


oyvindln:
It's actually spelled "persistent"

hwoarangmy:
oups ^^

akien-mga:
I've tested locally, it works fine. I couldn't really test the "refreshing" part, since it implies to play with at least 3 players, to scout the base of one player and then to let him get attacked by the second player, so that the status of the temple and portal can change without you having vision on it.

hwoarangmy:
> I couldn't really test the "refreshing" part, since it implies to play with at least 3 players

FYI, I've changed the 2v2 small skirmish test map. In the bottom right seat, there are 2 claimed tiles to let you drop units to get vision and 3 BigDragons that are attacking him (hurry, they kill him fast). That can allow you to get vision as many times as you want during the attack and to check if the objects are destroyed. Moreover, if you want to see what happens when objects on which you have vision get destroyed, just make sure the dragons come to your dungeon. They will be like BigKnights in @Bertram25 's dungeon...

Bertram25:
> They will be like BigKnights in @Bertram25 's dungeon...

Eh eh. I'll test asap.

source/game/Seat.h
```diff
@@ -277,7 +277,10 @@ class Seat
     //! if the spawning conditions are not empty and are met, we will set it to true and force spawning of the related creature
     std::vector<std::pair<const CreatureDefinition*, bool> > mSpawnPool;
 
-    //! \brief List of the tiles this seat has vision on
+    //! \brief List of all the tiles in the gamemap (used for human players seats only). The first vector stores the X position.
+    //! The second vector stores the Y position. The first bool from the pair is set to true if the seat had vision on the concerned
+    //! tile during the last turn. The second bool is the same but for the current turn. That allows to notify only the tiles where
+    //! vision changed
```
Bertram25, line 283: Thanks for that. :)

source/network/ODServer.cpp
```diff
@@ -306,15 +306,17 @@ void ODServer::serverThread()
...
+                        for (int jj = 0; jj < gameMap->getMapSizeY(); ++jj)
                         {
-                            for (Seat* seat : gameMap->getSeats())
+                            for (int ii = 0; ii < gameMap->getMapSizeX(); ++ii)
                             {
                                 gameMap->getTile(ii,jj)->notifyVision(seat);
                             }
                         }
+
+                        seat->sendVisibleTiles();
```
Bertram25, line 322: Here, seems you fixed vision on the editor, right?

source/network/ODServer.cpp
```diff
@@ -606,6 +610,7 @@ bool ODServer::processClientNotifications(ODSocketClient* clientSocket)
             const std::string& nick = clientSocket->getPlayer()->getNick();
             int32_t id = clientSocket->getPlayer()->getId();
             int seatId = seat->getId();
+            seat->setMapSize(gameMap->getMapSizeX(), gameMap->getMapSizeY());
```
Bertram25, line 616: Why are you also doing this in ClientNotification::setNick:?

source/network/ODServer.cpp
hwoarangmy, line 322: yep ^^

source/network/ODServer.cpp
hwoarangmy, line 616: Because in the editor mode, we don't have the ClientNotification::seatConfigurationSet event. So, in editor mode, we do that in the ClientNotification::setNick event (ie last state before game mode).

Bertram25, line 616: Ok, then I'd propose to either add an if (== EditorMode) if it is feasible there or simply document it.
Both can be done later. :)

hwoarangmy, line 616: Well, this code is protected by a "if(mServerMode != ServerMode::ModeEditor) break;" 20 lines up. Is it not enough ?

Bertram25, line 616: Arf, nevermind. Short contextual diffs are sometimes evil for that sort of things.

akien-mga:
I'll let you merge if you're happy with the implementation @Bertram25

Bertram25:
Ok, I'll test it tonight. :)

hwoarangmy:
I've implemented trap logic ^^
@akien-mga @Bertram25 It is not too badly tested but I've not really well tested multiplayer. So, if you could also have a look...

Bertram25:
I couldn't test yesterday... If everything goes well, I'll be able to test before this week-end. :)

Bertram25:
Tested and working. The comment can be fixed later. :)

PR #417, opened 2015-01-04
Some more work on the console commands...

Bertram25:
I'll think I'll finish that in another PR, though. As I'll som work to do before coming to OD development.

Feel free to comment though. :)


Bertram25:
Ah, this PR doesn't remove the config/console_commands.txt file.
I'll do it when finishing that part. If you guys don't feel like merging half-finished code (even if it is stable and working AFAIK), it's fine and can wait. :)

Bertram25:
As I thought it won't merge anymore, I'll finish the work on console command management and rebase.

Bertram25:
@oyvindln @hwoarangmy I've been fiddling with this in order to finish what's planned in this PR earlier, and I wanted to do this:
- Rename ODConsoleCommand to ServerConsoleCommand as that's what they are, and do accordingly to the corresponding file.
- Rename the ConsoleCommands class into ConsoleCommandsHandler to make it less confusing.
- Factor the client only commands into functions in the ConsoleCommandsHandler.
- Finally drop the config/console_commands.txt file.

Are you ok with this?

hwoarangmy:
ok. But try to not have too long class names. They are already pretty long so if we could shorten that a bit, it won't be bad...

Bertram25:
> They are already pretty long so if we could shorten that a bit, it won't be bad...

I agree, in fact I already did name the ServerConsoleCommand children classes SCC\* but do you agree about the rest of the philosophy of the change?

hwoarangmy:
yep. That seems good

Bertram25:
Ok, cool. :)

oyvindln:
So you are basically doing what I did in my earlier PR.

Bertram25:
Well, at least we agree. The move to cegui will be left to be done I guess.

Bertram25:
I've finished the work on this. @oyvindln Could you have a look at whether you saw other improvements I may have missed?

In any case, the code looks saner now. :)

oyvindln:
My main idea was to also put the command functions in the map, and use that for lookup, rather than a big if chain.
Something like this:
https://github.com/oyvindln/OpenDungeons/blob/ceguiconsole_old/source/modes/ConsoleInterface.h
What do you think?

Bertram25:
Yeah. The idea/concept is good. We would thus add a function pointer in the commandHelp to be renamed map, basically, right? The fact to add an unordered map is also good, but as often, I think the shared ptr use isn't worth it. ;) (Well, you know me about this about such things...)

I won't be able to add much on that this Week-End so I'll request to first work on merging this one. Then, we can see together who does what to get your idea merged. Would you be ok with it?

oyvindln:
Yep. (Well, std::function rather than function pointer) Sounds good.
The use of shared_ptr was to have multiple names for one function, doing it with plain pointers would be rather complicated. It has nothing to do with the map being ordered/unordered. Don't know if we really need that though.

Bertram25:
Ah, the aliases. I get it.
AFAIK, the unordered map can be faster in certain cases but here we wouldn't see any difference.

oyvindln:
Yep, unordered has faster lookup, but that's really not important here anyhow.

hwoarangmy:
Tested and approved ^^

Bertram25:
Cool. :)

@oyvindln Want to bite the std::function implementation or will this be my punishment for stealing some job from you? ;)

oyvindln:
I will look in to it.

PR #423, opened 2015-01-07
Gui upgrade

Bertram25:
Started adding missing GUI icons, and upgrading certain GUI part that will be used soon, such as the checkbox.
This will permit to focus on the code and have the graphics ready.


akien-mga:
Tested, it looks fine :-)

PR #430, opened 2015-01-08
Several fixes/improvements

hwoarangmy:


Bertram25:
I tested. It is working rather fine.
The error string is a bit rough in the replay window., though.

I would rather display the replay version and then between parentheses (wrong version)
then skip two lines and then display the replay summary.

Could you do that?

hwoarangmy:
ok, I will do that. But we need to keep in mind that until the map description, we should not change which parameters are sent in the loadLevel message because doing what you request supposes backward compatibility (at least until reading mapdescription).

hwoarangmy:
that's done

Bertram25:
>  But we need to keep in mind that until the map description, we should not change which parameters are sent in the loadLevel message because doing what you request supposes backward compatibility (at least until reading mapdescription).

Sure, I wouldn't like to get in the way this. I'm just speaking of some less confusing text. :) From what I saw in the code, it seems pretty well doing btw. I'll test now.

hwoarangmy:
Well between OD version and map description, there is map size and map filename level. That doesn't sound too weird to keep that (even if level filename is not used anymore, it seems relevant to keep it).

Bertram25:
Ok as far as I've seen, let's merge :)

Issue #192, opened 2014-09-28
Add the portal and temple rooms as available rooms for the editor


Bertram25:
New icons are also needed.

Bertram25:
@Danimal696 @hwoarangmy @akien-mga please, tell me which ones to make :)
Icons proposal
- Portal: 
  http://game-icons.net/lorc/originals/orbital.html
  http://game-icons.net/lorc/originals/magic-portal.html
  http://game-icons.net/lorc/originals/reticule.html
- Dungeon Temple:
  http://game-icons.net/lorc/originals/caged-ball.html
  ![caged-ball](https://cloud.githubusercontent.com/assets/765917/4440300/6ad61ae0-47c3-11e4-8f7c-156a7db9901f.png)

What do you think?

akien-mga:
I like both the orbital and magic-portal icons for the Portal. The reticule one doesn't fit IMO. Maybe the magic-portal is best?

As for the Dungeon Temple, the proposal looks nice :-)

Bertram25:
Added the reworked dungeon temple icon. I'll wait for the others before doing more.
Ping @Danimal696 @hwoarangmy

Danimal696:
I like the magical portal icon, and the temple one fits for now

Bertram25:
Here it is, still hot from the newest foundry model: ;)
![magic-portal](https://cloud.githubusercontent.com/assets/765917/4441265/a28327dc-47d0-11e4-8248-fe536e77c24a.png)

hwoarangmy:
I also prefer the magic portal. The dungeon heart seems clear enough IMO.

PR #436, opened 2015-01-13
Added support to handle Temple/Portal in the editor + deharcoded workers.

Bertram25:
Part of #192

I also added support to handle moving the portal and temple only within the editor.
But this still triggers an assert failure when removing tiles including the one
with the mesh.
E.g.:
ERROR: Assert failed file opendungeons/source/network/ODClient.cpp line 622 info : name=RenderedMovableEntity_Portal2_50

I've made a few tests but so far I really can't tell what's triggering it.
Any advice? ;)

Otherwise, it' working fine.

### DIFF
--- a/source/utils/ConfigManager.h
+++ b/source/utils/ConfigManager.h
@@ -73,7 +73,11 @@ public:
 
     const std::vector<const SpawnCondition*>& getCreatureSpawnConditions(const CreatureDefinition* def) const;
 
+    //! \brief Get the fighter creature definition spawnable in portals according to the given faction.
     const std::vector<std::string>& getFactionSpawnPool(const std::string& faction) const;
+
+    //! \brief Get the worker creature definition spawnable in portals according to the given faction.
+    const std::string& getFactionWorkerClass(const std::string& faction) const;
 
     inline const std::vector<std::string>& getFactions() const
     { return mFactions; }
@@ -130,6 +134,10 @@ private:
     std::map<const CreatureDefinition*, std::vector<const SpawnCondition*> > mCreatureSpawnConditions;
     std::map<const std::string, std::vector<const CreatureMood*> > mCreatureMoodModifiers;
     std::map<const std::string, std::vector<std::string> > mFactionSpawnPool;
+
+    //! \brief Stores the faction default worker creature definition.
+    std::map<const std::string, std::string> mFactionDefaultWorkerClass;
+
     std::vector<std::string> mFactions;
     std::map<const std::string, std::string> mRoomsConfig;
     std::map<const std::string, std::string> mTrapsConfig;
--- a/source/utils/ConfigManager.h
+++ b/source/utils/ConfigManager.h
@@ -76,7 +76,7 @@ public:
     //! \brief Get the fighter creature definition spawnable in portals according to the given faction.
     const std::vector<std::string>& getFactionSpawnPool(const std::string& faction) const;
 
-    //! \brief Get the worker creature definition spawnable in portals according to the given faction.
+    //! \brief Get the default worker creature definition spawnable according to the given faction.
     const std::string& getFactionWorkerClass(const std::string& faction) const;
 
     inline const std::vector<std::string>& getFactions() const
--- a/config/factions.cfg
+++ b/config/factions.cfg
@@ -7,6 +7,9 @@
 [Faction]
     Name	Keeper
     [SpawnPool]
+    # Worker
+    Kobold
+    # Fighters
     CaveHornet
     Wyvern
     Dragon
@@ -26,6 +29,9 @@
 [Faction]
     Name	Hero
     [SpawnPool]
+    # Worker
+    DwarfWorker
+    # Fighters
     Gnome
     Knight
     BigKnight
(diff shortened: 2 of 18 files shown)


hwoarangmy:
Concerning the assert/segfault problem, if it only happens when removing tiles, I would guess that you are removing the dungeon temple/portal object twice and that's why it is not found the second time. Could you please post some log ?
If you want, I can have a look tonight.

hwoarangmy:
Since the FOW, as you have seen in the PR, the objects creation/remove has become a bit tricky. So, there might be some bug hidding. Moreover, I've not tried to manually remove objects like it is done here. That's why we can also imagine that there is some bug or incompatibility with the architecture.

Bertram25:
> I would guess that you are removing the dungeon temple/portal object twice and that's why it is not found the second time. Could you please post some log ?

Ok, I'll do it, I just wanted to know whether I did something plainly stupid. I do think it is a good way to practice with objects with the latest architecture, though. :)

> That's why we can also imagine that there is some bug or incompatibility with the architecture.

That's the thing I feared. Doing something incompatible with the current architecture. Also, here, I first wanted to only move the temple/portal and I changed the update*Position() accordingly to use a setPosition(there, isMoving = true); on the temple object but this is not working. That's why I changed the logic to delete/recreate the object each time.
The second problem I had is that CreateMeshLocal() is called each time before absorbRoom() if there is an absorption, so I had to let the mesh creation there, while I obtained a duplicate when the room is absorbed.
Maybe you have a better way to do it? (As you basically rewrote that system, eh. ;])

Best regards,

hwoarangmy:
Well, Building::updateActiveSpots was supposed to handle active spots. It is overriden in Room and it is the function that removes/creates the activespots when needed (by calling notifyActiveSpotCreated and notifyActiveSpotRemoved). Since the active spots behaviour for Portal/Dungeon temple is pretty specific, IMHO, the easiest would be to override Room::updateActiveSpots (which I forgot to set as virtual btw). IIRC, it is called only after room absorbtion and it works with the editor so it is probably well fitted for that ^^

hwoarangmy:
While I'm thinking about it, if you override Room::updateActiveSpots, you should remember that you will have extra calls when a neighboor wall is claimed. That means that you would need to check if the room object has moved before removing/adding it

Bertram25:
Thanks for the hints, I'll redo the patch. :)

Bertram25:
Should be working good now. :)

akien-mga:
It works fine here. Room merging seems to work great with all kind of room sizes, even though portals and temples are meant to be on 3x3 rooms.

Bertram25:
> It works fine here. Room merging seems to work great with all kind of room sizes, even though portals and temples are meant to be on 3x3 rooms.

Cool :)

> > Ideally, the mode ougt to be passed in as a parameter where it's needed.
> > That's what I've done when I had to but I have to say that updateActivespots is used in every trap/room so that will mean many changes...

So yes, I'd rather add a function in the gamemap class, since it's the true class making the most the separation between the two threads logic. What about:

``` c++
bool GameMap::isInEditorMode() const
{
    if (isServerGameMap())
    {
        return (ODServer::getServerMode == ServerMode::ModeEditor)
    }

    return (ODFrameListener::getSingleton().getModeManager->getCurrentModeTypeExceptConsole() == Modetype::EDITOR)
}
```

hwoarangmy:
> I'd rather add a function in the gamemap class, since it's the true class making the most the separation between the two threads logic

That seems ok to me

Bertram25:
Note to self:
- Add function to test whether one is in map editing mode in the game map object.
- Add support to define a default worker class in Seat, when initialization the spawn pool.

I'll do that asap. :)

hwoarangmy:
If I am not mistaken, you have done all you wanted in this PR, right ?
If it is the case, I will test tonight ^^

Bertram25:
Yep, normally I'm done. :)

hwoarangmy:
Seems like rebasing is needed ^^

Bertram25:
k, I do that asap. :)

Bertram25:
Rebased :)

hwoarangmy:
Nice. I will test tonight

source/utils/ConfigManager.h
```diff
@@ -130,6 +134,10 @@ class ConfigManager : public Ogre::Singleton<ConfigManager>
     std::map<const CreatureDefinition*, std::vector<const SpawnCondition*> > mCreatureSpawnConditions;
     std::map<const std::string, std::vector<const CreatureMood*> > mCreatureMoodModifiers;
     std::map<const std::string, std::vector<std::string> > mFactionSpawnPool;
+
+    //! \brief Stores the faction default worker creature definition.
+    std::map<const std::string, std::string> mFactionDefaultWorkerClass;
+
```
oyvindln, line 140: Not that it's going to make a huge difference, but the strings in the map probably shouldn't be const, as that may prevent move-construction if the map has to reallocate memory.

hwoarangmy, line 140: For maps? I thought it was for vectors only...

oyvindln, line 140: Hm, you are right about maps it seems.

hwoarangmy:
Works well, I merge

PR #441, opened 2015-01-16
More work on console

oyvindln:
Combined descriptions with the commands themselves, and use a map rather than a huge switch statement.

(Also includes a few other cleanups)

NOTE:
This sort of makes scrolling through command history buggy, but I don't feel like fixing that since the plan is to move to using a cegui window and drop the old console class anyway.


akien-mga:
>  This sort of makes scrolling through command history buggy, but I don't feel like fixing that since the plan is to move to using a cegui window and drop the old console class anyway.

I haven't tested your PR yet, but I can say the command history was already really buggy before, so this is not necessarily a regression.

Bertram25:
Ok, I've seen two functional regression.

When you type, say 'cre' and "tab", it will show you the two possible commands, which are alias.
There are then two issues:
the first is that the original command doesn't get completed cycling with the two proposal like before which pretty much defeats the autocompletion feature in that case since you have two aliases that prevent actual quick-autocompletion from one another.
And the two possibilities gets added to the commands history while they were never actually triggered, explaining why the history gets flawed.

oyvindln:
Well the first issue is really a question about what we want to happen when there are more than one matching functions to autocomplete. We could probably add a test that checks if all the matches are the same alias, but in other cases I think not completing makes more sense.

As for the history, I'm not quite sure how the old console class works, whether I broke something, or if it worked before. The functionality works in the new console class, I just didn't enable it, though I might be able to.

Bertram25:
> Well the first issue is really a question about what we want to happen when there are more than one matching functions to autocomplete. We could probably add a test that checks if all the matches are the same alias, but in other cases I think not completing makes more sense.

To make it more bash-like I guess? Well, yes, It can work also like that. In fact, I don't mind, I just wanted to point it out in case there were something flawed.

> As for the history, I'm not quite sure how the old console class works, whether I broke something, or if it worked before. The functionality works in the new console class, it's just not used yet.

I must say it's good question. I'll compile tomorrow and try to see whether the history used to work.
In any case, this can be fixed in another PR if you prefer merging a stable step.
We can even add a new issue not to forget about it in the meanwhile.

Well, I'll merge now as it is functional. I'll remove the using calls concerning the string and vector if you don't do it before me in another PR.

Congrats for the work done on this. :)

PR #442, opened 2015-01-17
Kept working on the GUI.

Bertram25:
- Added tab button hover images.
- Desaturated the tab buttons and tab panes to make them fit the rest of the widgets.
- Vertically centered the buttons for rooms, traps, tiles and creatures.
- Updated the tab content background.

The actual of the backgrounds is to show which faction the player is from.
Atm, only the keeper background is used.
The goal is also to add a bit of variety where the GUI can be the most unused
depending on the tab chosen.


Bertram25:
Feel free to comment if you see something unfitting. :)

akien-mga:
It looks good :-)
About the tab content background, currently it does not blend so well with the plain brown of the rest of the lower GUI. Do you have plans to improve this? Maybe by extending the background picture to fill the whole tab, and using parts of the artwork for the tab buttons too?

Bertram25:
Honestly, I wanted more to show you guys what is possible but I think I'll revert the background image for now. I'll do mockups until we all agree. Ok for you?

akien-mga:
Well we are in the development branch, so IMO it's okay to merge as is, and continue to build upon it. It's nice to see that we have such customisation possibilities with the UI, we'll go on from here and see what we come up with for 0.5.0 :-)

akien-mga:
Merging because the tab buttons improvements are great already. We can experiment with the GUI background image in your next PR :-)

PR #447, opened 2015-01-18
Fixed MinGW compilation issues

hwoarangmy:
Note to self: It would also be nice to replace the calls to Ogre::StringConverter::toString by Helper::toString

### DIFF
--- a/source/modes/ConsoleCommands.cpp
+++ b/source/modes/ConsoleCommands.cpp
@@ -58,13 +58,13 @@ Command::Result cSetFrameListenerVar(std::function<ValType(ODFrameListener&)> ge
 {
     if(args.size() < 2)
     {
-        c.print("Value " + name + " is: " + std::to_string(getter(fl)));
+        c.print("Value " + name + " is: " + Helper::toString(getter(fl)));
     }
     else
     {
         ValType v = Helper::stringToT<ValType>(args[1]);
         setter(fl, v);
-        c.print("Value " + name + " set to " + std::to_string(v));
+        c.print("Value " + name + " set to " + Helper::toString(v));
     }
     return Command::Result::SUCCESS;
 }
@@ -83,7 +83,7 @@ Command::Result cAmbientLight(const Command::ArgumentList_t& args, ConsoleInterf
     else if(args.size() >= 4)
     {
         // Set the new color.
-        Ogre::ColourValue v(std::stof(args[1]), std::stof(args[2]), std::stof(args[3]));
+        Ogre::ColourValue v(Helper::toFloat(args[1]), Helper::toFloat(args[2]), Helper::toFloat(args[3]));
 
         mSceneMgr->setAmbientLight(v);
         c.print("\nAmbient light set to:\n" +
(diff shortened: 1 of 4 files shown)


Bertram25:
Looks fine as far as I'm concerned. If another linuxian can confirm it is, let's merge. :)

akien-mga:
Builds and runs fine here, I merge.

oyvindln:
Is there any reason to reinvent the wheel instead of using  boost::lexical_cast or stringconverter directly? Also this means that console arguments that are mistyped become 0 instead of giving an error.

Bertram25:
> Is there any reason to reinvent the wheel instead of using boost::lexical_cast or stringconverter directly? Also this means that console arguments that are mistyped become 0 instead of giving an error.

I agree with you  @oyvindln . I see your point but the wheel will need to be cross-platform and supported on MingW.
Is there a way to achieve this while keeping readable code?

hwoarangmy:
> Is there any reason to reinvent the wheel instead of using boost::lexical_cast or stringconverter directly?

Well there was no use of boost::lexical_cast in any code I've read. Moreover, IIRC, boost::lexical_cast can throw exceptions when Helper::toDouble/toInt just return 0 when it fails. That's why changing to boost::lexical_cast  is not straightforward IMHO...

oyvindln:
The console code already relied on exceptions, the functions from the standard library worked that way. Other places returning 0 may make more sense.
But why not use ogre::stringconverter directly instead of adding a function in helper.h that calls it?

hwoarangmy:
> Other places returning 0 may make more sense.

Yes because AFAIK, most of the places where Helper::toDouble/toInt are called are during config file parsing (for example creatures.cfg). So using an exception thrower function can really change the current behaviour (even if I can understand that one do not want to use broken config files but that's just how it is currently done).

> But why not use ogre::stringconverter directly instead of adding a function in helper.h that calls it?

Well, the functions are inlined (so if I am not mistaken, the performances are equal). Moreover, using:
Ogre::StringConverter::toString() is longer than:
Helper::toString
And last, if we want to change, for example, to boost::lexical_cast, it will be easier to change in one place only than changing everywhere in the code.

That's why I used the Helper functions. But if most of us want to change, I won't bother.

Bertram25:
I'll add that the helper functions are also a way to make sure we use code approved to do certain common operations.
As for the exception handling. I'll daresay I hate them and prefer making programs that handle such cases and go on rather than triggering an exception for anything. Let's use them only when they matters. It is especially important in a game where the players want to keep playing.
What I agree about, on the other hand, is using boost libraries when they help us rather than coding stuff ourselves but that's a slightly different point.

hwoarangmy:
IMHO, there are 2 ways of handling unexpected states:
- Throw an exception
- Try to continue without segfault and see what happens

In the code there are loads of checks like testing nullptr that continues when there is an unexpected state. The choice has already been made for the second possibility. As a java coder, I'm more used for the first one but usually, when I start working on a code that is not mine, I try to adapt and to follow its philosophy ^^

PR #448, opened 2015-01-20
Fixed normal maps on coloured tiles + some small cleanups

oyvindln:
I did it by adding an extra pass at the end that adds a texture on top, using scene_add so only the white parts are added, and altering the colours dynamically.
It mostly works, though on the small singleplayer test map sometimes something happens that messes up the materials after a while.

Materials still have to be tuned and possibly fixed.


oyvindln:
This also fixes the console history.

akien-mga:
Well done, it's a nice improvement! The game gets a whole new looks.
I notice a big FPS drop on Greed or Might, and I think it's related to the lighting sources that I've added in the relevant points, they seem to use up lots of resources (with my Intel HD 4000 GPU).

Actually I also get down to 15 fps on Greed or Might with my Nvidia GTX 670MX GPU, so I think there's something to optimise somewhere.

With the nvidia driver, I also get a funny lighting:
od-red

And the green ally in Scream in the dark:
od-green

akien-mga:
I confirm that the console history is fixed. I'd just have one request to make it more bash-like: after you've browsed up in the history and run a command, I think the newly issued command should be added to the bottom of the history, and your history "position" should be one off the bottom.

It might be clearer with an example. Currently, with the following history:

```
help
addgold 1 1000
ls classes
hey guys
```

If you go to `addgold 1 1000` in the history and run it again, it will take effect, but then pressing up will bring `help` and pressing down will bring `ls classes`.
The bash history would do it differently: after running `addgold 1 1000`, the new history would be:

```
help
addgold 1 1000
ls classes
hey guys
addgold 1 1000
```

and pressing up would bring the last `addgold 1 1000` again.

hwoarangmy:
This seems really interesting. I will try tonight ^^

hwoarangmy:
as expected, std::this_thread do not compile with MinGW. Would you mind rolling back sleep ?

EDIT: I've tested after rolling back sleep and I see the same as @akien-mga (my graphic card is nvidia geforce gtx 560M)

oyvindln:
Seeing same as in shaded tiles? And is it all tiles?

Also, aren't there any more up to date versions of MinGW you can use?

oyvindln:
I also get 15 fps on greed or might, on my 9 year old computer with a Quadro FX 3500M card (Roughly equivalient to GeForce Go 7950GTX). I would have thought you would be getting better performance with newer computers. EDIT: Maybe because I run the game in 800x600.
 Though there are probably a lot of improvements we could do to the graphics,

hwoarangmy:
@oyvindln sleep works well with boost, thanks.

Concerning the display, a screenshot is better than words ^^
odscreenshot_2015-01-20_215455_0

oyvindln:
Hm, don't know what could cause the difference. Doesn't look like you have the normal maps either. Here it looks like this:
odscreenshot_2015-01-20_221802_2

Granted, I do think it looks better with the whole tile shaded, than than how it looked before with non-working lighting.

akien-mga:
It seems to be a driver-specific issue, since I have no issue with the open source intel driver, but I get @hwoarangmy's colouring with my nvidia GPU and its proprietary driver.

hwoarangmy:
I will try to update my drivers and we will see. But in any case, I'm not sure what we should do about this. We cannot expect players to have the most up to date drivers and nvidia is pretty popular amongst players...
I will also try tonight to remove the colouring pass of rendering to check what happens (on sources with and without this PR).

akien-mga:
@hwoarangmy I have the most up-to-date drivers for the Nvidia card (340.65 on Linux, actually it's no longer the latest long-lived branch, but still pretty recent), so it's not an update problem IMO.

By driver-specific I mean that we need to find an implementation that behaves the same regardless of the hardware. The bug is not in the driver, but in our usage of OpenGL I suppose. It's a bit the same as having code that works with gcc but not with MinGW, we need to have cross-platform code, and cross-hardware :-)

hwoarangmy:
> we need to have cross-platform code, and cross-hardware :-)

Yep, that's exactly what I think ^^

Bertram25:
> Granted, I do think it looks better with the whole tile shaded, than than how it looked before with non-working lighting.

@oyvindln If I'm not mistaken, we are several preferring to see only the cross in the middle affected by coloring. So please, stick to the idea. Apart from that, I'm glad of the work you did on this one. :)

Btw, I've seen you adapted the claimed wall texture (I'm not speaking the mask) but didn't adapt the claimed ground texture alike. Is that desired or something forgotten maybe?
https://github.com/oyvindln/OpenDungeons/blob/3b56c58f18827448ccc59eac17676bb12f7b3521/materials/textures/Claimedwall2B.png

oyvindln:
I changed the claimed wall texture as the alphaed part was all white, while on the ground tile, the alphaed part was all black so the color was different. 

> @oyvindln If I'm not mistaken, we are several preferring to see only the cross in the middle affected by coloring. So please, stick to the idea.

That was what I was trying, but if that's not possible, I would rather have shaded tiles, than tiles with the cross in the middle that not affected by lighting at all, as they were before this change. (At least on my system).

Bertram25:
> I changed the claimed wall texture as the alphaed part was all white, while on the ground tile, the alphaed part was all black so the color was different. 

k. Make sense. But then, it seems you removed the alphaed part on the wall mask. So I'm not sure I get the whole point yet. I'll have a look anyway.

> And, do you really think having tiles with non-working lighting look better?

No, don't worry. That's not what I meant. What I mean is that we'll have to find a way to make the cross colored of the team color without coloring the rest of the tile with it, while still lighting correctly the rest of the tile. This goes for walls and ground tiles.
Adding some emissive light from the central part may also be done later, I guess.

EDIT: To sum up, the screenshot you have here seems good to me. The next goal will be to make this work for everyone:
tiles ok

Danimal696:
once this particular pull is fully resolved and accepted, i would like a release, so i can start making normal maps for everything bit by bit

hwoarangmy:
FYI, I've hacked colourizeMaterial to not colorize anything to see if how tiles behave when not colourized. That seems ok. That might mean that our way to colourize is not what it should be ?

odscreenshot_2015-01-21_221440_0

oyvindln:
I don't see your point, if you disable the colouring, of course they aren't going to be coloured.
Also, it doesn't look ok to me, normal maps aren't working.

hwoarangmy:
> Also, it doesn't look ok to me, normal maps aren't working.

I'm not very used to normal maps. What would you expect ?

oyvindln:
> I'm not very used to normal maps. What would you expect ?

Look at the dirt tiles in my screenshot.

oyvindln:
Maybe it would be easier to just do colouring by using a different method instead.

Bertram25:
> I don't see your point, if you disable the colouring, of course they aren't going to be coloured.

My point is that not all the texture should be colored (of the team color), only the alphaed part.

> Also, it doesn't look ok to me, normal maps aren't working.

On your screenshot, I must say it's hard to know whether the normals are applied on the colored claimed tiles. I sincerely thought they were. That said, we agree the normal should apply, of course.

> Maybe it would be easier to just do colouring by using a different method instead.

Yeah, we might have to specialize that, indeed. Still need to get a way to make this work for every hardware first.

oyvindln:
> My point is that not all the texture should be colored (of the team color), only the alphaed part.

That wasn't what I was talking about. What I meant was that I didn't see what the point was of showing that disabling colouring the tile was disabling colouring the tile. 

> On your screenshot, I must say it's hard to know whether the normals are applied on the colored claimed tiles. I sincerely thought they were. That said, we agree the normal should apply, of course.
> It's hard to see on the claimed tiles, but you can clearly see the normal mapping on the dirt tiles on my screenshot, while they look flat on akien and hwoarangmy's screenshots, which means that the shaders are not working on those systems and something is broken.

The rendering pass that is doing colouring uses scene_blend add to mask areas that are not supposed to be coloured in the pass. The mask texture is black in the masked areas, which means the pass should add 0 to the existing pixel value, and thereby not do anything. However, on your systems, it seems that it adds light on the whole tile anyway. Maybe the fact that the shaders are not working are causing the technique that is used to render be some fallback default, which would mean that the colouring ends up applying to a normal lighting pass instead.

Are you getting any errors or warnings in the log about shaders or materials?

Bertram25:
> What I meant was that I didn't see what the point was of showing that disabling colouring the tile was disabling colouring the tile. 

Eh eh. I guess let's move on.

> Are you getting any errors or warnings in the log about shaders or materials?

I'll take the time to check thoroughly asap.

oyvindln:
I suppose we may want to consider dropping using the RTShader system. Even on my system, the lighting is buggy and breaks randomly, so maybe it will be easier to use the shaders from the ogre samples instead. It doesn't look too complicated, and it would make it easy to add in a custom shader pass that adds colour to an object.

Bertram25:
> I suppose we may want to consider dropping using the RTShader system. 

I was sure we would have to move using custom shaders one day or another, so I'm not surprised by your conclusions, even if, unfortunately, it means more work on shaders. Long-term it will be better though, IMHO.
Just in case, I know 0ad guys have been having problems with gl shaders and intel drivers and had to use glsl shaders as a workaround to properly render stuff for a few random glitches, like: http://trac.wildfiregames.com/ticket/3012#comment:1
(just in case it might help you.)

First thing first, do you think we should have one last attempt as trying glsl shaders here?
Or should we straight to simple pass/multipass samples shaders?

oyvindln:
I think you are misunderstanding something. We are using GLSL shaders in either case. The question was if we should drop trying to use the RTShader system thing that generates shaders automatically. I tried using the example normal map shaders from the ogre samples, which turned out not being supported by my card (my computer is old, so only DX9 support), which is exactly the issue of doing it manually. Though if RTShader is broken, we may not have much choice, though, while we wait for Ogre 2.0, there is [shiny](https://github.com/scrawl/shiny) which is supposed to do something similar. (Granted, it looks like Ogre 2.0 may drop support for DX9 as well...)
Though, it might be that RTShader isn't generating the right shaders on your cards that is the issue, the RTShader example has folders for two GLSL versions, and we are only including the older one right now.

Bertram25:
> I think you are misunderstanding something. We are using GLSL shaders in either case.

OMG, you're right. I simple glance to the materials files sufficed. Sorry, I somehow mixed this with something else. Needs some sleep. ;)

> Though, it might be that RTShader isn't generating the right shaders on your cards that is the issue, the RTShader example has folders for two GLSL versions, and we are only including the older one right now.

Yeah, this sounds sensible to me, while I know the GLSL150 shaders won't be used by the RTShaderSystem for my intel card. (I already tried that).
Also, why are you speaking about DX9 since in a way, we'd have to stick with OpenGL to stay cross-platform? Have you tried direct GL shaders use also, for instance?

oyvindln:
By DX9 I'm referring to what the card supports, the OpenGL counterpart would be OpenGL 2.1.

Bertram25:
> By DX9 I'm referring to what the card supports, the OpenGL counterpart would be OpenGL 2.1.

Ok. From what I could see, we can't really aim at a higher norm anyway - or maybe 3.1 using mesa?, as mesa itself doesn't give more than the 3.3 norm, and only if requested at context creation and if the driver is supporting it.
Plus, it seems Cegui is still having a few glitches with the opengl 3 backend which are being fixed.

Another question: What is your old graphic card? (Do you think it would be possible to use mesa?)

The best IMHO is that we start a new branch and make cumulative patches on it until it reaches a satisfying state, so that we can focus on the game logic in the development branch, but can work on shaders in the new branch. I'd name it new_shaders or something like that.
WDYT guys?

Also, could you tell me precisely how you did test the Ogre shader sample, and which one?
It would be cool that we start working on shaders and make them compliant with our different hardware.
For instance, I know the mesa shader IR compiler is sensitive to newlines, for one thing, thanks to @authenticate's work on them in VT.

Bertram25:
Btw, about this very PR, I think we should merge it only in the new branch, and backport the commits not linked to the materials into development. Do you agree?

oyvindln:
> Plus, it seems Cegui is still having a few glitches with the opengl 3 backend which are being fixed.

That's irrelevant since we are using the Ogre backend.

> Another question: What is your old graphic card? (Do you think it would be possible to use mesa?)

I'm using a Quadro FX 3500M card (Roughly equivalient to GeForce Go 7950GTX). I could use noveau instead, but that won't help since it lacks hardware support for DX9/OpenGL 3 and newer features, like the newer shader model. So I don't know if I will even be able to use Ogre 2.0 when it comes out. Having to deal with different cards and drivers supporting different features is one of the issues of doing the shaders in a manual way, which is why using say the RTShader thing, or ogre shiny, or in Ogre 2.0 the high-level material system is preferable if it's possible. 

> Btw, about this very PR, I think we should merge it only in the new branch, and backport the commits not linked to the materials into development. Do you agree?

That sounds like it would be complicated and difficult to maintain. Maybe we should just go back to not bothering with normal maps for the time being. I find it very grating to have to work on these kinds of things instead of actually improving the gameplay. If we are going to do some more major graphical work, I think we need to do more to separate rendering and gameplay logic first at least, otherwise keeping this in sync will be very frustrating.

One big issue here is that we want to be able to have more than 1-2 lights we need to do something about the lighting model in the game, which will be complicated, and also probably involve shaders. If we are going to do heavy work on graphics, I think we need to sort that first, before worrying about effects like normal maps.

Maybe it would be best that I change the code in this PR to go back to not using shaders for now, so we can merge this and the other changes in it. Though, first, I would appreciate if @hwoarangmy or @akien-mga could see if they get any errors in the log from this.

> Also, could you tell me precisely how you did test the Ogre shader sample, and which one?
> It would be cool that we start working on shaders and make them compliant with our different hardware.
> For instance, I know the mesa shader IR compiler is sensitive to newlines, for one thing, thanks to @authenticate's work on them in VT.

I tried the bump map shaders from the ogre examples. There are material files showing how it's set up. It doesn't require any hardcoding.

Bertram25:
> which is why using say the RTShader thing, or ogre shiny, or in Ogre 2.0 the high-level material system is preferable if it's possible. 
> Though, first, I would appreciate if @hwoarangmy or @akien-mga could see if they get any errors in the log from this.

Well, if we can debug the shaders potential errors/issues, it would permit to go forward without dropping the RTSS of course. I'll have my own debug session as I'm using Mesa 10.4.2 with an intel GPU myself. I wish also we can somehow use the HLMS in Ogre 2.0 as it would be a time savior on the long term, IMHO. But this is all speculation for now. :S 

> That sounds like it would be complicated and difficult to maintain. Maybe we should just go back to not bothering with normal maps for the time being. I find it very grating to have to work on these kinds of things instead of actually improving the gameplay. If we are going to do some more major graphical work, I think we need to do more to separate rendering and gameplay logic first at least, otherwise keeping this in sync will be very frustrating.

While this is true, remember you were the one starting an effort here, that's why I tried to find a solution without breaking your effort. So yes, I do agree, of course.
As for the syncing effort, It's usually me rebasing mostly, anyway, right? ;) Anyway, it's up to you atm concerning this PR.

> One big issue here is that we want to be able to have more than 1-2 lights we need to do something about the lighting model in the game, which will be complicated, and also probably involve shaders. If we are going to do heavy work on graphics, I think we need to sort that first, before worrying about effects like normal maps.

If you're actually speaking of how lighting applies to the whole in game world first (and any models in general), then indeed there is already a problem there.
For instance, even ambient fog and such things aren't working right. And maybe the octree renderer isn't the best choice?
Could you elaborate on the more urgent problems you've seen and the solutions you're thinking about for that? Maybe we can start making clearer priorities out of it. :)

> I tried the bump map shaders from the ogre examples. There are material files showing how it's set up. It doesn't require any hardcoding.

k.

Bertram25:
> That's irrelevant since we are using the Ogre backend.

Ah, yes, but even using the Ogre backend for textures and such, cegui does use direct GL calls AFAIK, so IMHO, this is not that irrelevant unless there has been changes I'm not aware of, yet.
Keep feeling free to tell whenever I'm wrong of course. :)

hwoarangmy:
> Though, first, I would appreciate if @hwoarangmy or @akien-mga could see if they get any errors in the log from this

As you've already seen, I'm not very aware about how to use Ogre. If you want, I can do some tries if you have ideas about how to fix something (just tell me what I should change in code/materials). Or I can provide you logs if it is more relevant.
Could you just tell me what you want me to search exactly ? If you want, I can paste the log file so you can have a look. In any case, I think I still have the sources before your PR got unmergeable so just tell me ^^

oyvindln:
@hwoarangmy Start up a game, load a level, and quit. Then make a copy of the log/output and post it here. I just want to see if ogre outputs any errors indicating that the shaders are not working.

> While this is true, remember you were the one starting an effort here, that's why I tried to find a solution without breaking your effort. So yes, I do agree, of course.
> As for the syncing effort, It's usually me rebasing mostly, anyway, right? ;) Anyway, it's up to you atm concerning this PR.

Well I want to get the other fixes merged anyhow, so we will have a go and see if we can find if there is a easy fix here, otherwise I'll drop the normal maps for now.

> Could you elaborate on the more urgent problems you've seen and the solutions you're thinking about for that? Maybe we can start making clearer priorities out of it. :)

Yes, though maybe we should discuss that in a separate issue, or on the forums, rather than in this PR.

Bertram25:
> Well I want to get the other fixes merged anyhow, so we will have a go and see if we can find if there is a easy fix here, otherwise I'll drop the normal maps for now.

k.

> Yes, though maybe we should discuss that in a separate issue, or on the forums, rather than in this PR.

Sure. :)

akien-mga:
Well it's true that there are errors in the logs, I should have had a look at it earlier:

```
17:19:49: GLSL compiling: 8d793fa8-3e93-5699-5228-6e26e7548b47_VS
17:19:49: GLSL compiled: 
17:19:49: Vertex Program:8d793fa8-3e93-5699-5228-6e26e7548b47_VS Fragment Program:8a95c894-c9e5-a4df-6b62-0b635cb83b1c_FS, Error linking program: Vertex info
-----------
0(198) : error C7616: global function texture1D is removed after version 140
0(229) : error C7616: global function texture3D is removed after version 140
0(237) : error C7531: global function textureCube requires "#extension GL_NV_shadow_samplers_cube : enable" before use

Fragment info
-------------
0(198) : error C7616: global function texture1D is removed after version 140
0(229) : error C7616: global function texture3D is removed after version 140
0(237) : error C7531: global function textureCube requires "#extension GL_NV_shadow_samplers_cube : enable" before use
```

Here is a complete log from launching OD with my Nvidia GPU and its proprietary driver (340.65):
http://paste.fedoraproject.org/173639/22030190

hwoarangmy:
Here is my log:
http://pastebin.com/FjBz4St6

Bertram25:
Just tested, no error on my side. The GLSL shaders compile just fine (unfortunately since I wanted some debug output).
The normals are just applied fine, btw. Coloring is also working while being more discrete now.

Bertram25:
Ah, for mesa users, does this provide more info?

`env MESA_DEBUG=1 ./opendungeons` from the repo root folder.

Bertram25:
For @hwoarangmy 's error: https://ogre3d.atlassian.net/browse/OGRE-414
There seems to be a patch there. I don't know whether it could fix the issue. Can you try porting it?
EDIT: Seems @tomluchowski did already complained about it: http://www.gamedev.net/topic/651751-solved-help-with-glsl-scripts/

hwoarangmy:
> Can you try porting it?

I'm trying

Bertram25:
As for @akien-mga 's error, it seems the Texture*D were removed in glsl 140. They have to changed by vectors AFAIK and this also speaks about it: https://www.opengl.org/discussion_boards/showthread.php/176016-Moving-to-GL_TEXTURE_RECTANGLE

The second error should be fixed by this, hopefully?
https://www.opengl.org/discussion_boards/showthread.php/184216-How-to-deal-with-error-c7531
Still have to check in the log if you have the extension.

Bertram25:
Btw, @hwoarangmy your error is the same kind than the first one for @akien-mga so this patch should fix both your errors, the extension problem remaining.

oyvindln:
Then I suspect we should readd the GLSL150 folder for the RTShader system, and select old or new depending on card capabilities, as the issue seems that it's trying to use functionality that is removed.

Bertram25:
> Then I suspect we should readd the GLSL150 folder for the RTShader system, and select old or new depending on card capabilities, as the issue seems that it's trying to use functionality that is removed.

Agreed. This can't hurt after all.
EDIT: Mind doing it?

Bertram25:
@akien-mga hasn't got the NV extension listed in the log. So hopefully switching to GLSL150 will help. If not, we'll see how to workaround it.

hwoarangmy:
It seems to work with the patch in @Bertram25 link. I've tried @oyvindln sources with only the changes to the glsl file.

odscreenshot_2015-01-23_221825_0

Moreover, I don't get the errors anymore in the logs

Bertram25:
Cool :)

akien-mga:
Same as @hwoarangmy, with the patch from the upstream bug report I don't have the issue anymore with my Nvidia GPU. No regression with the Intel GPU either :-)

oyvindln:
Okay, I added back the shaders and updated the code to select, like how it's done in the samples. So please test.

hwoarangmy:
@oyvindln currently compiling ^^

hwoarangmy:
I have the same result as before (wrong tile coloration) :-(
Moreover, I get same errors:

```
00:12:42: Vertex Program:935172cd-48c7-7e43-05a9-4893aab7c8f4_VS Fragment Program:8a95c894-c9e5-a4df-57dc-92f4b43cd459_FS, Error linking program: Vertex info
-----------
0(198) : error C7533: global function texture1D is deprecated after version 120
0(229) : error C7533: global function texture3D is deprecated after version 120
0(237) : error C7533: global function textureCube is deprecated after version 120

Fragment info
-------------
0(198) : error C7533: global function texture1D is deprecated after version 120
0(229) : error C7533: global function texture3D is deprecated after version 120
0(237) : error C7533: global function textureCube is deprecated after version 120

00:12:42: GLSL compiling: 4e752891-6e7b-990b-9f01-f41ba0242ccc_VS
00:12:42: GLSL compiled: 
00:12:42: GLSL compiling: de3f8655-4e71-22f2-f828-15494ae4edee_FS
00:12:42: GLSL compiled: 
00:12:42: Vertex Program:4e752891-6e7b-990b-9f01-f41ba0242ccc_VS Fragment Program:de3f8655-4e71-22f2-f828-15494ae4edee_FS, Error linking program: Vertex info
-----------
0(198) : error C7533: global function texture1D is deprecated after version 120
0(229) : error C7533: global function texture3D is deprecated after version 120
0(237) : error C7533: global function textureCube is deprecated after version 120

Fragment info
-------------
0(198) : error C7533: global function texture1D is deprecated after version 120
0(229) : error C7533: global function texture3D is deprecated after version 120
0(237) : error C7533: global function textureCube is deprecated after version 120
```

hwoarangmy:
@oyvindln FYI, I've had a look and it seems my graphic card do not support glsl150. The test gpuProgramManager.isSyntaxSupported("glsl150") fails :-(

oyvindln:
Do you get the same result if you try the ogre sample browser?

Bertram25:
No regressions at least on my part.

Bertram25:
About the marked coloring of tiles (yellow). It is fine for gold and dirt, but on claimed walls, the upper part isn't colored:
coloring-claimed-walls

Bertram25:
@hwoarangmy And if you force glsl150 by adding the glsl150 folder only in the resource file?

oyvindln:
@Bertram25  Hm, the whole tile gets coloured here. What map is it?

hwoarangmy:
> Do you get the same result if you try the ogre sample browser?

Yes. I get the same result as the screenshot in @Bertram25 link (wrong normals)

> @hwoarangmy And if you force glsl150 by adding the glsl150 folder only in the resource file?

I've tried to remove the test for glsl150 in ODApplication to make sure it is used but OD gets frozen (no error in the logs but nothing happens)

oyvindln:
The patch breaks the shader on my system, as the functionality is too new for my card. Maybe we will have to add another check then.

hwoarangmy:
I was going to ask if you could test with the patch but seems like it is not the way to go :-(

oyvindln:
If the patch work on your systems, I the solution would be to add a check that uses the patched version on glsl130 or newer.

hwoarangmy:
Seems like the problem was known (at least by @tomluchowski) :
http://www.gamedev.net/topic/651751-solved-help-with-glsl-scripts/

@oyvindln After digging a bit in the driver, here is the list of supported capabilities of my graphic card (there is no glsl130):
arbfp1
arbvp1
fp20
fp30
fp40
glsl
glsl100
glsl110
glsl120
gp4fp
gp4gp
gp4vp
gpu_fp
gpu_gp
gpu_vp
nvgp4
vp30
vp40

Do you want me to try something ?

Bertram25:
> Seems like the problem was known (at least by @tomluchowski) :
> http://www.gamedev.net/topic/651751-solved-help-with-glsl-scripts/

Yeah, put the same link above. ;)

> If the patch work on your systems, I the solution would be to add a check that uses the patched version on glsl130 or newer.

True, we should do that IMHO. The patch looks actually fine even on older cards, like mine.
What would then be the check for cards like yours?

> @Bertram25 Hm, the whole tile gets coloured here. What map is it?
> Yes. I get the same result as the screenshot in @Bertram25 link (wrong normals)

Ah, I didn't check the logs. I'll have a look.

hwoarangmy:
For those who want to list the capabilities (so we can find how to choose between old cards and newer ones), in ODApplication (after gpuProgramManager is created):

```
for(const std::string& str : gpuProgramManager.getSupportedSyntax())
{
    LogManager::getSingleton().logMessage("SupportedSyntax=" + str);
}
```

Bertram25:
Aren't they listed in the log near the top at startup? (or maybe I'm mistaken?)

oyvindln:
I don't know if we can use that function, it doesn't seem to output the glsl version supported properly, instead it outputs vendor-specific profiles. May have to test opengl version instead.

hwoarangmy:
> Aren't they listed in the log near the top at startup? (or maybe I'm mistaken?)

You are right, I had not seen

Bertram25:
> You are right, I had not seen

Ok, good to know it's the actual same thing.

hwoarangmy:
@oyvindln @akien-mga  Could you please tell us which glsl versions does your card support ?

If we are lucky, @oyvindln will not support glsl120 and we might be able to use the original or the patched version depending on that...

oyvindln:
Updated to use the patched shader. Though, given that both of you according to the logs have opengl 4 compatible cards, it seems that testing for the "glsl150" string may not work properly, as opengl 4 means the shader version should be at least 4. Are you able to test if the GLSL150 shaders work by editing the code?

hwoarangmy:
I get Shader version is: 420
It works well for me. But the game seems to become pretty laggy. I don't know how much did the FPS dropped down but it feels pretty much.

> Are you able to test if the GLSL150 shaders work by editing the code

As I said, when I remove the test to force using glsl150, OD freezes at launch (black screen). I get no error output.

oyvindln:
Are you sure you are not including both folders when you remove the test?

hwoarangmy:
BTW, I confirm @Bertram25 bug on marked claimed walls. I have exactly the same display (top is not yellow while other parts seem to be).

Concerning glsl150, I think I had a problem with my IDEbecause this time, it worked. When I use it, screen is pretty strange (see screenshot), normals do not work and I get the following error:
0(49) : error C3002: call to undefined function "void SGX_FetchNormal(sampler2D, vec2, vec4);"
0(49) : error C3002: call to undefined function "void SGX_Light_Point_DiffuseSpecular(vec4, vec3, vec3, vec4, vec3, vec3, float, vec3, vec3, vec3, vec3);"

odscreenshot_2015-01-24_023724_0

oyvindln:
That's weird. Maybe those shaders are just broken then.
EDIT: Unless there is some cache-mess up from using the other shaders.

akien-mga:
> Could you please tell us which glsl versions does your card support ?

If I'm to believe the opendungeons logs:

> Supported Shader Profiles: arbfp1 arbvp1 fp20 fp30 fp40 glsl gp4fp gp4gp gp4vp gpu_fp gpu_gp gpu_vp nvgp4 vp30 vp40

But as @oyvindln pointed out, this might be vendor-specific names (I don't know much about glsl versioning though).
My card supports OpenGL 4.4:

> GL_VERSION = 4.4.0 NVIDIA 340.65

akien-mga:
The current PR is perfect for me with both my intel and nvidia hardware.
I already gave the specifics of my nvidia hardware, here it is for intel:

> GL_VERSION = 3.0 Mesa 10.4.2
> Device Name: Mesa DRI Intel(R) Ivybridge Mobile 
> Supported Shader Profiles: arbfp1 arbvp1 glsl

hwoarangmy:
@akien-mga Could you please try to force glsl 150 to check if you have the same behaviour as me ?
Moreover, when you say the PR works, how do claimed walls look like when you mark them ?

akien-mga:
Claimed walls look fine, at least in the configuration in this screenshot:
od-marked

I'll try forcing glsl150.

akien-mga:
With GLSL150 and Nvidia, I get some graphical glitches such as @hwoarangmy's, and the overall looks is different from what I get with GLSL:
od-glsl150

With my Intel chipset however, I get the same as @hwoarangmy:
od-glsl150-intel

Bertram25:
I'll test  asap myself the PR as is on linux with intel/mesa drivers.
As for glsl150, do we really have to use it in the end?

oyvindln:
Well if it doesn't work, it can probably be removed.

hwoarangmy:
If I am not mistaken, the current PR with @oyvindln latest changes works for everybody here. But the glsl 150 only works for @oyvindln so far ( @Bertram25 didn't test yet). So, I guess it can be dropped as it doesn't seem very stable. Concerning the claimed walls not being coloured on the top, here is a screenshot showing marked claimed walls + dirt. Maybe @oyvindln will see something ^^
odscreenshot_2015-01-25_090110_0

Bertram25:
I've just tested on linux. I don't have the claim wall upper-part colouring problem anymore. (Mesa 10.4.2)
So I guess only testing on windows and fixing there is left.

oyvindln:
> But the glsl 150 only works for @oyvindln so far

No, it doesn't work here. My graphics card is too old.

Danimal696:
this sounds like a jerk, bit is it worthy to drop a functionality because some people still have very old graphic cards? It something to disscuse ourserlves

akien-mga:
> Is it worthy to drop a functionality because some people still have very old graphic cards?

While you are right generally speaking, in this case the glsl150 shaders seems not to work for anyone, and my nvidia card is pretty recent and capable (at least for a laptop), so to me it seems that the shaders are flawed somehow.

hwoarangmy:
> this sounds like a jerk, bit is it worthy to drop a functionality because some people still have very old graphic cards?

AFAIK, nobody could successfully try glsl150 (we don't know if it works for anyone and not even if the test done to check if glsl150 is available works). Not sure if we should keep an untested feature...

oyvindln:
The test checking if glsl150 exists in the supported shader models doesn't seem to work as intended, otherwise it would trigger for @hwoarangmy and @akien-mga given that they have systems supporting glsl 4.x. Checking shader version with getNativeShadingLanguageVersion() does seem to work, but unless @hwoarangmy and @akien-mga are doing something very wrong, the shaders that are supposed to be for version 1.50 and newer don't seem to work properly.

Bertram25:
> The test checking if glsl150 exists in the supported shader models doesn't seem to work as intended, otherwise it would trigger for @hwoarangmy and @akien-mga given that they have systems supporting glsl 4.x.

As reported here, adding a versioned profile seems missing: https://ogre3d.atlassian.net/browse/OGRE-298
See the `rsc->addShaderProfile("glsl150");` part.
Also, many tell that selecting the GL3+ renderer is needed before that.

Then, adding the glsl150 profile should probably be made only for capable hardware and in conjunction with the gl3+ renderer, then, I guess.
Do you think I may be worth a try?

Btw: I realize now we're almost there. :D Once this PR is merged, we'll have normals!! Yeah!

oyvindln:
Does the GL3+ renderer work for you? (I can't test it since my computer is too old.)

Bertram25:
I can't test either. My hardware is also not enough capable.

akien-mga:
Can you give me pointers on how to test it?

hwoarangmy:
I will test it tonight. I just need to compile it (IIRC, I don't compile this renderer when I compiled ogre).

oyvindln:
> Can you give me pointers on how to test it?

If the renderer is compiled, it should be possible to select a different renderer in the drop-down list.

akien-mga:
>  If the renderer is compiled, it should be possible to select a different renderer in the drop-down list.

Ah of course! :-)
It seems I did not enable the OpenGL3 renderer when I packaged OGRE 1.9.0, but I can fix that and give it a try.

akien-mga:
Just a hint for those who would try to build the OpenGL3+ renderer (from the original OGRE 1.9.0, not the v1-9 branch), you need to backport this patch to get it to build against mesa: https://bitbucket.org/sinbad/ogre/pull-request/431/make-gl3plus-rs-work-with-mesa-drivers/diff

hwoarangmy:
I've tried the GL3+ renderer.
When I use it, I get:
Supported Shader Profiles: glsl glsl130 glsl140 glsl150 glsl330 glsl400 glsl410 glsl420
I can see RTShaderLib/GLSL150 is used but I get a black screen (I can see in the logs that the game is launched but the screen stays black).
I've also tried to force using materials/RTShaderLib/GLSL/130 with the GL3+ renderer and the screen also stays black.

Bertram25:
It seems glsl150 is the new black. ;)
More seriously, I don't know if @akien-mga has tested the gl3+ renderer plugin already, but I'm less and less convinced we should keep those glsl150 shaders at all.
Let's maybe try again once when 1.10 is out, and anyway, this can wait for 2.0 to happen, IMHO.

WDYT?

What's left to solve is the upper part of marked claimed walls for @hwoarangmy with the latest additions, right?
EDIT: Ah, I would make the yellow colouring a bit stronger, as it is quite discrete now.

@oyvindln Once this PR is merged, do you think we could make the claimed cross color a bit emissive, to obtain to a more glowish effect?

akien-mga:
Here how the game looks like with the OpenGL 3+ renderer:
- Nvidia (supports OpenGL 4.4):
  od-opengl3
- Intel/Mesa (OpenGL 3.0):
  od-opengl3 _2

So basically the game is overall too dark, and there are some graphical glitches on some tiles. However, the claimed tiles lighting works really fine.

Bertram25:
@akien-mga Thanks for the test. :) Adding more ambient light can be easily done. (If the lack of light only comes from that.)
But the glitches are a no-go for me.

I propose to drop the glsl150 and OpenGL 3+ renderer for now if everyone agrees. Let's hop on that with Ogre 2.0/2.1. What do you think?

hwoarangmy:
> Here how the game looks like with the OpenGL 3+ renderer:

Hey, why does it works for you and not for me ? T_T

> But the glitches are a no-go for me.

Yes, that's really too visible.

akien-mga:
> I propose to drop the glsl150 and OpenGL 3+ renderer for now if everyone agrees. Let's hop on that with Ogre 2.0/2.1. What do you think?

Yes, Ogre 1.9.0 does say that the OpenGL 3+ rendering is _EXPERIMENTAL_, so we shouldn't bother too much with it (I even had to backport a patch to be able to build it). Maybe it will be stable in Ogre 2.0, then we can have another look.

oyvindln:
> More seriously, I don't know if @akien-mga has tested the gl3+ renderer plugin already, but I'm less and less convinced we should keep those glsl150 shaders at all.
> Let's maybe try again once when 1.10 is out, and anyway, this can wait for 2.0 to happen, IMHO.

Sounds good to me. In 2.0 we'll probably use the new material system anyway.

oyvindln:
Rebased. Should I remove the GLSL150 shader bit then?

hwoarangmy:
> Should I remove the GLSL150 shader bit then?

I guess we all agreed to removing it as it works properly for nobody.

Bertram25:
> I guess we all agreed to removing it as it works properly for nobody.

+1

oyvindln:
Since it works fine, I propose we merge this now.

hwoarangmy:
I'm currently compiling. Do you want to merge before fixing the claimed wall mark bug ?

oyvindln:
That's hard for me to fix as I am not getting it.
EDIT: You could try setting the emissive colour, maybe that will work, though I think that makes it look rather ugly.

hwoarangmy:
> You could try setting the emissive colour, maybe that will work, though I think that makes it look rather ugly.

Ok, I will make some tries and tell you if it works

Bertram25:
Fingers crossed. :)

hwoarangmy:
I've made many tries but as soon as I use the normal map, the top of the claimed wall is not coloured...

@oyvindln I've tried to add emissive color (randomly changing values) but the top is not coloured. Do you see something else to test ?

akien-mga:
Tested the rebased PR, it's good for me.

oyvindln:
You could try setting the alpha value of the lights to 1.0, for whatever reason is set at 0.3. I don't think that will help though. Otherwise I don't know. Though, I have the issue that the normal maps sometimes turn off for a short time, when a new model gets loaded, so maybe that's related.

akien-mga:
IMO we could merge this PR and open an issue for the claimed walls colouring problem. (it looks interesting by the way only partly coloured ;-)

hwoarangmy:
ok, as @akien-mga said, if I am the only one with this bug, let's merge and open an issue. We will see later for partly colored tiles. Anyway, selected tiles are still visible so it should be ok.

PR #450, opened 2015-01-21
Improved tiles in sight / visible tiles computation

hwoarangmy:
To test it visually, you can use console command creaturevisdebug.
The test map empty is interesting to see the shape of the visible tiles and how it behaves when vision is through a 1 tile wide corridor.


source/tests/test_ConsoleInterface.cpp
```diff
@@ -107,8 +103,6 @@ BOOST_AUTO_TEST_CASE(test_ConsoleInterface)
     interface.tryExecuteCommand("test1 123 abc",mt, modeManager); ++count;
     BOOST_CHECK(interface.tryExecuteCommand("aliasedcommand", TestModeManager::ModeType::EDITOR, modeManager) == Command::Result::WRONG_MODE); ++count;
     BOOST_CHECK(interface.tryExecuteCommand("aliasedcommand 1",mt, modeManager) == Command::Result::SUCCESS); ++count;
-    BOOST_CHECK(interface.tryExecuteCommand("aliasedcmd argument1",mt, modeManager) == Command::Result::INVALID_ARGUMENT); ++count;
-    BOOST_CHECK(interface.tryExecuteCommand("aliasedcmd 184467440737095516100",mt, modeManager) == Command::Result::INVALID_ARGUMENT); ++count;
```
oyvindln: Why  are you removing these checks?

Bertram25:
I took some time to play with it, and according to what I saw in the code. this is looking fine to merge.
If nobdody else has something against this PR, I think we should merge it.

akien-mga:
I've tested it shortly, it's working great AFAICS.

PR #454, opened 2015-01-26
Implemented creatures mood

### LINKED ISSUE #11: Creature mood system
See also: http://forum.freegamedev.net/viewtopic.php?p=56960#p56960

A new config file defining the mood will be introduced. It will define mood modifiers (each having a unique name). 
- Each creature starts with a default mood of 1000 (configurable)
- Mood modifiers are computed during the creature upkeep. But not each turn, it will be randomly computed to avoid computing it too often (with a random period between 0 and 5 turns)
- Modifiers can be > 0 (will make creature happy) or < 0 (will make creature angry)
- There are several levels of hapiness: (happy, neutral, upset, angry, furious). They depend on base mood + mood modifiers. When < 0, the creature will be unhappy (angriness level configurable).
- When the creature is happy/neutral, it behaves normally
- Upset creatures have up to 20% chances of stopping to work when working. They engage allied creatures with mood modifier < 0 in sight (if no ennemy in sight)
- angry creatures: same as upset + they might flee (20%) if an enemy is nearby + they don't work at all
- furious creatures: same as angry + They try to go to the portal and leave the dungeon if they reach it. An angry creature for more than 2 minutes will become rogue (cannot be picked up and attacks any player).

hwoarangmy:
I've not implemented creatures becoming rogue. First we need to decide how rogue creatures are. IMO, we should automatically add a seat id = 0 and team id = 0 for them. That will allow to add rogue creatures in the editor.
Moreover, now that mood is implemented, rogue creatures will be angry as they will not get their fee. We need something we can rely on to disable such unwanted behaviour.

WDYT about using a seat id = 0 for that ?

fixes #11

### DIFF
--- a/source/gamemap/GameMap.cpp
+++ b/source/gamemap/GameMap.cpp
@@ -193,6 +193,10 @@ bool GameMap::loadLevel(const std::string& levelFilepath)
     std::string levelPath = ResourceManager::getSingletonPtr()->getGameDataPath()
                             + levelFilepath;
 
+    // We add the rogue default seat (seatId = 0 and teamId = 0)
+    Seat* rogueSeat = Seat::getRogueSeat(this);
+    addSeat(rogueSeat);
+
     // TODO The map loader class should be merged back to GameMap.
     if (MapLoader::readGameMapFromFile(levelPath, *this))
         setLevelFileName(levelFilepath);
@@ -2051,7 +2055,7 @@ void GameMap::addSeat(Seat *s)
     }
 }
 
-Seat* GameMap::getSeatById(int id)
+Seat* GameMap::getSeatById(int id) const
 {
     for (Seat* seat : mSeats)
     {
@@ -2935,10 +2939,16 @@ void GameMap::clearCreatureMoodModifiers()
     mCreatureMoodModifiers.clear();
 }
 
-void GameMap::addCreatureMoodModifiers(const std::string& name,
+bool GameMap::addCreatureMoodModifiers(const std::string& name,
     const std::vector<CreatureMood*>& moodModifiers)
 {
+    uint32_t nb = mCreatureMoodModifiers.count(name);
+    OD_ASSERT_TRUE_MSG(nb <= 0, "Duplicate mood modifier=" + name);
+    if(nb > 0)
+        return false;
+
     mCreatureMoodModifiers[name] = moodModifiers;
+    return true;
 }
 
 int32_t GameMap::computeCreatureMoodModifiers(const Creature* creature) const
--- a/source/gamemap/GameMap.cpp
+++ b/source/gamemap/GameMap.cpp
@@ -2974,7 +2974,7 @@ std::vector<GameEntity*> GameMap::getNaturalEnemiesInList(const Creature* creatu
         Creature* alliedCreature = static_cast<Creature*>(entity);
         for(const CreatureMood* mood : moodModifiers)
         {
-            if(mood->getCreatureMoodType() != CreatureMood::CreatureMoodType::creature)
+            if(mood->getCreatureMoodType() != CreatureMoodType::creature)
                 continue;
 
             const CreatureMoodCreature* moodCreature = static_cast<const CreatureMoodCreature*>(mood);
--- a/source/gamemap/GameMap.cpp
+++ b/source/gamemap/GameMap.cpp
@@ -259,6 +259,7 @@ void GameMap::clearAll()
     // NOTE : clearRenderedMovableEntities should be called after clearRooms because clearRooms will try to remove the objects from the room
     clearRenderedMovableEntities();
     clearTiles();
+    clearCreatureMoodModifiers();
 
     clearActiveObjects();
 
--- a/source/gamemap/GameMap.cpp
+++ b/source/gamemap/GameMap.cpp
@@ -24,6 +24,9 @@
 
 #include "gamemap/GameMap.h"
 
+#include "creaturemood/CreatureMood.h"
+#include "creaturemood/CreatureMoodCreature.h"
+
 #include "gamemap/MapLoader.h"
 
 #include "goals/Goal.h"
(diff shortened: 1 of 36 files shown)


source/gamemap/GameMap.cpp
```diff
@@ -2888,3 +2913,69 @@ void GameMap::updateVisibleEntities()
...
+{
+    // We delete map specific mood modifiers
+    for(std::pair<const std::string, std::vector<CreatureMood*>>& p : mCreatureMoodModifiers)
+    {
+        for(const CreatureMood* creatureMood : p.second)
+            delete creatureMood;
+    }
+
+    mCreatureMoodModifiers.clear();
+}
```
oyvindln, line 2940: This isn't being done in the destructor from what I can see, so there will be a memory leak.

source/utils/ConfigManager.h
```diff
@@ -26,12 +26,21 @@
...
 {
 public:
+    enum CreatureMoodLevel
+    {
+        Happy,
+        Neutral,
+        Upset,
+        Angry,
+        Furious
+    };
```
oyvindln: Wouldn't it be better to put this in CreatureMood.h? Also, we should use enum class for new enums.

source/gamemap/GameMap.cpp
hwoarangmy, line 2940: That's correct. I forgot to call it from clearAll

source/utils/ConfigManager.h
hwoarangmy: I will have a look at enum class

Bertram25: True. Another flaw of c fixed by this strongly typed enums. I'll try to keep that in mind as well when coding from now on.

source/gamemap/GameMap.cpp
Bertram25, line 2940: Yeah, nice catch.

Bertram25:
Seems you already implemented the leave dungeon method. :)

I'll test asap what's done. 

> IMO, we should automatically add a seat id = 0 and team id = 0 for them. That will allow to add rogue creatures in the editor.

Sounds pretty relevant to me.

> Moreover, now that mood is implemented, rogue creatures will be angry as they will not get their fee. We need something we can rely on to disable such unwanted behaviour.

You mean for feral/rogue creatures, I guess. Then, yes, as discussed before, teamless creatures shouldn't get hungry, dizzy, or angry. They should fight until death, if not fleeing because of being wounded, IMHO.
@Danimal696 ? @akien-mga @oyvindln ? WDYT?

hwoarangmy:
IIRC, the seat is used in many places and adding checks for nullptr seats everywhere will hurt code readability.
That's why a seat with a forbiden seatid and teamid seems the best solution to me.

I agree that rogue creatures should not be affected by hunger/sleep/mood. But concerning fleeing, should they fight even when weak ?

Bertram25:
> I agree that rogue creatures should not be affected by hunger/sleep/mood. But concerning fleeing, should they fight even when weak ?

IMHO, yes they should still flee once weak. But no more for other conditions.

source/utils/ConfigManager.h
hwoarangmy: I've had a look at enum class. From what I've seen, it allows forward declaration (which is really good) but they should be declared outside the class (that makes sense since the the most important reason to declare them in the class was to gather enum values).

I would also say that we should add in the coding good practices to use enum classes for enums (and start replacing the unscoped enumerations currently in the code)

Bertram25: > I would also say that we should add in the coding good practices to use enum classes for enums (and start replacing the unscoped enumerations currently in the code)

We can, but there is no urgency. Let's switch "when we see one in our way" unless you plan to focus on that in a particular commit, of course. :)

Btw, with mood, and hopefully soon, workshop, new forge, and normals support. 0.5.0 will truly shine. :D

hwoarangmy: > Let's switch "when we see one in our way"

Yes, I was thinking about the same as for nullptr: when we see one. No need to rush.

> Btw, with mood, and hopefully soon, workshop, new forge, and normals support. 0.5.0 will truly shine. :D

Yep. And FOW ^^

Bertram25: Oh my, of course. I forgot the main new feature. :S Eh eh.

Bertram25:
Btw, interesting presentation for those who are still learning C++11 out of a wanna-be-up-to-date fashion:
http://www.enricozini.org/2014/cxx11-talk-notes/

source/utils/ConfigManager.h
hwoarangmy: > Oh my, of course. I forgot the main new feature. :S Eh eh.

Because you consider it as already part of the game. But not in the 0.4.9 ;-)

hwoarangmy:
We should take care when using c++11 stuff to not loose compatibility with compilers. There is not much to fear about gcc but MinGW and MSVS do not implement every features. And if want coders to help, we should try to keep the code as compiler compatible as possible. Building OD development environment is already complex enough to not make them change OS / compiler version / ...

Bertram25:
Sure. I guess that's why @oyvindln is introducing c++11 concepts one by one. ;)

akien-mga:
> IMO, we should automatically add a seat id = 0 and team id = 0 for them. That will allow to add rogue creatures in the editor.

I agree, seat id = 0 and team id = 0 sounds like the best solution. And no team color I guess.

akien-mga:
The PR seems to work fine, I managed to get some unhappy creatures by neglecting them completely :-P
I'm not sure how the mood modifiers work for now though, I monitored two spiders in an empty dungeons and saw their mood oscillate between 0 and -500 (-1000 sometimes, but not for long).

hwoarangmy:
> I'm not sure how the mood modifiers work for now though

The mood modifiers are defined in config/creaturesmood.cfg. For spiders, they get -500 per cavehornet they see. However, mood is not computed each turn (but between a random number between 1 and 5 turns - 5 turns being more or less 3 seconds).

hwoarangmy:
In the "real" levels, I will also remove the inactive players (when unused) to switch to the rogue seat so that rogue creatures do not bother with mood/hunger/...

Bertram25:
> In the "real" levels, I will also remove the inactive players (when unused) to switch to the rogue seat so that rogue creatures do not bother with mood/hunger/...

Good idea. Plus, it will make the seat config window cleaner.

hwoarangmy:
Now creature mood is completely implemented as stated in #11

Danimal696:
good one, even better, the way to kick creatures from your dungeon is almost paved ;D

hwoarangmy:
> the way to kick creatures from your dungeon is almost paved

I've been also wondering if we want creatures to leave dungeon when dropped on the portal ?
For now, we can already kill them by slapping or letting them gowithout paying (but while we have no door, it's pretty hard to handle)

hwoarangmy:
I've rebased

Bertram25:
Testing now.

Bertram25:
Works perfectly, I even lost control of one unit. So, IMHO, we can merge. @akien-mga ok for you?

hwoarangmy:
> I even lost control of one unit

To do that, it's pretty easy. In the 2v2 skirmish small test map, just drop your big knight in your ally dungeon. After 3 pay days, he will be furious and won't be able to leave. And then, he will destroy your ally dungeon as easily as if it was @Bertram25 's

Bertram25:
> as easily as if it was @Bertram25 's

Don't insult the monkeys. ;) (And don't feed them)

hwoarangmy:
> Don't insult the monkeys

I wouldn't dare. I know how @tomluchowski loves them ^^

akien-mga:
Merging then :-)

PR #457, opened 2015-01-27
OpenDungeons new GUI version 1.0 milestone reached! :)

### LINKED ISSUE #419: Upgrade the remaining UI
Most notably, the Check-boxes and combo-box need a lifting.
The tooltip may look better if they are black with some white text.

Bertram25:
Made the remaining widgets use the new skin:
- Sliders,
- MultilineList
- Spinner
- Popup Menu
- ...

I also turned the tab content pane background to black, to make it reusable
in other windows.

I also added a WidgetTest layout file to quickly have a look at the new widgets
when using CEED.

As pointed out by @hwoarangmy, The sliders need a better skin.
I also have to add back support for fancy tab widgets per faction.

Those two things will be part of another PR.

Fixes #419


akien-mga:
Looks very fine to me.

After a fight with CEED who wouldn't let me open the WidgetTest.layout file, I noticed that I had to define a CEGUI target version of 0.8 in the settings so that it works. The default value was 1.0, I might patch that in my CEED package :-)

Bertram25:
Good to know.

hwoarangmy:
I got an error. There is a typo. In OpenDungeonsSkin.scheme, you define PopupMenuFrameMiddle and in OpenDungeons.looknfeel, you use PopupMenuMiddle.

hwoarangmy:
BTW, I've tested after fixing the typo and it is pretty nice ^^

Bertram25:
Thanks :)
I've fixed the typo. Sorry, it must have slipped through when I was almost done with all that.

akien-mga:
Still works fine, I merge :-)

Bertram25:
Cool, next step will be to improve a bit the look of sliders and test them using the new options windows. :)

PR #466, opened 2015-01-29
CEGUI console window + ogre cmake fix + cleanups

oyvindln:
Changed to use a CEGUI window for the console and got rid of the old console class. (should we remove aswrapper too?)
Updated FindOGRE.cmake to fix an issue where plugin folder is not found.
Got rid of a lot of duplicade code in application modes.


hwoarangmy:
I've tested with MSVS and, as expected, compilation is broken. The 2 main reasons are in ODPacket the use of the template with varargs. The second being the "= delete" that doesn't seem to be recognized. Since ODPacket is included by most of the files, there were 625 errors so I might have missed some ^^

We should decide if we drop compatibility with MSVS (I would be against but if you all don't care, I can live with it) or if we try to adapt.

oyvindln:
What MSVS version are you using? Both features are supposed to work in VS2013, but not earlier versions.

hwoarangmy:
MSVS 2012. 2013 is for windows 8 so I can't use it ^^

oyvindln:
Looks like some versions support Windows 7: https://stackoverflow.com/questions/20516271/install-visual-studio-2013-on-windows-7

hwoarangmy:
Well, in reality, the question is: should we support MSVS 2012 ?

oyvindln:
Given that it would take some work to make it compatible, what is the purpose of it? For doing release builds on windows, we can use MinGW (or VS2013), and AFAIK there is no one of us actively using VS2012.

hwoarangmy:
Well, IMHO, it is more a question of closing a door for MSVS 2012 developpers that would like to contribute to OD. From my experience, the most complex the dev environment is, the less chances you have to get new contributors. I can tell you that when I started to want to contribute to an open source game (when I joined here), I first tried to compile another open source game based on another oldie: wildands. Because I thought there was too many dependencies to install: lua, many SDL stuff, python, ... I gave up.
I tried a little bit harder for OD because I prefered to work on a DK clone than on a Settlers 2 clone and because I was interested for technical reasons in the project (ie Ogre and C++). Otherwise, I would have probably also given up (at that time, there was no how to compile since I wrote it later and I was the only one to use windows)...

oyvindln:
Yes, that would make sense if it didn't work in MSVC at all, though in this case it's only about having to update to the latest version. (And not using an pre-2012 version, which would even more work to support.) I can't imagine there being that many people out there who is forced to use 2012 and nothing else?

Bertram25:
Hey there, :)

I'm on MSVC 2012 as well, and on Windows 7 too, (when I boot on windows).
I also think that dropping MSVC support is a mistake as many windows developers use it. And if we claim being open-source, we shouldn't remove support for MSVC compilation because it isn't a free as in freedom to use tool. It's more some kind of sectarism to me. If windows doesn't want to be open, it's their problem, but I'm all for welcoming MSVC users if they want to participate.

Also, while I'm all for a complete set of c++11 feature use, would it only be for learning and enjoying the new possibilities, before stating we should drop MSVC2012, could you give us some time to see whether it's feasible (as in not too hard) to move to 2013 first?
Simply put, for instance, Ogre doesn't provide any prebuilt dependencies package for MSVC2013 and Cegui doesn't either.
So building the deps for 2013 is so far untested AFAIK. For those reasons, I'd like we do not go too fast here.

Again, I understand the will to use new c++ features, but we have to take in account all the constraints.

Best regards,

hwoarangmy:
I will try to download MSVS2013 ASAP and to compile OD from there (including dependencies) as I'm pretty used to it since I had to do that for MSVS 2012 ^^ Not sure when I will be able to do that, though, as compiling boost and ogre is pretty long.

Anyway, compilation on MSVS 2012 is already broken so I guess it shouldn't block merging this PR.

Bertram25:
> I will try to download MSVS2013 ASAP and to compile OD from there (including dependencies) as I'm pretty used to it since I had to do that for MSVS 2012 ^^ Not sure when I will be able to do that, though, as compiling boost and ogre is pretty long.

Right there, too! ;)

> Anyway, compilation on MSVS 2012 is already broken so I guess it shouldn't block merging this PR.

Ok, but I'd still request to revert frame sleep part constant change. I mean the 2.0 -> 1.0 change.
And I still need to see the new console live. :D

hwoarangmy:
> I still need to see the new console live

I've also not tried this PR yet. But hopefully during the WE ^^

hwoarangmy:
I confirm MSVS2013 can be installed on windows 7 sp1. I will try to compile OD ASAP.

Bertram25:
> I confirm MSVS2013 can be installed on windows 7 sp1. I will try to compile OD ASAP.

Cool :) I'll have a go at it, too.

Bertram25:
Btw, I guess issues about this PR are fixed, right?
I'll test asap, if that's the case.

oyvindln:
Yeah, except the extra spaces.

hwoarangmy:
ok, I confirm I could build OD with MSVS2013. I will update the how to as some changes are needed (including in the dependencies).

In OD code, there is an error in the ConsoleCommand class. To fix it, I did like it is said here:
http://stackoverflow.com/questions/23778883/vs2013-stdfunction-with-member-function

In our, code, it looks like:

```
return cSetFrameListenerVar<float>(std::mem_fn(&ODFrameListener::getActiveCameraNearClipDistance),
                                     std::mem_fn(&ODFrameListener::setActiveCameraNearClipDistance),
                                     ODFrameListener::getSingleton(),
                                     "near clip distance", args, c);
```

@oyvindln @Bertram25  Could you please confirm you are ok with this change and that it compiles/run with GCC/linux ? For my part, I've tested with MinGW and it works like that.

If you are ok with it, I will do a PR with the needed changes to compile with MSVS2013.

hwoarangmy:
I've updated the how to compile wiki for MSVS2013 (I've replaced MSVS2012 since it doesn't compile anymore)

Bertram25:
> @oyvindln @Bertram25 Could you please confirm you are ok with this change and that it compiles/run with GCC/linux ? For my part, I've tested with MinGW and it works like that.

I tried the snippet at ConsoleCommand.cpp Line 585 and it's working. :)

I'll test the sleep constant part now.

Bertram25:
Apart from @hwoarangmy change request, it is working fine. IMHO if someone else can test this, we can merge.
@hwoarangmy  Mind fixing the MSVC2013 compilation afterward?

hwoarangmy:
Since you've tested with gcc and I've tested with MinGW and MSVS2013, I merge and do the needed changes for MSVS2013

PR #467, opened 2015-01-29
Coefficient weakness + room & trap price configurable

hwoarangmy:
To test weakness coef, I've changed the skirmish small test map to check it. Now, the big dragon is under the player starting position and it shouldn't flee until dead

### DIFF
--- a/source/rooms/Room.cpp
+++ b/source/rooms/Room.cpp
@@ -365,25 +365,25 @@ int Room::costPerTile(RoomType t)
         return 0;
 
     case treasury:
-        return 25;
+        return ConfigManager::getSingleton().getRoomConfigInt32("TreasuryCostPerTile");
 
     case dormitory:
-        return 75;
+        return ConfigManager::getSingleton().getRoomConfigInt32("DormitoryCostPerTile");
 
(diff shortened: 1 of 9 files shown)


source/rooms/Room.cpp
```diff
@@ -365,25 +365,25 @@ int Room::costPerTile(RoomType t)
         return 0;
 
     case treasury:
-        return 25;
+        return ConfigManager::getSingleton().getRoomConfigInt32("TreasuryCostPerTile");
```
oyvindln: I think we should really avoid using singletons like this.

Bertram25:
Nothing to say. I'll test asap.

Bertram25:
Works fine. We can merge IMHO.

Bertram25:
I'll take the responsibility to merge now as I can't see anything really hurting here.

hwoarangmy:
> I'll take the responsibility to merge now as I can't see anything really hurting here.

That's a huge responsability to merge my PRs ;-)

Bertram25:
It's more to have someone to yell at, just in case. ;)

PR #484, opened 2015-02-05
Added "call to war" spell + workers higher priority to carry things

hwoarangmy:


akien-mga:
Big PR! :-) I'll have a look at it in the afternoon.

akien-mga:
This works very well. I played around with the call to war spell, it's great :-P
I haven't spotted any regressions so far, I'll let @Bertram25 or @oyvindln comment on the code :)

About commit 7616c86, it seems to work even though it does not seem to chose the absolute nearest building (see screenshot), but some more tests would be needed to see how it behaves exactly. In my screenshot, workers carrying gold retrieved at the top and at the bottom right (i.e. the arrows) dropped their bag in the tile that I circled in red, even though the workers coming from the bottom right should have used the treasury room in the middle.

od-gold

akien-mga:
For the editor change, it seems to be possible to move lights around indeed (and to destroy them with a right click), but it's quite hard to actually click on them properly.

hwoarangmy:
> it's quite hard to actually click on them properly.

Yep, I've seen it. I guess it's because the mesh is a little bit too small (or if it is the perspective) but sometimes, it is a bit hard to click on them.

> workers carrying gold retrieved at the top and at the bottom right

In your screenshot, crow distance is nearly the same on both (I would say the one where gold was dropped is the nearest). To avoid computing paths , there is a "magic" number to choose if other paths should be computed or not. ATM, this number is pretty high (2.0) and it explains why the other path wasn't computed. If it is needed, we can lower it.

Danimal696:
which one did you use? i want a pic of it :)

hwoarangmy:
The training dummy 1 (the one on walls) ^^

Bertram25:
I'll rebase my PR, and test this one now.

Bertram25:
- AI does spawn workers but I haven't seen yet when they start to spawn less of them.
- Call to war spells are functioning perfectly fine. Well, they made me lose but it was plainly my fault. ;)
- Training hall is working as expected.
- My kobolds go to the nearest carrier tile and even if it's not perfect, it's much more than fine for now, IMHO.
- Lights are movable in the editor but they are indeed hard to click on. Deleting them with a right-click seems much easier to achieve somehow. Can someone else confirm?
- Not linked to this PR but, the editor menu seems a bit messy now. The lights tab is empty, the add worker button only adds kobolds, but not the default seat worker, the Spell tab shouldn't be there, IMHO. I'll open an issue if everyone agrees about that.
- The code looks good IMHO even if I didn't get the last commit message but I guess it must be linked to previous changes.

Bertram25:
If someone else tells it's ok, I think we can merge.

akien-mga:
> Lights are movable in the editor but they are indeed hard to click on. Deleting them with a right-click seems much easier to achieve somehow. Can someone else confirm?

Yep I confirm. For now, a good workaround to the lights being hard to move would be to have to possibility to add new lights using the editor (this way you could delete lights with right click, and then place a new one, if moving them around is too hard).

> Not linked to this PR but, the editor menu seems a bit messy now. The lights tab is empty, the add worker button only adds kobolds, but not the default seat worker, the Spell tab shouldn't be there, IMHO. I'll open an issue if everyone agrees about that.

+1

Merging then.

hwoarangmy:
> Not linked to this PR but, the editor menu seems a bit messy now. The lights tab is empty, the add worker button only adds kobolds, but not the default seat worker, the Spell tab shouldn't be there, IMHO. I'll open an issue if everyone agrees about that

ok for me ^^

Bertram25:
K, I'll open issues while it is compiling. :)

PR #488, opened 2015-02-11
Drop FreeMono, add appdata file and other misc fixes

akien-mga:


Bertram25:
> IMHO, for files, we should only use english standard chars (as there can be pretty nasty behaviours when we use special chars).

This should be added in the guidelines while I don't know where, yet.

Bertram25:
This looks good, the gcc4.8+ + Mingw reference update apart.
Let's merge once this is fixed. :)

akien-mga:
> This looks good, the gcc4.8+ + Mingw reference update apart.

I just rebased to fix that.

akien-mga:
Nota: as we're still trying to understand how PRs react to rebasing, in this case it seems that all comments made on the removed commit (and the commit that came afterwards, c10105b) are lost.

So if one rebases to fix the first commit of the PR, all commit-based comments should be lost.

hwoarangmy:
yes, that's what happens when rebasing. I've seen that comments on file changes (when you click on the file changed list) are lost only when the difference changes.
But in your case, when you have a PR, I think it is better to add a new commit to fix what is asked instead of changing an existing commit.
BTW, I'm curious to see what happens if you change a commit that is already merged ^^

akien-mga:
> But in your case, when you have a PR, I think it is better to add a new commit to fix what is asked instead of changing an existing commit.

True, I could have done that. Since the new commit would basically be a complete rewrite of the former, I chose to override it though :)

> BTW, I'm curious to see what happens if you change a commit that is already merged ^^

We would probably lose all commit-based discussions in the development branch after the modified commit. We don't discuss much on merged commits though, mostly on issues and PRs.

And IMO we should refrain from using `git push -f` on the development branch anyway (apart from squashing together 20 single-change commits by @Danimal696 learning how to use git ;-).

hwoarangmy:
> We would probably lose all commit-based discussions in the development branch after the modified commit. We don't discuss much on merged commits though, mostly on issues and PRs.

Well, I hope it is not possible to change a commit after it is merged. I guess they are PR related and once a PR gets merged, its commits cannot be changed. I can't think it is possible...

akien-mga:
> Well, I hope it is not possible to change a commit after it is merged. I guess they are PR related and once a PR gets merged, its commits cannot be changed. I can't think it is possible...

Ah you mean changing a commit in a PR once its merged? It's not possible AFAIK, since updating the branch from which you created the PR does not further update the PR once it's merged.
What is possible though, it to edit commits directly in the main repo, but we should refrain from doing that as much as possible.

hwoarangmy:
> What is possible though, it to edit commits directly in the main repo, but we should refrain from doing that as much as possible.

Wow, that's so nasty I didn't even think about it ^^ Even if I guess it's possible

Bertram25:
> Wow, that's so nasty I didn't even think about it ^^ Even if I guess it's possible

Almost anything is possible, just like changing the commit authors, date and so on. But I don't think anybody would call that sane without a very specific reason.

Bertram25:
I'll need to test because of the CMakeLists.txt file change but I do think it looks fine enough to merge otherwise. :) Thanks for taking care of all this.

oyvindln:
Since we're not using the renamed texture, wouldn't it be better to remove it and put it in the source repo instead?

akien-mga:
> Since we're not using the renamed texture, wouldn't it be better to remove it and put it in the source repo instead?

It is, but IMO there are other textures that are currently unused, so we should probably open another issue against it and do a proper cleanup of the repo. WDYT?

hwoarangmy:
IMHO, we should remove every not needed stuff since we have the media repository that should already have those.

akien-mga:
> IMHO, we should remove every not needed stuff since we have the media repository that should already have those.

+1, and we can reduce the size of our releases :)

Bertram25:
Sure, we'd need a list of the textures/models we can move back to media, though.

Bertram25:
I just tested. It is working just fine. IMHO, we can merge and take care of unused textures in another issue/PR. I'll open one.
@hwoarangmy Do you want to test on windows just in case?

hwoarangmy:
It works well

PR #495, opened 2015-02-12
Added entities in OO way + fixed editor + used crafted traps seat + misc updates

hwoarangmy:


Bertram25:
Looks like a great cleanup PR. :D 
I'll check https://github.com/hwoarangmy/OpenDungeons/commit/b2f887c4f2a9b8599b6390baeb52c9f9f9858a8e a bit more and test all this asap.
I've also got to finish my mockup of the research tree, btw.

hwoarangmy:
> I'll check hwoarangmy@b2f887c a bit more and test all this asap.

Funny, that was not the biggest code breaker IMO ^^
BTW, hopefully, I should post tonight a PR about code cleaning of the GameMap (I wanted to remove the included Tile.h dor a while and the Tile's friendshipness with GameMap and ODServer).

Bertram25:
> Funny, that was not the biggest code breaker IMO ^^

Yep, but the most interesting for me to fully understand. :D 

> BTW, hopefully, I should post tonight a PR about code cleaning of the GameMap (I wanted to remove the included Tile.h dor a while and the Tile's friendshipness with GameMap and ODServer).

Cool. That said, I know you're refraining yourself from working on the research tree, so I'll try to work on the mockup and the gui layout in priority. :)

hwoarangmy:
> I know you're refraining yourself from working on the research tree, so I'll try to work on the mockup and the gui layout in priority. :)

:-)

Bertram25:
Ok,
- Editor is working fine.
- Moving lights is working.
- I cheated and indeed kobold can carry a lot when you change the value ;)
- To test the traps, I made every of them need only one point. So this is working + the crafted trap seat logic seems to work fine as far as I played.
- Placing default workers is also working.

The PR is fine :D 
Btw, the last game I played was fantastic. With the FOW, the creatures mood, the crafted traps and the workers carrying corpse and enemy toughly stolen gold, it was epic. :D 
The call to war is a super addition, really.
Not linked to this PR, but I still think the pay day is coming too soon, I think its turn length should be multiplied by 1.6-2.0 at least.
Also, I had creature turning against me while fighting! I guess receiving hits made them mad, but I do think an unhappy creature should run away from battle faster to avoid this situation.
I also wondered:
When a creatures meets a natural enemy, is the same seat check done? Because you can't have a creature going mad when encountering other not allied seats natural enemies. On the contrary, it should happy to hit them, right?

In any case, I'll merge. :)

hwoarangmy:
> I think its turn length should be multiplied by 1.6-2.0 at least.

As I said, I've set the time between 2 pay days pretty short to allow testing easily. Now that it is tested, we can increase it. If you see a value that fits, please don't hesitate to change. To be honest, I've not played enough real games to really have an idea. I've just seen in the skirmish test map that you need really much gold ^^

> When a creatures meets a natural enemy, is the same seat check done?

We test only allied units. But in any case, an angry creature will attack ennemies rather than natural ennemies.

> I do think an unhappy creature should run away from battle faster to avoid this situation.

That's something we should think about. I guess it will be a good time to play a lan to check that once we have the research tree ^^

akien-mga:
> When a creatures meets a natural enemy, is the same seat check done? Because you can't have a creature going mad when encountering other not allied seats natural enemies. On the contrary, it should happy to hit them, right?

+1 for a mood bonus when creatures attack non-allied natural enemies :-)

Bertram25:
> As I said, I've set the time between 2 pay days pretty short to allow testing easily. Now that it is tested, we can increase it. If you see a value that fits, please don't hesitate to change. To be honest, I've not played enough real games to really have an idea. I've just seen in the skirmish test map that you need really much gold ^^

Don't worry, nothing personal at all. In fact, I'm more than happy we're already at the time where we want to tweak values. :D I've done a PR about that among other things, btw.

> We test only allied units. But in any case, an angry creature will attack ennemies rather than natural ennemies.
> That's something we should think about. I guess it will be a good time to play a lan to check that once we have the research tree ^^

You're right, indeed. But after thinking about it more, I'm sure that adding a small mood bonus when dealing to an enemy should be considered. :) I'll get back to that once we have the research tree and we've done a lan or two.

PR #497, opened 2015-02-14
Changed enumeration values from enum class FloodFillType to avoid repeat...

hwoarangmy:
...ing type in values


akien-mga:
Cool, a code change that I'm competent enough to review for once :-P

akien-mga:
It looks straight-forward, and I couldn't spot FloodFillTypeG\* leftovers, so I'm merging.

PR #502, opened 2015-02-15
Avoid sending a message to every player when a creature owned by an inactive keeper becomes rogue

### LINKED ISSUE #501: In Big test map, we get a message telling that PitDemon5 leaves your team while you don't won the creature

hwoarangmy:
fixes #501


akien-mga:
Testing it now. BTW, I seem to have problems with incremental compilation, each time I build a new PR, it rebuilds the whole project. This might be due to the fact that I use the same git clone to test all PRs (in various build directories though, build-hwo for @hwoarangmy, build-ber for @Bertram25, etc.). I wonder how you handle that?

akien-mga:
No regression, I merge.

PR #504, opened 2015-02-16
Disable angelscript + openal fix + minimap

oyvindln:
Completely disabled angelscript.
Add workaround for OpenAL segfault.
Fixed minimap camera issue + various cleanups in minimap
Fixed a bunch of warnings


source/render/Gui.cpp
```diff
@@ -243,9 +243,9 @@ void Gui::assignEventHandlers()
             CEGUI::Window::EventMouseClick,
             CEGUI::Event::Subscriber(&fighterCreatureButtonPressed));
 
-    mSheets[inGameMenu]->getChild(MINIMAP)->subscribeEvent(
+/*    mSheets[inGameMenu]->getChild(MINIMAP)->subscribeEvent(
```
Bertram25: Why did you simply comment those?

source/render/ODFrameListener.cpp
```diff
@@ -92,8 +92,6 @@ ODFrameListener::ODFrameListener(Ogre::RenderWindow* renderWindow, Ogre::Overlay
 
     mGameMap = new GameMap(false);
 
-    mMiniMap = new MiniMap(mGameMap);
```
Bertram25, line 95: I must say at first I wondered why you duplicated the minimap in each mode. But after thinking about it, it makes sense, for not all the modes will use a minimap.

Bertram25:
I comment about the commented code apart, the code change looks fine to me.
Is the ExportTileToPacket() a fine addition in your opinion, @hwoarangmy? 
EDIT: I'm testing now.

Bertram25:
Ah, btw, for later commit messages, you can add comments like:

Fixes #XXX XXX being the bug number. And when we'll merge the issue will automatically close.

Bertram25:
Everything is working as expected. Can you remove the commented code, or is there a reason you kept it?

source/render/Gui.cpp
oyvindln: That was just an oversight.

source/render/ODFrameListener.cpp
oyvindln, line 95: My goal was to have the minimap exist where it was used, rather than having a vague global object that may or may not be initialised and has to be manually cleared etc. That said there is still a fair bit of code duplication between editormode and gamemode which could probably be improved.

oyvindln:
> Is the ExportTileToPacket() a fine addition in your opinion, @hwoarangmy? 

Just if it wasn't clear, it's just a rename of the exportToPacket function that was overloaded in tile, as it had different parameters to the default one from the base class, as it generated a warning. Not sure if it's the best name but I couldn't come up with anything else.

source/render/ODFrameListener.cpp
Bertram25, line 95: True enough. We still have a bit of history here. But at least, you untangled the minimap, framelistener and the camera manager code. :)
The next interesting thing IMHO is to reduce the minimap update/draw calls to their minimum. There might be a good perf impact.

Bertram25:
The function name looks fine to me. Though, I shall let @hwoarangmy 's word prevail, for he is the maker of the most of it. </Epic sentence>

hwoarangmy:
> Is the ExportTileToPacket() a fine addition in your opinion, @hwoarangmy?

yes, I understand why he changed it and if it is more readable like that, no problem ^^

source/render/ODFrameListener.cpp
hwoarangmy, line 95: I've already run a profiler and filling the minimap was the most time consuming back then (before @Bertram25 's PR). I've not tried after.

oyvindln, line 95: Last time I checked it still was. Granted it was still way below the parts in ogre code.

Bertram25:
I guess the code looks good, right? I tested and there is no regressions AFAIK.

hwoarangmy:
> I guess the code looks good, right?

Yes but there was a change in the makefile. Please do not merge until I test tonight.

Bertram25:
> Yes but there was a change in the makefile. Please do not merge until I test tonight.

Don't worry, I won't. :)

akien-mga:
Builds and runs fine here :)

oyvindln:
The last commit enabled a bunch of warnings that I was already using. There are more that we may want to use, but some of them triggered for a lot of things, so I left them out for now.

hwoarangmy:
It builds and runs fine here, too. I merge

Issue #133, opened 2014-09-10
Some dirt wall tiles get a golden colouring


akien-mga:
It's been around for some time already, but now that @hwoarangmy has tackled bug #51, fixing this bug could be a nice further enhancement before the 0.4.9 release.

See the screenshot:
![od-bug-golden-dirt-walls](https://cloud.githubusercontent.com/assets/4701338/4220434/5e762d1a-3901-11e4-8b90-4118aeebf60d.jpg)

Some dirt wall tiles get a funny colouring. From the screenshot, it seems to be:
- Around a newly claimed wall (though not always)
- At the ends of a wall with a width of 1 tile

Bertram25:
This comes from the appliance of a normal on the first instance of dirt wall. I'll handle this.

akien-mga:
Fix confirmed, thanks :-)

PR #507, opened 2015-02-17
Goblin updated with normals Fixes #503

### LINKED ISSUE #503: The troll bed material should be tweaked to use normals.
As seen here: http://forum.freegamedev.net/viewtopic.php?p=62443#p62443

For now, with normals applied, the bed woods are looking metallic shiny.
Work on the different light values should be done to fix that and finally import the models with normals support.

Danimal696:


Danimal696:
ill upload them as i update the models

akien-mga:
> ill upload them as i update the models

Great! :) Tell us when you're done with the first batch.

Danimal696:
Okay, its best to have this pull open for a bit

Bertram25:
In any case, even if we merge, you can still push other commits and reopen a new PR. :)

Bertram25:
OMG, I can't wait to test all that :D

Danimal696:
its looking better now

Danimal696:
you can merge for now, ill make a new batch this weekend

Bertram25:
With normals, the forge and traininghall ground textures are gorgeous :D I'll have a look at the creatures now.

Bertram25:
To me it's working great. Some test on nvidia/proprietary drivers may be needed, though.

akien-mga:
> To me it's working great. Some test on nvidia/proprietary drivers may be needed, though.

I'm on it :)

akien-mga:
I see no difference with intel and nvidia, but I do see one good old issue coming back:
od-dan-intel

Yep, #133 :-)
@Danimal696 did not touch the wall models, so there's definitely something wrong in the way we render normals.

hwoarangmy:
@akien-mga For me, this issue has never been fixed ;-)
Could you try to edit the map and add rock tiles ? On my computer, the dirt tiles that also exist in the same level as rock tiles are not displayed like that

akien-mga:
> For me, this issue has never been fixed ;-)

Oh :-o I didn't know.
Adding rock tiles doesn't fix the issue, even in the editor I get broken dirt tiles next to the rock tiles. I also tested the edited level in skirmish mode, same issue.
od-strange-dirt-tiles

Danimal696:
i also noticed one thing, normals are not applied onto floor tiles, no doubt the odd pass system is the culprit

hwoarangmy:
> I also tested the edited level in skirmish mode, same issue

In your screenshot, the dirt and rock tiles are different (we can see a north oriented rock tile and a single one while the buggy dirt tiles are west, east and horizontal). Could you please try to have the exact same tiles (for example, 2 tiles dirt and rock with water around) ?

akien-mga:
> Could you please try to have the exact same tiles (for example, 2 tiles dirt and rock with water around) ?

I'm not 100% sure what you mean. From what I understood you think that if a rock tile is present with a certain orientation, then the dirt tiles with the same orientation won't have the bug?
From my tests I can't really find a reliable rule :-/

hwoarangmy:
> you think that if a rock tile is present with a certain orientation, then the dirt tiles with the same orientation won't have the bug?

That's what I've seen so far. For example, If I only add a 1x1 rock tile to the 2v2 skirmish map, the 1v1 dirt tiles will be correctly displayed while the other dirt tiles will not. Note that I've tried to add them at first (0x0) or at last (50x50) and it doesn't change anything. So, it seems like it is not the loading order that matters.

Bertram25:
I guess we need a minimal example level. This way, we could hunt down the way the tiles are created/refreshed, and surely obtain more info on the culprit code and/or material, IMHO.
Does this means creating a 2x1 level would be enough for the bug example level?

Bertram25:
Great, more normals! :D 

Btw, it seems you removed the normal for the bookcase but didn't remove the corresponding texture (which is white, btw.). Is that "normal"? ;)

Bertram25:
I tested them, works great. The game looks gorgeous. (Just start 0.4.9 and you should have a shock when seeing the changes...) Good to be merged IMHO. Please test and do so if it's ok.

Danimal696:
normals go a long way, and i barely started on creatures, ill erase the white texture and kill some indentations on kingthcoffin material

hwoarangmy:
That's also good to me. Should we merge even with the bug that @akien-mga saw ?
@Danimal696 Do you see a reason why there could be this behaviour ?

Bertram25:
> That's also good to me. Should we merge even with the bug that @akien-mga saw ?

EDIT: Depends, but I think yes. (In yes, I think we should merge.)
Is the issue here before this PR? IMO it is, as it is the case for me, too (before the PR).
In that case, this is not a regression due to the PR but rather to the way passes are handled by Ogre, and we'll have to use a minimal example to track down the problem.
Last word is on @akien-mga 's shoulders, though. ;)

oyvindln:
> Is the issue here before this PR? IMO it is, as it is the case for me, too (before the PR).

I also have this issue (without this PR), so it's something wrong in the code, but I don't know what. The affected tiles look like they have lighting applied more than once or something.

Bertram25:
>  The affected tiles look like they have lighting applied more than once or something.

Yeah, that's also my impression but it might be more twisted.

hwoarangmy:
ok. Then, everybody have the bug. I thought I was the only one with this bug. Maybe not after all ^^
In any case, that's not a regression from this PR so I merge

PR #508, opened 2015-02-17
Implemented research logic + small bug fixes

hwoarangmy:


hwoarangmy:
@Bertram25 if you want to test your gui, once the research list is configured, you need to send a server message ClientNotificationType::askSetResearchTree (see ODServer for more details on the expected data). The list of research will be checked (note that the order matters). If it is validated (ie all the researches sent are available in this order), the client will receive a message ServerNotificationType::researchTree that will update the Player class. To update the gui, you need to periodically test bool Player::guiResearchRefreshedPending() to know if the tree changed. If yes, you might need to add a function in the player class to export the needed data for you.
Note that testing the boolean is needed because you cannot be sure that you are in game mode while you receive the message (for example if the player opens the console).
BTW, since we clearly stated that a mode is just a way to grab input after @oyvindln's PR about console (and thus, several modes can live together), we might need to implement some kind of machine state for modes to handle such cases and avoid the update data flags.

ATM, I've set very basic dependencies for the researches. As we said, they are hardcoded because heavily linked to the gui. If you want to change them, they are built in ConfigManager::loadResearches.

Bertram25:
@hwoarangmy Hey,

First of all, sorry for lagging behind you in term of development, I first wanted to finish something and I ran out of time much more quickly than anticipated. 

Thanks for the explanation. This looks simple enough to plug-in. :) I'm still on it but I'll first add a basic looking research tree first and try to update it when the window is being shown only (and in game mode.).

> If you want to change them, they are built in ConfigManager::loadResearches.

Yeah, sure. Looks great and easy to read. :D 

> BTW, since we clearly stated that a mode is just a way to grab input after @oyvindln's PR about console (and thus, several modes can live together), we might need to implement some kind of machine state for modes to handle such cases and avoid the update data flags.

I guess we want to avoid sending data to the wrong mode, but Is it that bad if the game mode still gets updated while the console is shown, btw?
Maybe have you a more precise example in mind?

Anyway thanks for taking care of that so quickly. :)
I wondered whether the next step is to add a few chat message when researches are done and later lost but i can handle as it's very easy to add.

Bertram25:
To testers: Library room absorbtion will also have to be tested carefully.
I'll test asap, and once this is merged, I'll have a go at adding a simple research tree window that will permit us to go on with this until I've finished the fancier version.

hwoarangmy:
> Is it that bad if the game mode still gets updated while the console is shown, btw?

IMHO, not at all. But AFAIK, it's not how it is done right now. That's why I said that we should decide what a mode is and how it should be used.
Back then, we said that it was only an input grabber (thus, modes can live together). In this case, IMHO, we should notify events on all active modes, call periodic functions even if the mode has not the focus...
At best, the idea would be to communicate with modes only through the mode manager (removing ModeManager::getCurrentMode) and replacing ModeManager::getCurrentModeType by ModeManager::isModeActive(ModeType type)

Bertram25:
Sounds saner, indeed. @oyvindln Have you got anything against that?

hwoarangmy:
> To testers: Library room absorbtion will also have to be tested carefully.

Yes. Code is the same as the Forge and it has been changed in this PR so it would be nice to test both.

> I'll test asap, and once this is merged, I'll have a go at adding a simple research tree window that will permit us to go on with this until I've finished the fancier version.

ATM, the research logic is functional (of course, a player cannot change the research tree since there is no gui) even if hardcoded. You don't have to go for a simple version if it takes time.

hwoarangmy:
> @oyvindln Have you got anything against that?

IIRC, that would imply some changes in the console, too.

Bertram25:
> You don't have to go for a simple version if it takes time.

Believe me, art takes time, especially for more code guys like me, much more than making a simple layout, so I'd rather bring the layout first if you haven't anything specific against it. :)

hwoarangmy:
> so I'd rather bring the layout first if you haven't anything specific against it. :)

Of course, not. It's just that I know how messy it can be, sometimes, to deal with CEGUI and if you can avoid unnecessary fight, it won't be a waste ;-)
But if you have something almost ready, don't hesitate to release it ^^

akien-mga:
Quite some work done here, as usual :) I'll test asap.

Bertram25:
Yeah, impressive. :) I couldn't test yesterday, but I will have a go tonight if possible.

hwoarangmy:
You can go on. I will not add commits to this PR (maybe for the missing "per turn" but not for something important).

Bertram25:
> You can go on. I will not add commits to this PR (maybe for the missing "per turn" but not for something important).

Cool :)

Bertram25:
Btw, "just" 7 issues fixed with this PR. Congrats. ;)

hwoarangmy:
> Btw, "just" 7 issues fixed with this PR

7 small issues ;-)

Bertram25:
- The research is working fine (even if done fast but it's a good start :])
- Forge & Library are absorbtion are working fine.
- The library logic is good. :)
- Wyverns do work in forge and lib again.
- The beds now disappear well, even if from corpses.
- Names in editor ok.
- I tested insta-heal by changing the heal while sleeping. It's working.

IMHO, good to be merged. :)

akien-mga:
I tested it briefly, but couldn't find how to configure the research list. Is it because the gui part is missing, so currently it can only be tested via code tweaks? :)

Bertram25:
Yep, the order is predetermined. Adding the Gui is the next step.

hwoarangmy:
> I tested it briefly, but couldn't find how to configure the research list. Is it because the gui part is missing, so currently it can only be tested via code tweaks? :)

Yes. With this PR, the research tree is hardcoded. If you want the gui to come sooner, the easier is to send as many mails/messages to @Bertram25 as you can.

Bertram25:
Eh eh. Please also send copies to Danimal for the workshop :)

hwoarangmy:
> Eh eh. Please also send copies to Danimal for the workshop :)

And for the kobold replacement ;-)

oyvindln:
> At best, the idea would be to communicate with modes only through the mode manager (removing ModeManager::getCurrentMode) and replacing ModeManager::getCurrentModeType by ModeManager::isModeActive(ModeType type)

That sounds reasonable. I think the original idea was to have multiple modes being active at the same time (which is why they are put in a vector), but currently only the top mode (or console) receives any input. I also think the "modes" responsibility should be to be glue between input and game functionality, which is mostly what they are now, though it may make sense to have a "game" class or similar contained in the mode rather than having gamemap/client/server classes living outside that are manually cleared/updated as they are now. Though, this is an issue for later. I'm currently doing some cleanup in the modes/gui which I want to finish first.

Bertram25:
> though it may make sense to have a "game" class or similar contained in the mode rather than having gamemap/client/server classes living outside that are manually cleared/updated as they are now. 

If it can make things easier to read/follow, I'm all for it. :)

> I'm currently doing some cleanup in the modes/gui which I want to finish first.

Are some layout files touched by the update?

Bertram25:
A quick question while we're at it, The tabs at screen bottom are now taking a lot of useless space.
Do you think it would be a good idea to reduce it on the right side, and why not move the minimap bottom right for instance?

Bertram25:
@akien-mga @oyvindln Can we merge this one btw?

oyvindln:
> Are some layout files touched by the update?

Nope

> @akien-mga @oyvindln Can we merge this one btw?

I haven't tested it but it's okay for me if it works fine for you.

> Do you think it would be a good idea to reduce it on the right side, and why not move the minimap bottom right for instance?

That may work yes.

hwoarangmy:
> I think the original idea was to have multiple modes being active at the same time (which is why they are put in a vector)

When I read the code, at first, I had guessed that modes were used for taking the whole screen (because only the last mode was accessible). I thought the vector was there to remember which mode we came from and allow going back (not a bad idea, especially for menus). But the console was clearly not adapted for that. I guess it's because it was not clear for everybody how modes were supposed to be. Now that we have decided, I guess the machine state is the most fitted architecture since we all agree that modes should work together.

> it may make sense to have a "game" class or similar contained in the mode rather than having gamemap/client/server classes living outside that are manually cleared/updated as they are now

Yes, now that modes are well defined, we can think about pluging at least the gamemap on it. Concerning the Server, I always had in mind that we should get things ready to have the server completely independent from the client (that's one of the reasons why it has its own thread and gamemap). More precisely, to be able to launch the server in a special mode, without client. I know it's not completely true as for now since the seat configuration requires one of the clients to run as a server. But a change to that might be pretty easy and if we want, we can achieve that easily.
Because OD is a spare time project, I think it is a good time to do an architecture as advanced as we can (since there is no due date) and that's definitely the way I would have gone if I was working on a game (who would have believed it ?)
In any case merging those will probably require a lot of work (and testing). It would have to be pretty well thought to improve pretty much the game since there would have high probably of regressions.

> I'm currently doing some cleanup in the modes/gui which I want to finish first.

I guess there are many things that can be done there ^^

> Do you think it would be a good idea to reduce it on the right side, and why not move the minimap bottom right for instance?

Well, you are the gui master so, if you feel that it would be better, why not ? Just remember  that we will probably have more rooms so we have to find a way to allow that (that could be adding an "advanced rooms" tab ^^ )

Bertram25:
> In any case merging those will probably require a lot of work (and testing). It would have to be pretty well thought to improve pretty much the game since there would have high probably of regressions.

Just to be clear, I, for one, have no merging in mind, especially as we'll one day want a dedicated server mode, as you stated above. Please, remember good objects are reusable objects.
I rather thought the game class as some kind of wrapper or the needed components.

> Well, you are the gui master so, if you feel that it would be better, why not ? Just remember that we will probably have more rooms so we have to find a way to allow that (that could be adding an "advanced rooms" tab ^^ )

Yeah, it could be, or we could reduce the buttons slightly and have two rows. :)
Plus, I'd like to make a newly researched "button" appear next to the previous one one day instead of fixing their location.

hwoarangmy:
> Plus, I'd like to make a newly researched "button" appear next to the previous one one day instead of fixing their location.

That will probably be needed when we will have enough rooms/traps/spells to have truly different paths of evil

PR #513, opened 2015-02-21
Text overay over creature head + support move camera to battles

hwoarangmy:
To test the text creature overlay, just press 'n' in game mode. Note that the background color is a material so it should not be too hard to display the skull flag ^^

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -308,6 +308,7 @@ set(OD_SOURCEFILES
     ${SRC}/network/ServerNotification.cpp
 
     ${SRC}/render/Gui.cpp
+    ${SRC}/render/MovableTextOverlay.cpp
     ${SRC}/render/ODFrameListener.cpp
     ${SRC}/render/RenderManager.cpp
     ${SRC}/render/TextRenderer.cpp
--- a/materials/scripts/CreatureOverlay.material
+++ b/materials/scripts/CreatureOverlay.material
@@ -1,0 +1,18 @@
+material CreatureOverlay
+{
+   technique
+   {
+      pass
+      {
+         scene_blend alpha_blend
+         lighting off
+         depth_write off
+         texture_unit
+         {
+            colour_op_ex source1 src_manual src_current 1 0 0
+            alpha_op_ex source1 src_manual src_current 0.4
+         }
+      }
+ 
+   }
+}
--- a/source/entities/Building.cpp
+++ b/source/entities/Building.cpp
@@ -250,7 +250,7 @@ double Building::takeDamage(GameEntity* attacker, double physicalDamage, double 
         return damageDone;
 
     // Tells the server game map the player is under attack.
-    gameMap->playerIsFighting(player);
+    gameMap->playerIsFighting(player, tileTakingDamage);
 
     return damageDone;
 }
--- a/source/entities/Creature.cpp
+++ b/source/entities/Creature.cpp
@@ -36,6 +36,7 @@
 #include "network/ODServer.h"
 #include "network/ServerNotification.h"
 
+#include "render/MovableTextOverlay.h"
 #include "render/RenderManager.h"
 
 #include "rooms/RoomCrypt.h"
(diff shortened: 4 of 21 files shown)


Danimal696:
Skull Flags are coming!!  :D

Bertram25:
Woohoo need to test those text rectangle with overlapping handling :D

Bertram25:
Tested it, the space key works like a charm. I guess we'll need a chat message telling we're under attack soon enough. :)
The creature text overlay are great to see in game, really :D 
I've noticed a few things, just tell me what you think about it:
- The overlays seem to overlap, despite the addition of the Rectangle Handler class.
  (I didn't check the code well yet, but I guess there is something off here.)
- The overlay offset from the creature is changed by the camera movement. It means they go to the right when you're moving right, ... Just push N and move the camera around some creatures and you'll get what I mean.
- Even is those overlays are meant to be basic at first, I would still make them a bit more useful by displaying the HP / MaxHP instead of the creature name. I mean we're evil, we don't get attached to our creatures so we don't want to know their names, right? ;)
- I would use the background color of the seat instead of always red, even if it's a temporary thing. :)

It's great you gave a nudge about this feature. :)

hwoarangmy:
> I guess we'll need a chat message telling we're under attack soon enough

Fair, I will add that

> The overlays seem to overlap, despite the addition of the Rectangle Handler class

If you mean that they overlap when the creatures are close, that's the expected behaviour. I've not checked if there is already another overlay before displaying a new one (this can become tricky very easily).

> The overlay offset from the creature is changed by the camera movement. It means they go to the right when you're moving right

If you mean during camera moves, I'm not sure that something we can avoid. We use Ogre to get the text position and to convert it to screen position. But from what I've seen, once the camera stops, the overlays are correctly displayed. Do you confirm ?

> Even is those overlays are meant to be basic at first, I would still make them a bit more useful by displaying the HP / MaxHP instead of the creature name

IMHO, this will be done when we will display the skull flag

> I would use the background color of the seat instead of always red

To be honest, I didn't do that (nor the skull flag) because I was tired of fighting with Ogre. Moreover, you and @oyvindln know better Ogre than I do and you might have a better idea of the best way to change the material to adapt to the seat color/hp (shall we create different materials? Or change them at runtime as we do for tiles?)

Bertram25:
> To be honest, I didn't do that (nor the skull flag) because I was tired of fighting with Ogre. Moreover, you and @oyvindln know better Ogre than I do and you might have a better idea of the best way to change the material to adapt to the seat color/hp (shall we create different materials? Or change them at runtime as we do for tiles?)

Eh eh. I completely understand if Ogre tried to get the best part out of you. ;)
Please don't take the things I point out something you'll absolutely have to fix yourself. I'm just pointing out stuff while I see it, as usual. :) Plus, graphical stuff is overwhelming sometimes.
For the record, I have seen no actual flaws in the code and in the game so far, but I'll have a closer look today.

> If you mean that they overlap when the creatures are close, that's the expected behaviour. I've not checked if there is already another overlay before displaying a new one (this can become tricky very easily).

Yep. After thinking about this, I guess that once we have skull flags, we'll be able to decide on whether we actually want to avoid overlapping or not. (Offsetting creatures when they are on the same tiles will also help already.)

> If you mean during camera moves, I'm not sure that something we can avoid. We use Ogre to get the text position and to convert it to screen position. But from what I've seen, once the camera stops, the overlays are correctly displayed. Do you confirm ?

Yes, I confirm. :) What's interesting to say is that the rectangle offset change is linked to the camera direction. Thus, I'm sure we can fix the offset change when moving based on such data. It can be done later also.

> IMHO, this will be done when we will display the skull flag

Ok. I just wondered whether you were done for now (since the PR is tricky enough) and wanted to leave this one as is for a while. Explaining my notice. But you're right, using an icon is way better. :)

> Or change them at runtime as we do for tiles?)

I'd be for changing them at runtime since the team color is a parameter and it's better that way.
Again, we can deal with each point step by step and by anyone. :)

Bertram25:
Is ok to be merged IMHO.

akien-mga:
Testing now, and I'll merge afterwards :)

akien-mga:
Everything works great :-D

PR #514, opened 2015-02-21
Gui code cleanups + fixed replay checkbox

oyvindln:


hwoarangmy:
arg, I've done the exact opposite in my PR. I've even removed addEventConnection as I had seen that it was not very used. Now, I guess it was because you were preparing to use it for everything. Seems like I had a nice intuition here ^^
Concerning your PR, usually, I prefer the object vision of each mode taking care of itself (and, thus, subscribing to the buttons they want themselves) like you did here. I will try to remove my last commit and wait for your PR to be integrated ^^

hwoarangmy:
I've removed the concerned commit. I will see what I can do once your PR is integrated. BTW, are you going to add commits here ? Or can we already test ?

oyvindln:
> Concerning your PR, usually, I prefer the object vision of each mode taking care of itself (and, thus, subscribing to the buttons they want themselves) like you did here. I

Yeah that was the main point of the PR. I didn't move the gamemode/editormode events yet to avoid having too much merge conflicts from your previous PR. I'll wait with doing that until your PR is merged I think, so feel free to test.

I also fixed an issue with the replay checkbox. For whatever reason it was returning "true" instead of "True" for me, so I changed it to get the bool directly by casting instead to avoid naming issues.

hwoarangmy:
I've tested and it seems ok to me. I will have a look closer to the commits asap

akien-mga:
I've checked that it builds and runs fine. I'll let you code gurus decide if you want to merge or if it needs more work :)

hwoarangmy:
seems like there is a merge conflict :-(

hwoarangmy:
> if it needs more work

@oyvindln already said he will remove the included gamemap and rendermanager from the odframelistener header.

Bertram25:
Yes, I'd request that removal happens before the merge if not too dirty to do.

oyvindln:
Yeah I will sort it soon, I just haven't been able to work much the last few days.

Bertram25:
> Yeah I will sort it soon, I just haven't been able to work much the last few days.

Take the time needed.
Btw, I reverted the changes you made on the code guidelines, as there was stuff that was right among the things you removed.
Please discuss the changes before doing them.

oyvindln:
Rebased and fixed requested issues.
I also added a workaround that seems  to fix the rtss lighting issues on my part, but I don't know it the problem I was having was the same as what you are having.

hwoarangmy:
@oyvindln I confirm you have fixed the tile problem (at least for me, probably for all) ^^
Congrats. Now, I realize what you said about normals. The tiles are really better and much more detailed. It even seems less laggy in debug mode ^^

But with your PR, the yellow coloured selected tiles stay yellow when they are unselected. We can see in the minimap that they are not selected but the color don't change. Same applies for dirt tiles. Funnily, gold tiles seem to work normally (getting back to normal when unselected).

Bertram25:
> put rm and gamemap on heap to cut down on header dependencies

Fair enough for me. I'll test all that asap. :)

hwoarangmy:
Thanks for the fixes. The code looks ok to me. I will test it now

Bertram25:
There is a bug. Gold tiles become dirt tiles when being refreshed.
Just ask a few workers to go and dig a patch of gold and you'll see what I mean.

Apart from that, the change is visually impressive. Especially on walls and on claimed tiles. \O/
I even earned a few FPS (4 or 5, I'd say). Congrats for the work done @oyvindln :D

hwoarangmy:
I confirm the bug I've seen previously is fixed. But I reproduce the bug @Bertram25 have seen.
Tiles are really better like that. Once this PR is merged, I will try to remove rtshaders to see the difference ^^

Bertram25:
Yeah, it is really coming along. Woohoo :D

hwoarangmy:
BTW, I think it also fixes the light bug issues seen by Vandar

Bertram25:
:) Let's ask him on next binary release. Real good stuff tonight. It's Christmas. :D

Bertram25:
I'll retest now.

akien-mga:
I've built the current PR and tested it, it's great :-)
This game is really starting to look good :-D

About the gold issue, I confirm that 425a75d fixes it. The only detail is that dug gold tiles are now using the "tile fullness 0" gold texture, but we decided earlier to stop using it and just display plain dirt ground (the "there was gold here previously" texture is not that great looking).

Bertram25:
> The only detail is that dug gold tiles are now using the "tile fullness 0" gold texture, but we decided earlier to stop using it and just display plain dirt ground (the "there was gold here previously" texture is not that great looking).

Ah yes, gold with 0 fullness should be dirt for now, until we have a better tile.
EDIT: Can be done after merge though, as it's minor.

Bertram25:
Looks very good. There is something funny looking with Lava and the FOW, but I don't know whether it's a regression. The lava animation gets unsynchronized between the visible lava and the seen once but not currently watched lava tiles.
As it gets updated, more animation sync gets funky.

oyvindln:
I don't think FOW was enabled on lava/water tiles before, I enabled it now but if it's broken maybe only the dirt part of the tiles ought to be darkened, or it could just be disabled again for now. It's probably due to the animation, and creating a duplicate material might mess up the timing.

Bertram25:
> I don't think FOW was enabled on lava/water tiles before, I enabled it now but if it's broken maybe only the dirt part of the tiles ought to be darkened.

I agree and maybe that's what was decided before even if I can't recall.

hwoarangmy:
I couldn't test but since @Bertram25 and @akien-mga say that the gold tiles bug is gone, I think we can merge.
Concerning the unsynchronized lava/water tiles, if we really feel it is a bug, we can open an issue and fix that later.

Merging

Bertram25:
> Concerning the unsynchronized lava/water tiles, if we really feel it is a bug, we can open an issue and fix that later.

Yes, it is one as I now recall saying they should always be set as active/visible to avoid such problems. Bu yeah, let's open an issue and fix that separately. :)
Congrats for the job done and the finding about lighting @oyvindln 

I'd still wonder what made you discover that. I'm eager to hear the explanation. :)

Bertram25:
I opened #522

PR #515, opened 2015-02-22
Normal and fixes for training room objects and library

Danimal696:


akien-mga:
It looks great :-)

Danimal696:
Put a normal in your game, you will never repent ;-)

Danimal696:
With this the rooms and objects have fixed materials and normals added;
the only exception are the knigth statues and coffin from Crypt that
didnt benefit at all from normals.
From now on its creatures that are final and their beds

Bertram25:
Wow, looking really great now. Especially the doll in the training room, the library objects, too. :D 
Fine to me to be merged.

PR #519, opened 2015-02-24
Skull flag over creatures + code cleaning

hwoarangmy:
Skull flag is displayed while "alt" is pressed over every visible creature.
It is also displayed if a creature is hovered by the mouse cursor while no action is selected (like building room or destroying trap)


akien-mga:
Wow, this looks awesome in game :-D

Bertram25:
Wow, nice cleanup (especially for the gamemap) and I don't even speak about the feature. :D 
I'll test asap, but the code looks overall fine, my usual small comments apart.

hwoarangmy:
> nice cleanup (especially for the gamemap)

That was allowed by removing AS because many unused (or not well fitting) functions were binded. That's why I think that's the good time to clean that and use only the correct ones when we will have a go at AS.

hwoarangmy:
Please don't merge this until @oyvindln PR is merged or it will break again his PR

Bertram25:
NP. But I'll test anyway. I just can't wait to see this. :D

Bertram25:
I tested it and it's awesome. It really gives an identity to the game and to the creatures somehow. It's great :D 
I don't know whether you have seen but when playing and often during battles when you grab and drop a lot, when you press alt to see how your creatures are doing, there are sometimes ghost flags shown on screen linked to no creatures. (Full and 3/4 health from what I saw). Those flags disappear after a few seconds though.
I'll try to give you a screenshot asap.

Bertram25:
I couldn't reproduce and anyway, it shouldn't prevent the merge. Once I'll be able to reproduce, I'll open an issue.

Bertram25:
This is still mergeable and IMHO it is fine to be merged. :D

Bertram25:
Ok, since everyone has pushed on the merge button except me, I'll do it, for once. :P

Bertram25:
Btw, @hwoarangmy would you mind if I (later) add another "skull flag" icon for the Hero faction? It would become the "Knight flag" or something I suppose.
Or do you think it would be better to be able to choose a flag per seat?

hwoarangmy:
> Or do you think it would be better to be able to choose a flag per seat?

That's what was said in the original discussion. I don't really mind. But if we find a way to dynamically change the picture (which should not be too hard I guess), it will not be harder to use per seat flags.

In any case, do as you want, there will always be time to change that later if we feel like it.

Bertram25:
> In any case, do as you want, there will always be time to change that later if we feel like it.

Ok :) The feature as it is now is already a dream come true, but one day, I may work on that. ;)

PR #521, opened 2015-02-26
Moved the mingled icons and skin images back to their respective image file.

### LINKED ISSUE #512: Icons should all be in the OpenDungeonsIcons imageset
At the moment, the tiles icons and soon the Light editor icon will be in the skin imageset.
As there is an imageset specifically for icons, we should move those ones there.

The icon imageset may have to be resized or split into two if needed.

If nobody takes it, I'll do it once the Light icon is added. :)

Bertram25:
Fixes #512

I also remade the Gold, Territory and Mana icons.
I also added other icons in preparation for the PR after that one. ;)


akien-mga:
This is straight-forward, I'll merge. I love the new icons :-D

PR #523, opened 2015-02-26
Reordered the Rooms buttons to fit more the current game state.

Bertram25:
Just a small quick-fix/improvement.


akien-mga:
Works fine :-)

Issue #517, opened 2015-02-23
Add the research tree GUI


Bertram25:
Following #411 , the Research tree GUI and adjustments to the tab button and GUI is needed.
Here is the research tree first layout:
http://forum.freegamedev.net/viewtopic.php?p=62549#p62549
(The layout file is also available there.)
![research-tree-basic](https://cloud.githubusercontent.com/assets/765917/6252062/dcae03a0-b797-11e4-9502-66fa1ce0c650.png)

**How to adapt the OD/ImageButton:**
I'll be honest, I don't actually know yet. I guess a check mark or something should appear on the bottom right of the researches done, and something like a cog should appear on pending ones. The unavailable ones should simply appear disabled, even if you should be able to click on them to add them to the pending queue.

For that, I'll add new ImagerySection in ImageButton or maybe I'll duplicate that one and make something like a ResearchButton look'n'feel class:
https://github.com/OpenDungeons/OpenDungeons/blob/development/gui/OpenDungeons.looknfeel#L4091

Then, I'll need to have a look at how I can change the layers used on the different button states here depending on the research state: https://github.com/OpenDungeons/OpenDungeons/blob/development/gui/OpenDungeons.looknfeel#L4130

In the meanwhile, only available researches will be enabled, the rest (What's done and unreachable) will be disabled until I've figured out how to dynamically change the layers in a button state.

Useful link: http://static.cegui.org.uk/docs/0.8/classCEGUI_1_1StateImagery.html

Bertram25:
I had an idea. I can add an empty layer for the standard button behaviour and ask to fill/empty them with an image from the code. :D It will be much simpler and can be used to show the pending/done states.

I still do think I'll overload the OD/ImageButton look'n'feel widget into a ResearchButton, but I'll add a label (text) support for both. Might be useful...

hwoarangmy:
It could also be nice to display the research order somehow and to allow to edit it.
For that, you could display a number (1, 2, ...) on the researches telling which one is the next to be done. 1 would be the current one, 2 the next, ... But editing the order would not be straight forward.
Or you could add a "Pending researches" panel in the research tree and add the researches icons there when they are selected. Editing the order would be easier but it would need to add a new panel...

Bertram25:
> It could also be nice to display the research order somehow and to allow to edit it.
> For that, you could display a number (1, 2, ...) on the researches telling which one is the next to be done. 

True, the image/research buttons will definitely need label support.

[...]

PR #525, opened 2015-02-26
GUI update.

Bertram25:
This is in preparation of the addition of the research tree and the settings menu.

This part does only change the GUI part, no cpp was changed.
The logic part will be part of another PR.

Part of #3 
Part of #517


Bertram25:
I'll add code to make the options menu open instead of the settings window. I'll add it there if I have the time but feel free to merge if it's ok for you guys.

akien-mga:
Hey, is that OpenDungeons 0.6.0 already? This GUI looks gorgeous :-D

Bertram25:
Thanks :)

akien-mga:
Just one thing: the options window can be shown with F10, but it can't be close with F10. It would be nice to have this implemented, as it is done currently for the help (F1) and objectives (F3).
When you add the research screen, the F4 key should also have a toggle behaviour IMO.

Bertram25:
Yeah, I'm on it at this very moment in fact. :)

Bertram25:
I added the Game Options Menu. The available buttons should be working, while the settings window should still not be functional.
I'll have a go at doing the same for the editor.
Then, I'll start working on the research tree. :)

Bertram25:
Ah, of course, if you have feed-back concerning usability or anything, just tell.  Let's adapt until it feels right.

hwoarangmy:
I like the new GUI. The next release will really rock ^^

hwoarangmy:
I've just tested and it seems that in editor mode, the option menu doesn't popup (by pressing F10 or the button)

Bertram25:
> I've just tested and it seems that in editor mode, the option menu doesn't popup (by pressing F10 or the button)

Yeah, that's almost done.

Bertram25:
For the record, I'd like to add stuff little by little, and not hide the buttons if possible. If we reach a release while I'm not finished, I'll hide them back at that moment.

akien-mga:
Nice! The only issue (or better said, "yet unimplemented feature that you are probably aware of, but I tell it nonetheless") I found is that the "Quit game" button in the options menu does not bring up the "quit" dialog.

Bertram25:
>  I found is that the "Quit game" button in the options menu does not bring up the "quit" dialog.

ah ? It should, I'll have a look.

Bertram25:
I just tested, the exit pop-up does show in the game mode. Did you speak about the editor mode?

akien-mga:
> I just tested, the exit pop-up does show in the game mode. Did you speak about the editor mode?

Indeed, I just tested again it works fine. I was speaking about the game mode, but either I'm misclicked, or I hit some corner case where the GUI is not responding properly when you've clicked on all buttons in random order... I couldn't reproduce any issue with the Quit game button, so let's assume it's a PEBKAC :)

Bertram25:
> Indeed, I just tested again it works fine. I was speaking about the game mode, but either I'm misclicked, or I hit some corner case where the GUI is not responding properly when you've clicked on all buttons in random order... I couldn't reproduce any issue with the Quit game button, so let's assume it's a PEBKAC :)

Now you may have catched something here. I've noticed several times the quit window being show at game start for no reasons. So there is definitely something but it's not linked to this PR. Maybe in the popupExit() function or the way it is called?

Bertram25:
Allright. Evil plan part II is online. If you're fine with this PR, feel free to merge. :)

Bertram25:
> I like the new GUI. The next release will really rock ^^

Thanks a lot :) Feel free to point out anything, even it seems obvious just as @akien-mga did.
We're one step closer from having a complete GUI. :)
Next step: The research tree.

akien-mga:
Great! At least with have a visible "Save level" button, I guess early modders will have a hard time finding the F8 key on 0.4.9 ;-)
For me this is really good stuff, I merge.

PR #527, opened 2015-02-27
Implemented save games

hwoarangmy:
For now, there are still some limitations:
- Room activity is not saved (points in library or forge)
- Carried entities will be at their original position after saving instead of being near the worker carrying them
- Beds are not saved (all the creatures will go to the dormitory to claim a new bed)
- Vision is not kept on persistent objects (traps/rooms)

Points 1,2 and 3 can easily be fixed.
Point 4 will not be easy to fix because the gamemap can be different for every player. Moreover, the server do not save the state for each client. For example, in a game with 2 players and feral traps, if player 1 sees a trap and looses vision while player 2 destroys the trap, player 1 should see it until he gets vision (and, on server side, the trap will be destroyed).

Of course, we can decide to keep all relevant information on server side. But concerning the map level, that would mean to save each tile state for each player...

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -292,6 +292,7 @@ set(OD_SOURCEFILES
     ${SRC}/modes/MenuMode.cpp
     ${SRC}/modes/MenuModeConfigureSeats.cpp
     ${SRC}/modes/MenuModeEditor.cpp
+    ${SRC}/modes/MenuModeLoad.cpp
     ${SRC}/modes/MenuModeMultiplayerClient.cpp
     ${SRC}/modes/MenuModeMultiplayerServer.cpp
     ${SRC}/modes/MenuModeReplay.cpp
--- a/gui/GameOptionsMenu.layout
+++ b/gui/GameOptionsMenu.layout
@@ -20,7 +20,6 @@
             <Property name="Area" value="{{0.5,-80},{0,136},{0.5,80},{0,169}}" />
             <Property name="IconImage" value="OpenDungeonsIcons/SaveIcon" />
             <Property name="Text" value="Save Game (F5)" />
-            <Property name="Disabled" value="True" />
             <Property name="TooltipText" value="Not implemented yet" />
         </Window>
         <Window type="OD/IconButton" name="LoadGameButton" >
--- a/gui/OpenDungeonsMainMenu.layout
+++ b/gui/OpenDungeonsMainMenu.layout
@@ -27,9 +27,9 @@
             <Property name="Text" value="Skirmish" />
             <Property name="AlwaysOnTop" value="True" />
         </Window>
-        <Window type="OD/MainMenuButton" name="StartReplayButton" >
+        <Window type="OD/MainMenuButton" name="LoadGameButton" >
             <Property name="Area" value="{{0.5,-310},{0.5,100},{0.5,-170},{0.5,180}}" />
-            <Property name="Text" value="Replay" />
+            <Property name="Text" value="Load" />
             <Property name="AlwaysOnTop" value="True" />
         </Window>
         <Window type="OD/MainMenuButton" name="StartMultiplayerClientButton" >
@@ -43,8 +43,13 @@
             <Property name="AlwaysOnTop" value="True" />
         </Window>
         <Window type="OD/MainMenuButton" name="MapEditorButton" >
-            <Property name="Area" value="{{0.5,10},{0.5,40},{0.5,150},{0.5,120}}" />
+            <Property name="Area" value="{{0.5,10},{0.5,-10},{0.5,150},{0.5,70}}" />
             <Property name="Text" value="Map Editor" />
+            <Property name="AlwaysOnTop" value="True" />
+        </Window>
+        <Window type="OD/MainMenuButton" name="StartReplayButton" >
+            <Property name="Area" value="{{0.5,10},{0.5,100},{0.5,150},{0.5,180}}" />
+            <Property name="Text" value="Replay" />
             <Property name="AlwaysOnTop" value="True" />
         </Window>
         <Window type="OD/MainMenuButton" name="QuitButton" >
--- a/gui/OpenDungeonsMenuLoad.layout
+++ b/gui/OpenDungeonsMenuLoad.layout
@@ -1,0 +1,65 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<GUILayout version="4" >
+    <Window type="OD/StaticImage" name="Background" >
+        <Property name="Area" value="{{0,0},{0,0},{1,2},{1,0}}" />
+        <Property name="Image" value="ODMainMenuBackground/MainMenuBackground" />
+        <Property name="MaxSize" value="{{1,0},{1,0}}" />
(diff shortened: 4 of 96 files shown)


hwoarangmy:
points 1 and 2 are fixed

akien-mga:
Nice! I just tested the feature and it works great! Some points will need polishing but it's already an impressive achievement.

Small bugs/enhancement suggestions that I found so far:
- We're using F8 as save key both in the editor and in the game mode, but in @Bertram25's GUI the save key is F5, and F8 would be the load key, so we'll have to homogenise this.
- Some carriable stuff may be lost when saving at the wrong moment, e.g. on Scream in the Dark, if you save while the workers are digging the nearest patch of gold, part of the gold might end up disappearing (you will get e.g. ~1500 gold even though the whole should be ~3000).
- Displaying the date + time for each saved game in the load menu would be a nice plus :)

akien-mga:
Man, that game! The skull flags, the new UI, the save feature... It's getting awesome :-D

Bertram25:
> We're using F8 as save key both in the editor and in the game mode, but in @Bertram25's GUI the save key is F5, and F8 would be the load key, so we'll have to homogenise this.

Ah true. Well, I used F5 for (quick) saves since it's a standard I have seen in many games. There can be counter examples, and I don't mind about the actual key used. Just tell me. :)

akien-mga:
> Ah true. Well, I used F5 for (quick) saves since it's a standard I have seen in many games. There can be counter examples, and I don't mind about the actual key used. Just tell me. :)

I like F5 too, so I'd propose to change the current save key for editor and game modes.

hwoarangmy:
> Ah true. Well, I used F5 for (quick) saves since it's a standard I have seen in many games

I've used my copy/paste skills and I didn't notice the keys didn't match ^^

Bertram25:
Still reviewing code. There is a lot. :)

Bertram25:
> I've used my copy/paste skills and I didn't notice the keys didn't match ^^

Well, I introduced the mismatch, silly me. XD

Later, I'd request the saves are stored in sub-folders according to the game type: single-player, multiplayer or later campaign.

> Point 4 will not be easy to fix because the gamemap can be different for every player. Moreover, the server do not save the state for each client. For example, in a game with 2 players and feral traps, if player 1 sees a trap and looses vision while player 2 destroys the trap, player 1 should see it until he gets vision (and, on server side, the trap will be destroyed).
> Of course, we can decide to keep all relevant information on server side. But concerning the map level, that would mean to save each tile state for each player...

I'll be honest, as a player, I don't care about the room tile last vision. This is only relevant for traps and strategical entities such as crafted spells, and dropped gold. I'd be for not storing the last vision about rooms for every players but why not store it for the aforementioned entities.
Feel free to correct me if you see something wrong, of course.

That's some great and hard work. I'll be testing now.

hwoarangmy:
> Later, I'd request the saves are stored in sub-folders according to the game type: single-player, multiplayer or later campaign.

I would be for changing that when we will have the campaign because when we will be there, we will have a clearer vision of how the saves should be splitted. To be honest, ATM, I don't really know.

> I'll be honest, as a player, I don't care about the room tile last vision. This is only relevant for traps and strategical entities such as crafted spells, and dropped gold. I'd be for not storing the last vision about rooms for every players but why not store it for the aforementioned entities.

Well, if we do the work for traps, it won't be very different for rooms anyway. As a player, it would feel strange to have a different map between when I saved the game and when I load it. And ATM, there are huge differences. Rooms are not the most important. The biggest problem is about tiles. Since only gold and rock tiles are notified at map load, that means that you will see the gold dug by an enemy as dirt tiles (because they aren't gold tiles in the saved game) on tiles you never had vision. Moreover, tiles you got vision on and you have seen as dirt fullness = 0 will be dirt fullness = 100 after map load until you get vision.
Just make some tries, it feels strange. I was wondering if we should somehow save data on clients about the client gamemap or if we should save everything on the server side. But the first solution would break the logic that the server should be the only one to load a map. SO I guess the second one is better...

hwoarangmy:
beds are now saved.
Only the hardest part is left: tiles ^^

hwoarangmy:
@Bertram25 @oyvindln Since now, modes bind themselves to GUI events and since the GUI is defined in text files, don't you think it would make sense to remove all the GUI components from the Gui class and let the modes define it themselves ? I'm thinking of Gui::EDM_BUTTON_BACK, Gui::EDM_BUTTON_LAUNCH, Gui::SKM_BUTTON_LAUNCH, ...

hwoarangmy:
> Some carriable stuff may be lost when saving at the wrong moment, e.g. on Scream in the Dark, if you save while the workers are digging the nearest patch of gold, part of the gold might end up disappearing (you will get e.g. ~1500 gold even though the whole should be ~3000).

Are you sure of that ? If there is no bug, carried gold is saved, just like treasuries (and their values). Have you checked by dropping your kobolds on a treasury tile that they were not carrying gold ? Are you sure some treasury was not hidden by FOW ?
I've tried a few times and couldn't reproduce.

oyvindln:
> Since now, modes bind themselves to GUI events and since the GUI is defined in text files, don't you think it would make sense to remove all the GUI components from the Gui class and let the modes define it themselves ?

Yep

Bertram25:
> Since now, modes bind themselves to GUI events and since the GUI is defined in text files, don't you think it would make sense to remove all the GUI components from the Gui class and let the modes define it themselves ?

I'm sure @oyvindln had planned it already. ;)

> Just make some tries, it feels strange. I was wondering if we should somehow save data on clients about the client gamemap or if we should save everything on the server side. But the first solution would break the logic that the server should be the only one to load a map. SO I guess the second one is better...

This means we'd have a (human) player view part in the data, then? I tried a few things. It makes sense, (unfortunately). Ok, let's save the human players view. It will give us an excuse to have more professional loading times (aka longer). ;)

Bertram25:
It seems you still didn't link the option save button with the save functionality. Wanna do it in this PR, or shall I ignore this for now?

Bertram25:
Apart from my previous question it is working fine to me. :)
As for tiles, I wondered whether you think saving marked tiles are also part of the player's data you plan to add?

hwoarangmy:
I'm going to add the links between UI and save functionality right now. Concerning tiles, it will be for later as it may be pretty complex. And this PR will probably need rebasing at each new PR submitted so I guess it is wiser to merge this one and do another one later.

hwoarangmy:
Done ^^

akien-mga:
Tested, it works great :)
I'll let one of our code reviewers do the merge if all commits are fine.

akien-mga:
@Bertram25 Feel free to merge if the PR is fine for you. I'm pretty sure @hwoarangmy already has lots of stuff committed and he's impatiently waiting to be able to run "git push" again :-p

hwoarangmy:
Well, I'm not done with the tilesets yet. But in any case, I will wait for this PR to be merged before pushing again ^^

hwoarangmy:
SInce @Bertram25 reviewed my commits except the last (which is straight forward) and @akien-mga tested it, I will push the green button myself as I have other things to push ^^

akien-mga:
Fair enough :-)

PR #528, opened 2015-02-28
Fixed #522

### LINKED ISSUE #522: Misc stuff to fix on tiles
- [x] Gold tile with 0 fullness should use the ground dirt tile until a better tile is made.
- [x] Lava and water should always use the visible format (according to the use of the FOW) so that their animation never get desync.

Bertram25:


hwoarangmy:
Straight forward and works as expected. I merge

PR #532, opened 2015-03-05
Implemented tile sets

hwoarangmy:
I will add another commit tomorrow that defined how the tiles connect to each other


source/gamemap/GameMap.h
```diff
@@ -48,10 +49,70 @@ class CreatureMood;
 class RenderedMovableEntity;
 class Room;
 class Spell;
+class TileSet;
 
 enum class GameEntityType;
 enum class FloodFillType;
 enum class RoomType;
+enum class TileVisual;
```
oyvindln: I think it would be better to put these classes in a separate file, the gameMap file is already massive.

hwoarangmy: Ok. I will see if I can move it to TileContainer as it makes sense.

hwoarangmy:
@oyvindln Since @Bertram25 seems not to be around (let's hope he is on holydays), could you please review this PR ?

oyvindln:
It seems to work fine here, though the tileset could use some tweaking, the wall texture doesn't look particularly good on the ground, and the gold texture isn't tiling properly (I think we really need a new gold texture any case.)
odscreenshot_2015-03-06_151328_0

oyvindln:
The UV mapping seems to be bad for the tile models (see the row of tiles on the left and the corner tiles):
odscreenshot_2015-03-06_153849_0

Also noticed a bug here, claiming wall tiles seems to make tiles behind them visible. I would guess this isn't from this PR though.

hwoarangmy:
> though the tileset could use some tweaking

This tileset is based on old tiles I've found in the repo while cleaning. I used it only to test several tilesets.

> Also noticed a bug here, claiming wall tiles seems to make tiles behind them visible.

That's the expected behaviour. Claimed tiles give vision on themselves and on their neighbours. Do you think it would be better to not give vision when claimed walls ?

hwoarangmy:
@oyvindln If you want to try to link different tiles, you can use the tileset to see if it increases FPS ^^

oyvindln:
> This tileset is based on old tiles I've found in the repo while cleaning. I used it only to test several tilesets.

I guess we'll have to let @Danimal696 play with it to see what he comes up with.

> That's the expected behaviour. Claimed tiles give vision on themselves and on their neighbours. Do you think it would be better to not give vision when claimed walls ?

Ah, okay. I don't know what would work best, though I don't think that was how it worked in the old DK at least, but anyway, that's not really relevant to this PR. 

I'll test the new commits.

oyvindln:
There seems to be some issue with the material on claimed tiles in the old tileset. The file says it's supposed to use the "Claimed" material, but it's using the "ClaimedWall" one instead. When I tried to put in Axe as the material it did show the axe material.
EDIT: Forget that, it is using the right material, it's just scaled so for whatever reason I thought it was the other material.

hwoarangmy:
How do you know it is using claimedwall ?

Danimal696:
Whoa, what a mess you doing in there :O 
Are you sure you are using the newest tileset?:
http://opengameart.org/content/3d-dungeon-tileset

hwoarangmy:
> Are you sure you are using the newest tileset?

Yes. I've defined 2 tilesets. One based on older tiles I'e found in the repo (the ones you can see in @oyvindln screenshot) and the actual tileset (on every other level than skirmish 2v2 test small map). That was just to make tests on using different tilesets.
FYI, I've removed the unused tiles and used as few as possible (instead of dupplicating the tiles, I've used rotations). I might have not used the expected ones everywhere. But anyway, it is pretty easy to change with the new tileset ^^

hwoarangmy:
> The file says it's supposed to use the "Claimed" material, but it's using the "ClaimedWall" one instead

I've added logs in the render manager to see if a tile was using claimedwall material and I've seen none. Could you please tell me how did you find it ?

hwoarangmy:
@Danimal696 I will upload a new binary on SF so you can test. I will let you know when I'm done on the forum.

hwoarangmy:
BTW, note that I've re-added issue #522.
Now, gold tiles are always notified as full gold tiles when starting a new game (the main reason being not to show the dug gold tiles when loading a saved game). As a side effect, that will also allow map creators to hide creatures in gold tiles. For example, that would be something pretty nice in @akien-mga map greed or might.
But because of that, gold ground tiles should be different than dirt ground tiles as there is a difference now. And, last, if we really want to display dirt for gold tiles, that can easily be done in the tileset now.

oyvindln:
> I've added logs in the render manager to see if a tile was using claimedwall material and I've seen none. Could you please tell me how did you find it ?

I edited my post, it's not using the wrong material, I was just being stupid.

hwoarangmy:
ok, no problem. If you see something, please don't hesitate to tell.

@Danimal696 The new binary is uploaded ^^

oyvindln:
The change does seem to make 2x2+ lava areas look a bit odd. I don't think this was an issue in the past. Rotating the tiles also make the lava animations not line up. (Though it's not like the lava look all that great in the first place).
odscreenshot_2015-03-07_001641_0

hwoarangmy:
I've had a look at an old screenshot and, indeed, I've not used the expected lava mesh. Do you want me to change that ?

oyvindln:
I think that would be preferable yes.

hwoarangmy:
@oyvindln That's done
Concerning the lava animations, if you don't like, we can go back to using different meshes instead of rotating them. Or we may merge as it is and change later if needed.

oyvindln:
I'll merge this now then, the animations can be sorted later if we want.

Danimal696:
i think the best use would to be make different claimed walls, yup this has potential :)

PR #533, opened 2015-03-07
Fixed some memory leaks

oyvindln:


akien-mga:
Looks like you added a duplicate commit.

oyvindln:
I updated the commit as it was missing a newline somewhere.

akien-mga:
Ok, there were just two commits displayed in the PR, but that might have been a refreshing issue :)

akien-mga:
It builds and runs fine here.

source/gamemap/GameMap.cpp
```diff
@@ -2021,25 +2021,24 @@ bool GameMap::seatIsAWinner(Seat *s) const
     return false;
 }
 
-void GameMap::addGoalForAllSeats(Goal *g)
+void GameMap::addGoalForAllSeats(std::unique_ptr<Goal>&& g)
```
hwoarangmy, line 2024: you like unique_ptr as always.
I have to admit you are slowly converting me to using them ^^

source/network/ODServer.cpp
```diff
@@ -613,6 +613,8 @@ bool ODServer::processClientNotifications(ODSocketClient* clientSocket)
 
             Seat* seat = seats[0];
             seat->setPlayer(curPlayer);
+            //This makes sure the player is deleted on exit.
+            gameMap->addPlayer(curPlayer);
```
hwoarangmy, line 617: Have you seen a memory leak ? If yes, adding the player to the gamemap here is not the way to go because at this stage, there might be more players than available seats. That's why they are deleted in the ClientNotificationType::seatConfigurationSet message.
If there is a memory leak, we should find why the player is not deleted as it is expected.

source/network/ODSocketServer.cpp
```diff
@@ -52,7 +51,7 @@ bool ODSocketServer::createServer(int listeningPort)
     }
 
     mSockSelector.add(mSockListener);
-
+    mNewClient = new ODSocketClient;
```
hwoarangmy, line 54: Nice catch

source/network/ODServer.cpp
oyvindln, line 617: Yeah the player was not deleted in the editor mode (according to valgrind). As you are saying, I don't know if it was the best solution, the code was a bit hard to follow. Feel free to suggest something else. In gamemode the players are added to the  gamemap later, and deleted when it's cleared but it didn't look like that was happening in editormode.

source/network/ODSocketServer.cpp
oyvindln, line 54: That was just pure random luck, I wanted to look into making the server try an extra port if it failed. (As it seems, at least on my system there is some delay before they are released after the game is exited, which can be a bit frustrating when testing things.) That seemed to require a lot more code change, so I left it for now, but I noticed that there was a leak there.

hwoarangmy:
> the code was a bit hard to follow. Feel free to suggest something else

Yes, the server code in the initialisation phase is not the easiest to read. The basic idea is that when a new client connects to the server, it sends a message with its OD version (ClientNotificationType::hello).
The server checks that the OD version is the same and then, sends all the data about the loaded map (gold/rock tiles), name of the map, size, ... (ServerNotificationType::loadLevel). Note that the client do not load the level. Only the server needs it.
Then, if everything goes well, the client sends ClientNotificationType::levelOK and the server asks for the player nick (ServerNotificationType::pickNick).
The client answers by sending its nick (ClientNotificationType::setNick).
From there, the player is ready. That's where there is a difference between gamemode and editormode.
If we are in gamemode, the client will launch the configureseats mode. Once it is launched, a message will be sent to the server to make him know the client is ready for seat configuration (ClientNotificationType::readyForSeatConfiguration).
During seat configuration, messages will be sent to the server so that it can update all the clients (ClientNotificationType::seatConfigurationRefresh).
When the player acting as server will have finished to configure seats and will press start button, a message (ClientNotificationType::seatConfigurationSet) will be sent. Accepted players will start gamemode and rejected ones will get disconnected (and players will be deleted).

If we are in editormode, after receiving the ClientNotificationType::setNick, we assign automatically the player to the first available seat and the editor mode is launched. If guess that's where the player should be added to the gamemap. It will be more or less in ODServer.cpp:615 (after setting the player in the seat).

oyvindln:
> It will be more or less in ODServer.cpp:615 (after setting the player in the seat).

That is pretty much where I added it.

hwoarangmy:
lol. Stupid me. I had made a find to search where this line was added and I somehow ended before the editor test ^^
That means that the server code is not that unreadable (at least for you) ;-)

Testing it right now

hwoarangmy:
it works well. I merge

PR #534, opened 2015-03-08
Changed TileVisual to make it describe alone how the tile should look like

hwoarangmy:


akien-mga:
I don't know if it is linked to this PR or a previous one where I missed the regression, but I'm seen a strange behaviour with marked tiles:
od-marked
od-marked2

Another thing which is probably not related to this PR, the "Summon a worker" spell does not show the needed mana properly (at least on "Stonekeep Siege", where you start with two existing workers).
For me it shows "[0]" as the price, even though it requires the proper amount of mana (i.e. the first kobold that you summon costs 3000, since you already have two kobolds from the start).

oyvindln:
The tile selection issue is not happening for me when testing without this PR at least. (Haven't tested the PR yet.)

> For me it shows "[0]" as the price, 

For me it shows 0 when it's not over a tile that the spell can be cast on (Again without this PR), is it the same for you?

hwoarangmy:
> I don't know if it is linked to this PR

It is probably linked to this PR. I've changed the way the client knows a tile is markable or not (by using only the TileVisual and not the fullness). That can explain this regression if I forgot a test somewhere. As well as the mana needed for a spell. Differentiating client from server code in the gamemap is not always easy ^^. I will have a look tonight and fix it.
I did that beacause:
- It is much more straight forward to define tile links in the tileset only once and allow to link full tiles or not.
- It will make saving the tiles seen by each player more simple is we only have to save the tilevisual and seat (since there will be a lot of tile, not using the fullness makes a difference).

hwoarangmy:
Note to self: I need to update the comments in config/tilesets.cfg to adapt to how the tileset is defined now.

hwoarangmy:
@akien-mga The bugs you have seen should be fixed now ^^

Bertram25:
@hwoarangmy I'll test asap as well :) Yes, I've returned! Mwahaha

akien-mga:
Disclaimer: I'm testing from a different system than usually (still Mageia 5 on the same laptop, but a clean install of the RC to test the "out of the box" experience).

It's not related to this PR, but before I open an issue I wanted to check if it's a problem on my end or a real regression. When I click "Join game", OD segfaults with:

```
CEGUI::UnknownObjectException in function 'CEGUI::NamedElement* CEGUI::NamedElement::getChildElement(const CEGUI::String&) const' (/home/iurt/rpmbuild/BUILD/cegui-0.8.4/cegui/src/NamedElement.cpp:150) : The Element object referenced by 'LevelWindowFrame/ClientButton' is not attached to Element at 'Background'.
terminate called after throwing an instance of 'CEGUI::UnknownObjectException'
  what():  CEGUI::UnknownObjectException in function 'CEGUI::NamedElement* CEGUI::NamedElement::getChildElement(const CEGUI::String&) const' (/home/iurt/rpmbuild/BUILD/cegui-0.8.4/cegui/src/NamedElement.cpp:150) : The Element object referenced by 'LevelWindowFrame/ClientButton' is not attached to Element at 'Background'.
```

Bertram25:
Ok, I tested it. Seems to get there with yet another major feature added. :) Congrats!
Now as for the regressions:
Walls on the minimap aren't shaded in a different way than ground tiles, making it less sexy.
Also, while playing, my workers can now claim ground tiles under enemy rooms, making the whole enemy room object logic change in a funky way when updated against the FOW.
I also think the gold ground tiles should be dirt in the config. Even if that's only config, now. :)

Bertram25:
I confirm @akien-mga 's issue even on @oyvindln PR. I think it's a stealthy regression spotted by our hunter.

oyvindln:
The crash is probably from my earlier PR, I'll have a look at fixing it. It happens here as well.

hwoarangmy:
> Yes, I've returned! Mwahaha

Cool. I was fearing that you flee because scared of falling before my minions during our next lan

> When I click "Join game", OD segfaults

I've not done anything about this menu I can remember lately but I will have a look tonight

> Walls on the minimap aren't shaded in a different way than ground tiles, making it less sexy.

Another thing I forgot ^^ Will be fixed soon

> my workers can now claim ground tiles under enemy rooms

Hey, that's the second time this PR raises an old bug. But it should not be hard to fix.

> I also think the gold ground tiles should be dirt in the config

The reason why I did this PR is to remove "claimed" from tile types to avoid loosing information about its original type (for map saving).
Moreover, for the same reason, now, ground gold tiles are sent by the server to the clients as full gold tiles. The side effect is that is will allow interesting map changes like hiding creatures within gold tiles (instead of having dirt tiles that makes too visible that there is something hiden there).
That's why IMHO, it would be better to have different tile materials for ground dirt and ground gold.

akien-mga:
> > When I click "Join game", OD segfaults
> > I've not done anything about this menu I can remember lately but I will have a look tonight

This was a regression from an earlier PR, @oyvindln fixed it already in his.

>  That's why IMHO, it would be better to have different tile materials for ground dirt and ground gold.

I'll have a go at making a quick and dirty placeholder for the ground gold, basically the ground dirt with small patches of gold as if there were still a handful of gold veins.

Bertram25:
> > Yes, I've returned! Mwahaha
> 
> Cool. I was fearing that you flee because scared of falling before my minions during our next lan

Yeah, sorry for disappearing like that but here I had no choice. I'll try to warn before if it happens again.

> I'll have a go at making a quick and dirty placeholder for the ground gold, basically the ground dirt with small patches of gold as if there were still a handful of gold veins.

Cool, I was about to propose something similar. I do think just a tweaked texture + copy/pasted material for ground gold will do. Nothing hard at all if I'm not mistaken. :)

Bertram25:
Ah, and maybe the minimap should display ground gold tiles using the dirt color or something?

akien-mga:
> Cool, I was about to propose something similar. I do think just a tweaked texture + copy/pasted material for ground gold will do. Nothing hard at all if I'm not mistaken. :)

It is not as easy as I thought, the ground dirt material is based on a _blue_ texture (Dirt_dark_nor3.png) and the custom lighting is what makes it appear brown it seems, unless I missed something. So adding some gold in it produces a bad result since the lighting makes the gold appear dark brown in game.

Bertram25:
> It is not as easy as I thought, the ground dirt material is based on a blue texture (Dirt_dark_nor3.png) and the custom lighting is what makes it appear brown it seems, unless I missed something. So adding some gold in it produces a bad result since the lighting makes the gold appear dark brown in game.

(Just a wild guess, I'll try tonight) What happens if you create a new material file, which material is named for instance, 'GoldGround', copy/pasted from the 'Gold' material one and change the texture there to use yours.
Then, in the tileset you change the corresponding material parameter to 'GoldGround' ?

akien-mga:
> What happens if you create a new material file, which material is named for instance, 'GoldGround', copy/pasted from the 'Gold' material one and change the texture there to use yours.
> Then, in the tileset you change the corresponding material parameter to 'GoldGround' ?

Yes that's what I did initially, though I copied the 'Dirt' material and not the 'Gold' one (Gold still uses the old system without normals). Actually the 'Dirt' material has both a normal (the blue image, 'Dirt_dark_nor3.png') and a decal ('Dirt_darkE.png'). Since the normal is the one that gives it its relief, I misunderstood the whole system. Editing the decal is much better to achieve what I was looking for :-)

I'll do a PR based on @hwoarangmy's, so that it can be merged after this one.

akien-mga:
Well actually I'll make the PR once this one has been merged, so that you don't have to build @hwoarangmy's commits a second time :-)
My PR will be this commit: https://github.com/akien-mga/OpenDungeons/commit/156cff2879dfe378c579a92c0167b3b1c685bd05

hwoarangmy:
> Well actually I'll make the PR once this one has been merged, so that you don't have to build @hwoarangmy's commits a second time

It is better, otherwise, you will have to rebase after my PR gets merged

hwoarangmy:
I will not add more commits unless we see bugs in this PR. So, you can merge once tested ^^

akien-mga:
There's a regression with the FOW I think, on Scream in the dark for example, I can now see the claimed tiles of all the enemies (I don't see what happens there, but I see them as if I had discovered this area already).

hwoarangmy:
yep, big regression spotted. I should have launched a level to check. I'm on it ^^

Danimal696:
> Kobolds should not claim tiles where there is an enemy building

I hope traps are left out of that logic, we agreed claiming the tile they are over destroys them.

hwoarangmy:
> I hope traps are left out of that logic, we agreed claiming the tile they are over destroys them.

Yes, that's right. Only rooms should avoid kobolds from claiming, not traps

Bertram25:
Just tested it, works like a charm even after reloading and saving several times. Tested every rooms and traps and fought the green to death. The FOW is well restored AFAIK.
The only thing missing is that marked tiles aren't saved but IMHO it's very minor. Mergeable IMHO. :)

Congrats for the work done on this one.

hwoarangmy:
> The only thing missing is that marked tiles aren't saved but IMHO it's very minor

That might change. I will probably restore the old behaviour with asking to mark tiles and letting the server tell which tile is marked or not. Anyway, now, the server already knows the tile state for every seat so that's pretty straight forward. But that will be for another PR ^^

Bertram25:
:) Let's merge that one anyway. It's big enough. :D

PR #535, opened 2015-03-09
More gui cleanup + lighting tweaks

oyvindln:
Moved most of the callbacks out of gui class and created a shared base class for editor and game since they seem to share a lot of functionality. (I left the editor tile ones since @hwoarangmy is doing changes to that the latest PR.
Tweaked lighing values. The lighting [attenuation](http://www.ogre3d.org/tikiwiki/tiki-index.php?page=-Point+Light+Attenuation) was set up in a not very optimal way, the range was very large, which did seem to have some impact on performance. I lowered the range, and changed them to make the light attenuation more quadratic. They probably need more tweaking to make things look good though.


oyvindln:
Also it appears my editor decided to change the tabs in the map files to spaces. Don't know if that's a problem or not.

Bertram25:
Seems I can't drag 'n'drop a light in the map editor anymore.
No regression spotted apart from that. FPS look better, though. Need to try more to be sure.

oyvindln:
It's going to depend on the map a bit, I didn't go through and tweak all of them.

oyvindln:
I found the problem with the drag issue. I don't understand how this was done before though, I can't see any code that translates from the name of the indicator to the light node.

hwoarangmy:
> I left the editor tile ones since @hwoarangmy is doing changes to that the latest PR

Thanks ^^

> Also it appears my editor decided to change the tabs in the map files to spaces. Don't know if that's a problem or not.

It might be because some stuff use string.split at map load. Moreover, if one changes a level in the editor, he will end up with the whole file being modified. If I understand correctly what you've done (I've not been through your PR yet), you have only changed the lights. If you don't want to fight with your editor, I can change the lights according to the values you want in my PR. But please don't break the levels format.

Bertram25:
> But please don't break the levels format.

Does this mean you want him to fix back the tab/spaces in the level files?

hwoarangmy:
> Does this mean you want him to fix back the tab/spaces in the level files?

Yes please. At least revert to the original files. Again, otherwise, we would loose all the history the next time a level will be updated in the editor.

oyvindln:
Looking at the commit again, it's only removed a trailing tab in the comments, and in the light values, not the whole file, so I can change the light values back to tabs.
Are the trailing tabs supposed to be there, or is it a bug?

hwoarangmy:
> Are the trailing tabs supposed to be there, or is it a bug?

Yes, all the entities that same something end the line with a trailing tab (to allow upper entity to begin without knowing if there is already something or not)

oyvindln:
Should I readd the trailing tabs in the comments again then, or does it only matter for things that are not comments?

hwoarangmy:
The easiest is to open the map in the editor and save it without changing anything. That allows to let OD do what it wants and makes sure there will be no problem.
The basic idea being that if someone edits a map, he should see only his changes.

hwoarangmy:
It's good, I merge

PR #537, opened 2015-03-11
Fixed fps limiter

oyvindln:


Bertram25:
Yes, IMHO, using boost for such things is the way to go. Plus, we're using it for similar fashion, so it makes sense. Why? Because boost seems more mature than SFML in certain area. (No offence intended.) and using it to wide-spreadly isn't what it was included at first (we just wanted to use it for sound and network at first, AFAIR). And as you said, one less lib is always something better, I guess.

As for threading support, it's also a matter of taste, but as I know SFML can be both a good and bad thing, as it can be heavy while providing very useful wrapping functions.
I don't whether using boost::thread is better than boost for that but it can be seen after studying a bit of its code, I guess.

oyvindln:
I tried to use Boost::timer instead, though it is apparently a library, rather than just header only like much of boost, which means having to link to it if we want to use that as a timer rather than SFML. So I don't know if it's worth the bother. (I can put the fps limiter in a wrapper class in either case for easy replacing later if we want.)

Bertram25:
> (I can put the fps limiter in a wrapper class in either case for easy replacing later if we want.)

Arf, since we would lose on linking side, I guess let's forget about it. Does the fps limiter work better for you btw? I'm trying it now.

oyvindln:
Yeah, it limits the FPS to 60 as it's supposed to.

Bertram25:
Cool :) I'm compiling.

hwoarangmy:
note that we already embed boost thread lib in windows package. IIRC, it doesn't launch without it.

EDIT: I confirm it doesn't launch without it.

Bertram25:
ah, since we're using it. ;) Mind making the change, @oyvindln ?
Sorry for the back and forth, but well, it's development is hellish as a dungeon. ;)

oyvindln:
boost timer isn't in boost thread though. I'm not sure which library it is in, but it's one that we aren't linking to, as I got linker error when trying to use it.

oyvindln:
Hm, 60 fps seems to give about 59.5 fps, but when I test with 90 I get 85-87.

hwoarangmy:
> boost timer isn't in boost thread though

There is a library in boost called timer. Probably the one needed. In any case, IMHO, it would be better to not use it for such a simple problem.

akien-mga:
For the reference, boost libraries as of 1.55.0:

```
libboost_atomic1.55.0
libboost_chrono1.55.0
libboost_context1.55.0
libboost_coroutine1.55.0
libboost_date_time1.55.0
libboost_filesystem1.55.0
libboost_graph1.55.0
libboost_iostreams1.55.0
libboost_locale1.55.0
libboost_log1.55.0
libboost_math1.55.0
libboost_mpi1.55.0
libboost_prg_exec_monitor1.55.0
libboost_program_options1.55.0
libboost_python1.55.0
libboost_random1.55.0
libboost_regex1.55.0
libboost_serialization1.55.0
libboost_signals1.55.0
libboost_system1.55.0
libboost_thread1.55.0
libboost_timer1.55.0
libboost_unit_test_framework1.55.0
libboost_wave1.55.0
libboost_wserialization1.55.0
```

Bertram25:
> > boost timer isn't in boost thread though. I'm not sure which library it is in, but it's one that we aren't linking to, as I got linker error when trying to use it.
> 
> There is a library in boost called timer. Probably the one needed. In any case, IMHO, it would be better to not use it for such a simple problem.

Ok, then I guess keep using SFML is fine for now. The only fix I'd still request is the one from @hwoarangmy 's notice aka use standard integer types in the FpsLimiter class and no direct sf::\* types.

Later, if it is becoming problematic, we'll switch to whatever seems better at that moment. We're getting there. :)

oyvindln:
rebased and changed to use unsigned ints in the interface.

source/entities/MapLight.cpp
```diff
@@ -20,15 +20,12 @@
...
-#include "modes/ModeManager.h"
 #include "network/ODPacket.h"
 #include "network/ODServer.h"
 #include "network/ServerNotification.h"
 #include "render/RenderManager.h"
-#include "render/ODFrameListener.h"
-#include "utils/Helper.h"
 #include "utils/LogManager.h"
-#include "utils/Random.h"
+//#include "utils/Random.h"
```
Bertram25, line 28: Btw, I guess this can be removed (later)

oyvindln, line 28: Yeah, there was some code I commented out about the light flickering, and that's part of it. I wasn't sure if I should fully remove it or not, so I left it commented out for now.

Bertram25, line 28: Np. This can wait for another PR or something. It's already good to see lights being functional again, even if not (yet) changing.
There is not much before release btw. :)

hwoarangmy:
works well. I merge

PR #538, opened 2015-03-11
Save marked tiles when saving game + small fix

hwoarangmy:


Bertram25:
I've got a crash when loading this save (saved with latest tip of your repo):
http://filebin.net/nprabgnk6u

The log doesn't indicate much. Just that it was on turn 0.

Bertram25:
Other than the crashy save, it is working fine. :)

hwoarangmy:
I've seen what was going on. The tile states is read while building seats. The problem is that when reading the level, if we read a tile owned by a seat that has not been read so far, the seat will be set to nullptr and it will crash when the seat will be used.
I'm fixing it right now by only saving the seat id instead of the seat.

hwoarangmy:
@Bertram25 Note that your savegame is perfectly valid and you can use it to test the fix ^^

Bertram25:
Cool, I'll test asap. :)

Bertram25:
Works great. I was able to resume my game. :) To me it's mergeable.

akien-mga:
It works great. I found an issue with the name of the save games, but I'll open a specific issue for that, it's not particularly related to this PR AFAICT.

PR #543, opened 2015-03-13
Move keeper hand and light more smoothly

### LINKED ISSUE #542: keeper hand should move more smoothly
A stated in #420

hwoarangmy:
fixes #542


akien-mga:
This simply makes the game look gorgeous :-D

hwoarangmy:
I've just seen that the keeper hand is too low (under full tiles). I will make a commit for that

source/modes/GameMode.cpp
```diff
@@ -440,12 +423,19 @@ bool GameMode::mouseMoved(const OIS::MouseEvent &arg)
 
     handleMouseWheel(arg);
 
-    handleCursorPositionUpdate();
+    if (mMouseX != inputManager->mXPos || mMouseY != inputManager->mYPos)
+    {
+        mMouseX = inputManager->mXPos;
+        mMouseY = inputManager->mYPos;
+    }
```
oyvindln: Is there really any need for doing a test before assigning here?

hwoarangmy: You are right. This was used to avoid setting light position when it didn't move from the tile. Since now, we always update the keeper hand position, I will remove them completely

hwoarangmy:
If everybody agree, I will also do the same in the editor

akien-mga:
> If everybody agree, I will also do the same in the editor

I agree, it's better to be consistent IMO.

Bertram25:
Yes ! :)

Bertram25:
I've tested it. My god, it makes so smoother and somehow feeling more responsive. :D 

Some tweaks I'd request:
As you noticed, the Keeper hand and light is too low compared to the tiles.
The mouse pointer is pointing on the hand's bottom which is counter-intuitive. I would make it point on the middle of the hand model fingers.

Apart from that, it is working great. :)

hwoarangmy:
That's done

> The mouse pointer is pointing on the hand's bottom which is counter-intuitive

I agree but that would mean to hardcode an offset for the hand. Wouldn't it be better to change the mesh (as we already need other animations anyway) ?

Bertram25:
> I agree but that would mean to hardcode an offset for the hand. Wouldn't it be better to change the mesh (as we already need other animations anyway) ?

I agree only if @Danimal696 can fix this before next release. I f he can't, we should hardcode it, IMHO.

akien-mga:
It works great as far as I'm concerned :)

hwoarangmy:
@Danimal696 I've had a look at the keeper hand but the animations seems not to be the same in the blender file thanin game. Could you please tell me if the one in the media repo is the latest one ?

Danimal696:
i had a look at it for the first time, it turns out it has all the animations we need and more! its just a matter of remaning and cleaning NLA tracks. Its also the only model of it that im aware of:

opendungeons-svn-284\mediaSource\Models\Keeperhand 
AND git -> https://github.com/OpenDungeons/OpenDungeons-media-source/tree/master/Models/Keeperhand are the same

Bertram25:
Seems we have a good surprise with the model... at hand.

hwoarangmy:
> i had a look at it for the first time, it turns out it has all the animations we need and more!

Yes, I've seen that. BTW, I've been fighting with it to change the center so that the fingers are displayed where the cursor is (that worked). But I couldn't manage to rotate it. When I rotate in the blender file, it works, but when I export it to mesh file, it is always in the same position (and I have to rotate in game, what I want to avoid)
@Danimal696 Is there something special to do to activate rotations ?

> Seems we have a good surprise with the model... at hand.

^^

Danimal696:
have you tried to apply rotation before exporting?  Ctrl+a -> apply everything while you are at it (location, rotation, scale) if it doesnt work ill have a look myself.

hwoarangmy:
It seems to work when applying rotation, thanks ^^
I will test and upload that if it works ASAP

hwoarangmy:
Here is the keeper hand with some new animations. I've not removed the unused ones so we can test if we want.

akien-mga:
Nice new hand animations, we can now slap creatures and grab them properly! \o/

I think though that the mouse pointer is too high now (it's at the tip of the fingers), I would expect the hands active spot to be under the palm or the beginning of the fingers. It's particularly visible when grabbing a creature: the hand grabs the air of the tile below the targeted creature :)

akien-mga:
@Bertram25 Our unstoppable coder is starting to push commits to improve the AI, I think it's time we merge the beast :-p I guess the further adjustment of the keeper hand can be addressed in another issue/PR.

Bertram25:
> @Bertram25 Our unstoppable coder is starting to push commits to improve the AI, I think it's time we merge the beast :-p I guess the further adjustment of the keeper hand can be addressed in another issue/PR.

eh eh thanks :D I'm testing atm. The code looks fine. If I see no big regressions, I'll merge. :)

Bertram25:
Working for AI and for the hand. It's fun :D

PR #550, opened 2015-03-16
Research GUI - Step I

Bertram25:
This PR adds the GUI for the research tree that gets updated while your skills are getting done.

I've done a few cleanups as well.

What's lacking is to display what research is currently being done.

And as a second step being able to reorder the research queue.

Part of #517 

Feel free to comment :)

### DIFF
--- a/gui/OpenDungeons.looknfeel
+++ b/gui/OpenDungeons.looknfeel
@@ -4059,13 +4059,13 @@
         <ImagerySection name="label">
             <TextComponent>
                 <Area>
-                    <Dim type="LeftEdge"><AbsoluteDim value="0" /></Dim>
+                    <Dim type="LeftEdge"><AbsoluteDim value="5" /></Dim>
                     <Dim type="TopEdge"><AbsoluteDim value="0" /></Dim>
                     <Dim type="Width"><UnifiedDim scale="0.9" type="Width" /></Dim>
                     <Dim type="Height"><UnifiedDim scale="1" type="Height" /></Dim>
                 </Area>
-                <VertFormat type="BottomAligned" />
-                <HorzFormat type="RightAligned" />
+                <VertFormat type="TopAligned" />
+                <HorzFormat type="LeftAligned" />
             </TextComponent>
         </ImagerySection>
         <ImagerySection name="normal">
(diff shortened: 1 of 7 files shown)


akien-mga:
It works great! :-D

akien-mga:
I've also confirmed that the research window gets updated when a new research is done :)

Bertram25:
For info, the ImageButton widgets now support a StateImage property as you have seen, displayed on the bottom-right part of the button, a ButtonImage as seen on the button center, and a label that you can set using the setText() function on the widget, seen on the top-left part of the button.

What I'll be working on is.
1. Showing what research is currently being researched.
2. Try to exposed what percentage of it is done. (I don't know how yet). Using something similar to a progressbar would be cool (Like the button gets undimmed proportionally to what is researched) but I don't know whether it is feasible yet. If not, I may show the percentage in text maybe? WDYT?
3. The buttons are already clickable but don't do anything. I'll use the text to show the pending queue. and in a second row have a look at how we could change it within the permitted researches.
4. Add tooltips where relevant / make the code more generic.

I'd like to do 1. and maybe 2. for 0.5.0. at least.

Feel free to comment if you have a better idea or see anything out of place.

hwoarangmy:
> If not, I may show the percentage in text maybe?

If displaying a progressbar is too hard, you can also display a color. From white (0% research) to black (100%), for example. That may be easier.
For having messed with CEGUI, I can see the work it is to achieve a nice looking GUI. That's already cool to have the GUI we currently have ^^

Bertram25:
> If displaying a progressbar is too hard, you can also display a color. From white (0% research) to black (100%), for example. That may be easier.

Sounds like a good idea. :) I'll keep that in mind. Thanks.

> For having messed with CEGUI, I can see the work it is to achieve a nice looking GUI. That's already cool to have the GUI we currently have ^^

Thanks. :)

hwoarangmy:
Testing right now ^^

hwoarangmy:
Tested. The GUI is cool and it is updated in real time. Really nice. If I may, some color would be cool if possible (green check for researches done)

Bertram25:
> Really nice. If I may, some color would be cool if possible (green check for researches done)

Sure :)  I'll do it for the next PR.

PR #554, opened 2015-03-17
Rename test levels and reorder iconset

akien-mga:
I also fixed the CEGUI version in the gui/OD.project file, since it was set to 1.0 and my CEED would throw an error since it's not compatible with CEGUI 0.8.x.

At first I wanted to edit the imageset by hand... then I realised that I can do it with CEED. I'm glad I packaged it for Mageia, it makes it relatively easy to edit such files :-)


hwoarangmy:
works well

Issue #529, opened 2015-02-28
Fix all materials with normals - Log error while parsing material files


hwoarangmy:
I guess it comes from @Danimal696 normal activation. I have logs:
Compiler error: unknown error in TrainingDummy1.material(23): token "ambient" is not recognized
Compiler error: unknown error in TrainingDummy1.material(24): token "diffuse" is not recognized
Compiler error: unknown error in TrainingDummy1.material(25): token "specular" is not recognized
Compiler error: unknown error in TrainingDummy1.material(26): token "emissive" is not recognized

I have that on many materials (StoneCoffin, TrainingDummy2, TrainingDummy3, TrainingDummy4, Treasury, Trollbed, ...)

If it is not a problem of my graphic card not supporting something, I guess they could be removed from material files safely.

Danimal696:
That is strange, those values are used to tweak how much ligth is received, deflected or emitted. Works well on my end, erasing them would make the models look weird.

oyvindln:
They are put in the wrong place.

Danimal696:
then what would be the correct one?

oyvindln:
You have to put the lighting values in the pass they are supposed to apply, not inside the texture unit. I'm not sure if they should go under the normal map pass or the texture pass, you will have to test.

```
// Treasury genrated by blender2ogre 0.6.0
import RTSS/NormalMapping_MultiPass from "RTShaderSystem.material"
material Treasury : RTSS/NormalMapping_MultiPass
{
    receive_shadows on

    technique
    {
        pass lighting
        {

             // Override the normal map.
             rtshader_system
             {
                lighting_stage normal_map Treasury_normal.png object_space 0 bilinear 1 -1.0
             }
        }

        pass decal
        {
            ambient 0.8 0.8 0.8 1.0
            diffuse 0.65 0.65 0.65 1.0
            specular 0.4 0.4 0.4 1.0 12.5
            emissive 0.0 0.0 0.0 1.0

            texture_unit decalmap
            {
                texture Treasury.png
                tex_address_mode wrap
                scale 1.0 1.0
                colour_op modulate
            }
        }
    }
}
```

Danimal696:
Axctually is hwoarangmy who will have to, please tell me when you get the rigth place and ill fix them

oyvindln:
You should be able to do that too, try to put them in one of the passes and which one makes the material look like you want. They should be generating errors for everyone. They are for me.

[...]

PR #555, opened 2015-03-17
Turned the Forge into the Workshop + Work on the Research GUI

### LINKED ISSUE #459: Turn the forge into the workshop
The current forge should be turned into the workshop.

The workshop will still be used to craft traps, but its models have to be changed as the forge models will be reused to create the new forge, where creature equipment will be forged.

Also the workshop logic will have to be upgraded so that traps are crafted only when needed and not at random. More info at: http://forum.freegamedev.net/viewtopic.php?p=61975#p61975 (and on)

Bertram25:
And changed every reference in the code, in levels and in config accordingly.

For I left the forge wall items in place i the wait for new ones from Danimal.

Fixes #459


Bertram25:
@Danimal696 As I badly wrote in the commit message above, I left the forge wall objects for now on the Workshop. The beast looks good in game. :)
I wonder whether you want the table to be rotated to 45° though. Just tell me what you prefer and I'll adapt accordingly. 

Do you think for instance it would look better if the creatures are rotated 45° also and placed so that we think they're facing the tables when working on traps?

hwoarangmy:
> Shouldn't we remove the cannon trap from the starting skills and add it to the research to be done instead?

ok, I see no problem. The reason was more to test than to have the final behaviour. Now that the research logic is tested, you can go on with what is more logical

> Do you think for instance it would look better if the creatures are rotated 45° also and placed so that we think they're facing the tables when working on traps?

If the tables are rotated, it would be better to rotate the creatures, yes.

Bertram25:
> ok, I see no problem. The reason was more to test than to have the final behaviour. Now that the research logic is tested, you can go on with what is more logical

That said, we should maybe let it as is in test maps since they're here for that.

> If the tables are rotated, it would be better to rotate the creatures, yes.

Depends on whether people prefer to have tables not rotated but I think I'll rotate the creatures if nobody else minds. ;)

hwoarangmy:
> Depends on whether people prefer to have tables not rotated but I think I'll rotate the creatures if nobody else minds

Anyway, that's easy to revert if players complain so...

akien-mga:
> Depends on whether people prefer to have tables not rotated but I think I'll rotate the creatures if nobody else minds. ;)

Let's rotate then ;)

Bertram25:
0.5.0 will be cool guys. :sunglasses: Can't wait to have a lan with you all. ;)

Danimal696:
yeah, it will be a nice touch to rotate creatures along with tables

Bertram25:
Research Gui upgraded! :)

akien-mga:
It's awesome! The new research GUI is really great. :-)

Some remarks on the workshop:
- The tables are properly rotated, but I think they are not positioned via their geometrical center, so they are now offset to the top and mess with wall objects:
  od-workshop
- If I may, how about rotation the tables by 30° instead of 45°? I think it could give a nice feel if they were not aligned on the diagonal.
- I haven't seen any creature go to the workshop by themselves to work, I just managed to get an orc working for one second by dropping him on the workshop. Is the logic already in place for that, or are creatures still looking for the forge?

Bertram25:
Heya :)

@Danimal696 Btw, while I see it, the Grimoire, the Workshop ground and machine1 materials are all three triggering warnings. Same thing than #529 

@akien-mga 

> The tables are properly rotated, but I think they are not positioned via their geometrical center,

Yeah, you must be right.

> If I may, how about rotation the tables by 30° instead of 45°? I think it could give a nice feel if they were not aligned on the diagonal.

Sure, I'll try that and push something if it turns out well. :)

> I haven't seen any creature go to the workshop by themselves to work, I just managed to get an orc working for one second by dropping him on the workshop. Is the logic already in place for that, or are creatures still looking for the forge?

During my tests, I was able to make the creature produce several traps (when they are not hungry, not sleepy and not brooding - Man, those minions still believe it's a democracy -, and that at least a trap has been placed on map waiting for its crafted counterpart.)
So normally, it should be fine. I'll test again though, just in case.

Btw, your (dark) picture makes me think that we may want to try to later create a simple light source per room centered on the room latest shape, now that @oyvindln has fixed the light problem quite well using sane parameters for them.

Danimal696:
Materials again? ok, ill fix it...

hwoarangmy:
> Materials again? ok, ill fix it...

I think @oyvindln said he will handle the materials. But it would be nice to not add new materials with warnings again because he might not see them.

akien-mga:
The buttons unlocked through research (rooms, traps, spells) are not reset when quitting a game, see #559.

Bertram25:
Ok, I'm on it. Thanks for the feed-back. :)

Bertram25:
Ah, btw, @Danimal696 didn't you say that the grimoire had animations?

Danimal696:
yes, its called "Loop"

Bertram25:
> yes, its called "Loop"

Ok, then I'll make it call that animation. :)

hwoarangmy:
Works well. I merge

PR #562, opened 2015-03-20
Display number of controlled creatures + fixes for #564 and #567

akien-mga:
Contributing code 101 \o/

Don't hesitate to be comment extensively, I mostly implemented this feature by mimicking what was already implemented for the other similar GUI elements (number of claimed tiles, amount of gold, etc.).

So I might have missed already implementing functions, or done things in a non-optimal way :-)

### DIFF
--- a/gui/OpenDungeonsGameModeMenu.layout
+++ b/gui/OpenDungeonsGameModeMenu.layout
@@ -71,6 +71,26 @@
                     <Property name="VerticalAlignment" value="Centre" />
                 </Window>
             </Window>
+	    <Window type="OD/StaticText" name="CreaturesDisplay" >
+                <Property name="Area" value="{{0,410},{0,0},{0,540},{0,24}}" />
+                <Property name="Text" value="0" />
+                <Property name="MaxSize" value="{{1,0},{1,0}}" />
+                <Property name="TooltipText" value="Your creatures pool" />
+                <Property name="HorzFormatting" value="RightAligned" />
+                <Property name="ClippedByParent" value="False" />
+                <Property name="VerticalAlignment" value="Centre" />
+                <Window type="OD/StaticImage" name="Icon" >
+                    <Property name="Area" value="{{0,0},{0,0},{0,30},{0,30}}" />
+                    <Property name="Image" value="OpenDungeonsIcons/CreaturesIcon" />
+                    <Property name="MaxSize" value="{{1,0},{1,0}}" />
+                    <Property name="AlwaysOnTop" value="True" />
+                    <Property name="TooltipText" value="Your creatures pool" />
+                    <Property name="FrameEnabled" value="False" />
+                    <Property name="ClippedByParent" value="False" />
+                    <Property name="BackgroundEnabled" value="False" />
+                    <Property name="VerticalAlignment" value="Centre" />
+                </Window>
+            </Window>
         </Window>
         <Window type="OD/ImageButton" name="HelpButton" >
             <Property name="Area" value="{{1,-53},{0,5},{1,-28},{0,30}}" />
--- a/source/ODApplication.cpp
+++ b/source/ODApplication.cpp
@@ -57,15 +57,15 @@ ODApplication::ODApplication()
     }
     Random::initialize();
     //NOTE: The order of initialisation of the different "manager" classes is important,
-    //as many of them depen on each other.
+    //as many of them depend on each other.
     ResourceManager resMgr;
     std::cout << "Creating OGRE::Root instance; Plugins path: " << resMgr.getPluginsPath()
               << "; config file: " << resMgr.getCfgFile()
               << "; log file: " << resMgr.getLogFile() << std::endl;
 
     Ogre::Root ogreRoot(resMgr.getPluginsPath(),
-                            resMgr.getCfgFile(),
-                            resMgr.getLogFile());
+                        resMgr.getCfgFile(),
+                        resMgr.getLogFile());
 
     resMgr.setupOgreResources();
 
--- a/source/game/Seat.cpp
+++ b/source/game/Seat.cpp
@@ -800,7 +800,7 @@ ODPacket& operator<<(ODPacket& os, Seat *s)
     os << s->mId << s->mTeamId << s->mPlayerType << s->mFaction << s->mStartingX
        << s->mStartingY;
     os << s->mColorId;
-    os << s->mGold << s->mMana << s->mManaDelta << s->mNumClaimedTiles;
+    os << s->mGold << s->mMana << s->mManaDelta << s->mNumClaimedTiles << s->mNumCreaturesControlled;
     os << s->mHasGoalsChanged;
     os << s->mNbTreasuries;
     uint32_t nb = s->mAvailableTeamIds.size();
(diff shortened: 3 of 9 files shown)


akien-mga:
Well, I just found a bug thanks to this new feature :-P
We don't check MaxCreaturesPerSeat when adding new workers with a spell (#564).

akien-mga:
Actually I just found the GameMap::getCreaturesBySeat function, so I guess the getter I added was not needed. I'll let you point me in the right direction before I change my commit though, maybe my getter is better than having to assess the size of a std::vector<Creature*>?

hwoarangmy:
Your getter is better. That's because GameMap::getCreaturesBySeat is more a server function. In this case, that would work because clients will see all their own creatures and, thus, computing it on client side should be correct. But because everything else is computed on server side and sent to the clients (where you have already seen), it is better like you did.

akien-mga:
The fix for #564 was not as trivial as I thought, since I had to do a check in both functions, so that you don't pay for a worker that will be refused afterwards ;-)

I'm not 100% about the variables definitions, I used int32_t as was done for the other variables, but I'm not familiar with the type so I don't know if it's the best choice. I'm also unsure whether some variables should be `const`, but since other variables like nbFreeWorkers are not, I chose the safe way :-)

hwoarangmy:
> I used int32_t as was done for the other variables, but I'm not familiar with the type so I don't know if it's the best choice

When you are not sure, int32_t is the best choice ;)

> I'm also unsure whether some variables should be const

You can't change the value of const variables. Usually, you will want to use them for things like maximum number or such. In this case, it could be for maxCreatures, nbFreeWorkers, nbCreaturesControlled, maxMana and pricePerWorker. However, I don't think it hurts performances too much. Maybe @oyvindln will have something to say about it.

hwoarangmy:
I've tested and it works fine. I will merge it tonight if nobody does before

akien-mga:
> I've tested and it works fine. I will merge it tonight if nobody does before

I just added another commit that would be worth testing on both Windows and Linux, since it involves messing with files.

Here it's working fine on Linux with boost 1.55.0.

Danimal696:
seriously against fix for #564 , its a game breaker

Bertram25:
> seriously against fix for #564 , its a game breaker

Funnily, I was somehow writing about that. There are few things, I'd like to precise about the creature number. The 15 value is fact what we defined as the max number of spawnable creatures through the portal.
That number should even be increased by something like 5 per supplementary portals.
The number of worker creatures and the number of creature spawned at the crypt aren't even considered by this limit.

Hence, this fix is looking good in term of code, but isn't what should be done for the gameplay.

I'd request that the max number of creature is turned into the max number of portal creature and that this number is displayed between brackets for instance in the GUI if displayed at all.
I'd also request that this limit is checked only for portals.

Other creature limits have to be defined, I guess.

Feel free to comment :)

akien-mga:
Ok, so basically #564 would be an invalid issue, and the real bug is the fact that workers are taken into account when spawning new fighting creatures.

That will be a real gameplay changer though (so far I'm sure none of us had more than 11-12 fighters because of this limitation), more than my current fix for #564 which just prevents cheating, so we should make sure that everybody agrees that workers should not be taken into account into the number of creatures.

I for one would be fine with the proposed change. I'm just unsure about what would be the best implementation? As a newbie with a very poor knowledge of our code base, I would propose:
- Add a test in GameMap::doTurn ( https://github.com/OpenDungeons/OpenDungeons/blob/development/source/gamemap/GameMap.cpp#L1027-L1037 ) to check whether the creature is a worker, and if so increment a new variable, e.g. mNumWorkersControlled. If not, increment the other variable that could be renamed to mNumFightersControlled (or do we use another name to differentiate workers from non-workers creatures?)
- Expose the two variables with getters
- Use getNumFightersControlled in RoomPortal::doUpkeep ( https://github.com/OpenDungeons/OpenDungeons/blob/development/source/rooms/RoomPortal.cpp#L101-L107 ) instead of getting the total number of creatures
- Do the same in RoomCrypt::doUpkeep ( https://github.com/OpenDungeons/OpenDungeons/blob/development/source/rooms/RoomCrypt.cpp#L144-L166 )
- Don't limit worker summoning for now since it's already limited by mana

Bertram25:
> Ok, so basically #564 would be an invalid issue, and the real bug is the fact that workers are taken into account when spawning new fighting creatures.

Yeah, well, sorry and welcome in development hell. ;P 

> That will be a real gameplay changer though (so far I'm sure none of us had more than 11-12 fighters because of this limitation), more than my current fix for #564 which just prevents cheating, so we should make sure that everybody agrees that workers should not be taken into account into the number of creatures.

IMHO, yes and no. This just needs to be defined more accurately IMHO.
Please note that the number of workers per seat is already limited by the mana required to spawn one increasing exponentially. Atm, you cannot have more than say 6 or 7 of them IIRC.
The number of creature spawn per portal is indeed 15 + 5 per additional portal (That's base values than and should be tweaked later, still IMHO.)

Also, I do think an absolute max number of creature per seat should be defined and applied in game, but it should be higher than 15 + 8 creatures. Let's 25-30 as a first row?
And the game should in any case warn the player it has reached the max.

WDYT?

Bertram25:
> copy file is worst than rename in this case. Actually, renaming is exactly what should be done. Could you please remove this comment ?

Agreed.

akien-mga:
Redid the fix for #567 without the comment about copy_file, as requested by @hwoarangmy.

akien-mga:
Ok, so I propose to nuke my two commits for #564 from this PR, since they implement a fix that would have to be reverted once the overall logic has been fixed :)

Maybe I should also copy-paste the relevant discussion in #564 and we can continue discussing it there?

hwoarangmy:
> Redid the fix for #567 without the comment about copy_file

IMHO, its better to had new commits rather than overwriting existing ones (to avoid loosing history).

> seriously against fix for #564 , its a game breaker

IMHO, it is a sane issue that should be fixed. I know you want to have more creatures than the current limit but using a bug (in fact, it really is a bug since you don't have the same behaviour when spawning workers before or after reaching the limit) is not the way to go.

We should define a strategie (maximum creatures computation like @Bertram25 said about owning portals). Then, we should define if workers should be taken into account when computing creature numbers (from what I've read here, it seems "no" is the most wanted answer).
When the time comes, we might also want to have some creatures that also do not count like skeletons.

WDYT ?

EDIT:

> Maybe I should also copy-paste the relevant discussion in #564 and we can continue discussing it there?

Good idea ^^

akien-mga:
> IMHO, its better to had new commits rather than overwriting existing ones (to avoid loosing history).

I'll keep that in mind :-)

Danimal696:
yes, creatures from the crypt and future prison shouldnt be counted in the max, else there is no reason to have those rooms built in the first place

akien-mga:
> IMHO, it is a sane issue that should be fixed. I know you want to have more creatures than the current limit but using a bug (in fact, it really is a bug since you don't have the same behaviour when spawning workers before or after reaching the limit) is not the way to go.

That sounds good to me. My PR fixes a bug in the current logic, so even though we all agree that the logic should be changed, there's no reason not to use the fix in the meantime. Then I can open a new issue to continue discussing the spawning logic.

Bertram25:
> yes, creatures from the crypt and future prison shouldnt be counted in the max, else there is no reason to have those rooms built in the first place

I'm precising: to the portal(s) max :) The workers number is already limited and should indeed not be taken in account for fighters.
What's left to defined as max is the overall total creature, which will implicitly define the max crypt creatures.

akien-mga:
Now keeping the last three log files with 6f0c6cd, as discussed on https://github.com/akien-mga/OpenDungeons/commit/c74fa49025af530e7f5023a779e84fb3c50bc8ab.

hwoarangmy:
Works well, I merge

PR #571, opened 2015-03-23
Material fixes

Danimal696:
if Ok, ill start fixing them all little by little


Danimal696:
The bridges you can see are actually tentacles using bridge mesh, those are just for show, but since bridges will be some kind of extension of floor tiles i adapted claimed material for it, since it didnt go well i tried modding the claimed material to use the one for bridges, the result is that somehow the normal for claimed are applied instead my woodbridge normals, they are somehow hardcoded.
bridges

Bertram25:
Bridges would need specific code to be testable but I no more have warnings on the materials.

Danimal696:
then merge as you see fit, ill begin fixing the other materials lil by lil

akien-mga:
Then I merge, but please give us the needed info to update the CREDITS for the bridge :)

Bertram25:
Ah yes, the credits... I'll handle it if you want but yeah, we'll need them.
Nice catch @akien-mga

Issue #579, opened 2015-03-24
Spring cleaning in the materials/ folder


akien-mga:
It is currently relatively hard to know which texture is linked to which material, and which material is used in game or not.

According to @oyvindln's input on #492, most material IDs (defined in the materials/script/*.material files by the `material` tag) are referenced directly in the binary meshes. The we reference the meshes in the config/ files.

So basically to know the status for each material/model, we can:
- Check that every texture is used in a material script
- Extract the material IDs from the scripts
- Check the meshes against those material IDs to know if they're using them
- Check the meshes against the config files (and the code if there are some hardcoded meshes) to know which are used

We don't necessarily have to do exactly this for each step, since the data folders are not so huge so for most models we know if we see them in game already.
I'm just trying to single out the textures/models that we might not be using actually.

akien-mga:
For the first two points, I wrote a shell script that checks all textures files against the material scripts, and extracts the material IDs.

```
#!/bin/sh

search_source=false

pushd materials/textures > /dev/null
ls * > ../../../textures-list.txt
popd > /dev/null

for texture in $(cat ../textures-list.txt)
do
    echo "== Texture: $texture =="
    echo ". Files matching the filename:"
    ack -il $texture
    for script in $(ack -il $texture | grep materials/scripts)
    do
        line=$(cat $script | grep "^ *material " | sed -e 's/^ *material //')
        echo -e ". Matched material definition:\t$line"
        for word in $line
        do
            if [ -n "$(echo $word | grep RTSS)" -o -n "$(echo $word | grep :)" ]
            then
                continue
            fi
            echo -e ". Extracted material name:\t$word"
            if $search_source
            then
                echo ". References to $word in the source code:"
                pushd source > /dev/null
                ack -il $word
                popd > /dev/null
                echo ""
            fi
        done
    done
    echo ""
done
```

It takes a little white to process everything, but the output can be useful already to see that we have quite a few apparently unused textures.

akien-mga:
Here is the output when running this script in the current development branch: https://gist.github.com/akien-mga/7c1d4825ee321d7ce4a3#file-texture-analyse-txt

Bertram25:
Excellent :D This will definitely help!
I'd also add that models that aren't using materials should whenever possible. So tracking which model isn't using a material is also something to do and fix, right?

Bertram25:
I vote for adding that script (with documentation header and copyright) into a tools/ folder for instance.
[...]

Issue #492, opened 2015-02-11
Certain textures are unused.


Bertram25:
We should list them, and then move them into the media repository.
- [x] materials/textures/Env-blaugruenF.png

Bertram25:
@Danimal696 While you're dealing with textures and materials, could you have a look at which are unused? I know I may be demanding but you'll have an opportunity to avenge on our next lan. ;)

Danimal696:
ok, ill have a look

akien-mga:
I'd like to give a hand on this one, since I know @Danimal696 is busy and we have more interesting tasks for him :-D

But I need to understand better how the textures are linked to our meshes, etc. Do we strictly rely on the text materials files, or are some textures hard-linked into binary meshes? Said differently, can I find out which textures are used and which aren't just by searching for their filename in text files, or do I also need to use blender? :)

Bertram25:
@akien-mga I'd say, watch the material files, move the textures not listed elsewhere and have an advanced play with both factions. Then, if the logs don't complain about something missing, you are good to go. :)

Then, you can list them here and we can let @Danimal696 decide for each of them whether he wants to keep it (and then we move it to the media repo) or not.

Then, the job is easy (Watch for the CREDITS file update. ;])

Danimal696:
by chance i discovered its used by the hand, now it depends if @hwoarangmy exported the hand using it or not to remove from texture files. All the the other textures seem to be ok and used.

hwoarangmy:
I just used blender to ogre feature. I didn't change anything about textures.

Danimal696:
then please look at the hand material, if its not in there it can be erased (you have a newer version than i do)

Bertram25:
@Danimal696 Sorry, I'm not getting what you both talked about here:
- Is the Env-blaugruenF.png file used by the hand mesh? Or is it another texture?
- Are the other textures all used for something?

Danimal696:
> Is the Env-blaugruenF.png file used by the hand mesh? Or is it another texture?

Yes, by the hand only

> Are the other textures all used for something?

at least for the ones into the texture folder, i did a clean up some time ago, the rest are clean

Bertram25:
Ok, bonus questions: ;)
- Is this material used by the hand? (easily testable but I can't atm)
  https://github.com/OpenDungeons/OpenDungeons/blob/development/materials/scripts/Keeperhand.material

More general question while I think about it (or note to self):
[...]

PR #582, opened 2015-03-24
Add a script to retrieve materials corresponding to textures and meshes

akien-mga:
The original script from #579 is in cac6a01, then I improved it a bit based on @oyvindln's suggestions.
I put it in a tools/ directory and move my Linux portable binary creation scripts there too as it's more fitting.

I also moved Dm3d to past contributors since he's not active anymore.


hwoarangmy:
The scripts seem pretty well done. I will let @Bertram25 or @oyvindln play with them as I don't have a linux machine to test ^^

Bertram25:
Works great but beware you didn't check for ack existence. ;)
Ah, and I would tell the usual package name where Ogre tools and ack are found.

Mergeable in any case, IMHO. This can be added later.

akien-mga:
> Works great but beware you didn't check for ack existence. ;)
> Ah, and I would tell the usual package name where Ogre tools and ack are found.

Done :-)

akien-mga:
I keep adding stuff in there :-)

Next step will be to make a synthesis with something like:

```
[model filename]
 - [material name]  ([script filename])
    o [texture1 filename]
    o [texture2 filename]
    o ...
...
```

Bertram25:
Cool! :D Will test asap. #579 and #492 will be trivial to fix now.

Bertram25:
textures_unused.log:
Seems there is a bug on ack invocation.

```
[texture filename]: [files referencing it]

AdventurerBed.png:  ack v1.39 Copyright 1993,94 Ogasawara Hiroyuki (COR.)
usage: ack [-{e|s|j|c[c]}] [-{a|A|o<file>}] [-zCntud] [-{E|S}] [<file>..]
...
```

meshes_tree.log:
Seems the script is never told and no texture is listed as well.

```
[mesh filename]:
 - [material name] ([script filename])
     * [texture filename]

AdventurerBed.mesh
 - AdventurerBed ()
Adventurer.mesh
 - Adventurer ()
Anvil.mesh
 - Anvil ()
Axe1.mesh
 - axt ()
...
```

Tested on Debian Sid/Exp.

akien-mga:
Hm indeed I have ack 2.14 and may have used functionalities that were not available in Debian's ack 1.39. I'll try to identify which ones and do without them :)

Bertram25:
> Hm indeed I have ack 2.14 and may have used functionalities that were not available in Debian's ack 1.39. I'll try to identify which ones and do without them :)

Too bad Debian is lagging behind about this package, though.

Bertram25:
@akien-mga Wait! Debian has both an ack and a newer ack-grep package. I'll test all that now.

akien-mga:
> > Hm indeed I have ack 2.14 and may have used functionalities that were not available in Debian's ack 1.39. I'll try to identify which ones and do without them :)
> 
> Too bad Debian is lagging behind about this package, though.

I guess it's considered a core package like grep or awk and they have cold feet for new major versions in such software :-)
I found a nice description of what's new in ack 2.x, so I can already see that --ignore-file might be problematic: http://beyondgrep.com/ack-2.0/

Bertram25:
My bad, ack is a kanji code converter, ack grep is the core util.

akien-mga:
> My bad, ack is a kanji code converter, ack grep is the core util.

Hehe, indeed the kanji code converter probably won't do the trick ;-)
Is the binary still /usr/bin/ack (thus conflicting with the "ack" package from debports), or is it named differently?

Bertram25:
> Hehe, indeed the kanji code converter probably won't do the trick ;-)

Yeah, plus it doesn't tell much when called from scripts and command line, easily misleading, IMHO.

> Is the binary still /usr/bin/ack (thus conflicting with the "ack" package from debports), or is it named differently?

I'll tell you after a reboot. ;)

Bertram25:
> Is the binary still /usr/bin/ack (thus conflicting with the "ack" package from debports), or is it named differently?

Funnily, an ack-grep binary is added with an alias to ack. I'm testing the script atm. I'll test whether there is a conflict when you install both, though. There might be a bug. ;)

Bertram25:
Works well now. Every scripts behave as expected. :)

ack has a conflict dependency to ack-grep on Debian as a side note. This is IMHO a kinda funny way to handle it but at least the packages aren't broken.

I'll merge now.

Bertram25:
#492 and #579 can be dealt with easily now. I'd request to move the unused textures and potential models to the media repo, for instance in: models_unused and materials_unused/textures as a start.
The media repo can be cleaned up afterwards.

akien-mga:
> I'd request to move the unused textures and potential models to the media repo, for instance in: models_unused and materials_unused/textures as a start.
> The media repo can be cleaned up afterwards.

Sounds good. I'll be on it for 0.5.0, the less unused stuff we ship, the smaller the source tarball and binary packages :)

Bertram25:
Thanks :D

PR #585, opened 2015-03-24
Fix #541

### LINKED ISSUE #541: Marked tiles weird color
Since attenuation has been reduced, marked tiles have strange color when far from the keeper hand. I guess this issue was not visible before because attenuation was too high and we couldn't reach a distance where the tile color changed

odscreenshot_2015-03-12_203615_2

Bertram25:
The marked tiles didn't have emissive light, explaining what we saw.


hwoarangmy:
Testing right now. I will merge right now if it is ok ^^

hwoarangmy:
I confirm it is fixed

PR #586, opened 2015-03-25
Made the Plugins.cfg file installed under /etc/opendungeons on linux... + Icon on Windows.

### LINKED ISSUE #400: Engine config files should not be installed in the data path on Linux
Currently we install user config files (plugins.cfg and resources.cfg) in the OD_DATA_PATH, e.g. `/usr/share/games/opendungeons` on Mageia and Debian.

I just met an issue that this situation creates: plugins.cfg hardcodes the path to the OGRE plugin dir (on Mageia, it's `/usr/lib/OGRE` on 32bit and `/usr/lib64/OGRE` on 64bit). Data in /usr/share is normally arch-independent and for Mageia I packaged it as "noarch", i.e. it's included in an arch-independent opendungeons-data package. Since the latter is built on 32bit by convention, the path in plugins.cfg is `/usr/lib/OGRE`, so the 64bit version of OD segfaults.

There are two possible solutions for this:
- Put the "general" config files (by opposition to the gameplay config files from ./config that can be left with the rest of the data) in `/etc/opendungeons`, which would be the standard path for system-wide config files.
- Do not package such config files, but have them generated automatically and put in `~/.config/opendungeons` at run time if they don't exist yet (just like the ogre.cfg file).

If we opt for the first solution, it would also be nice to copy the files from `/etc/opendungeons` to `~/.config/opendungeons` at the first run (or when the user space files do not exist), so that OD can be configured per-user and not only by the system administrator in /etc.

Bertram25:
... based systems.

I also made it so the local file is preferred to the system one when found for developers.

The behaviour shouldn't change on Windows.

Fixes #400


hwoarangmy:
I confirm it still works with MinGW. I will let linux people test and merge if ok

akien-mga:
Code looks good, I'll test it now.

BTW, I saw this in ResourceManager.cpp:

```
#if OGRE_PLATFORM == OGRE_PLATFORM_WIN32 && defined(OD_DEBUG)
//On windows, if the application is compiled in debug mode, use the plugins with debug prefix.
const std::string ResourceManager::PLUGINSCFG = "plugins_d.cfg";
#else
const std::string ResourceManager::PLUGINSCFG = "plugins.cfg";
#endif
```

So basically the plugins_d.cfg file is only used for Windows?

Ah actually I just checked the commit again and it's not installed on Linux, it's just @Bertram25's commit log which is misleading ;-)

akien-mga:
The `make install` instructions work fine, but I get a crash on startup when I run the system installed opendungeons out of the source tree: https://framalab.org/zerobin/?8a7f864bd0d94ca5#MsPtCibeA+d+7m80ZI3hGbUMDYgBSZzNfs9GjD1C7iU=

akien-mga:
It works fine when running the system-installed OD from the source tree though, here is the log for comparison: https://framalab.org/zerobin/?bada443d0f042fc5#W6i3b8UwsC3YwOGxONQ8DB/Z91AZu9HRPgO+hoc8aq8=

akien-mga:
I've confirmed that it works fine if I run the system installed OD from the build dir after having removed plugins.cfg (it then loads the one from /etc/opendungeons/plugins.cfg, so the issue seems to be with the data path, and there probably not linked to this PR.

akien-mga:
Looks like ef8cbd82 is the culprit, if I readd `FileSystem=materials/RTShaderLib/GLSL` to resources.cfg it works fine.

akien-mga:
> Looks like ef8cbd8 is the culprit, if I readd FileSystem=materials/RTShaderLib/GLSL to resources.cfg it works fine.

This is not the proper fix though, since this commit added a code chunk that should choose between adding materials/RTShaderLib/GLSL or GLSL150 depending on the hardware. Maybe this code doesn't run early enough and Ogre tries to load shaders before that?

https://github.com/OpenDungeons/OpenDungeons/commit/ef8cbd82e06ca6df6b9c91eb74fe6ec1c7fcdb3f#diff-05031a8e2d7251a0c1de5cf2beea8890R104

oyvindln:
> This is not the proper fix though, since this commit added a code chunk that should choose between adding materials/RTShaderLib/GLSL or GLSL150 depending on the hardware. 

No, that folder should be in resources.cfg, most of the shader files are there. It's materials/RTShaderLib/GLSL/120 or materials/RTShaderLib/GLSL/130 that gets added depending on the system.

akien-mga:
> No, that folder should be in resources.cfg, most of the shader files are there. It's materials/RTShaderLib/GLSL/120 or materials/RTShaderLib/GLSL/130 that gets added depending on the system.

You're partly right, I was looking at a diff that was not the final version. Here it is: https://github.com/OpenDungeons/OpenDungeons/blob/development/source/ODApplication.cpp#L97-L113

It still looks like it should be adding the materials/RTShaderLib/GLSL folder to the resourceGroupManager.

And it seems to work, since it's in my log:

```
Added resource location 'materials/RTShaderLib/GLSL' of type 'FileSystem' to resource group 'Graphics'
Shader version is: 130
Added resource location 'materials/RTShaderLib/GLSL/130' of type 'FileSystem' to resource group 'Graphics'
```

But then:

```
Parsing scripts for resource group Graphics
Parsing script materials/HardwareSkinningShadow.material
High-level program Ogre/RTShader/shadow_caster_skinning_2weight_vs_glsl encountered an error during loading and is thus not supported.
OGRE EXCEPTION(6:FileNotFoundException): Cannot locate resource HardwareSkinningShadow_TwoWeights.glsl in resource group Graphics or any other group. in ResourceGroupManager::openResource at /home/iurt/rpmbuild/BUILD/ogre_src_v1-9-0/OgreMain/src/OgreResourceGroupManager.cpp (line 756)
```

oyvindln:
Maybe it has something to do with relative vs absolute paths then. I don't know what else would be the issue.

Bertram25:
> Ah actually I just checked the commit again and it's not installed on Linux, it's just @Bertram25's commit log which is misleading ;-)

Btw, why is the plugins_d.cfg file only used on Windows? (I can't remember)

> Looks like ef8cbd8 is the culprit, if I readd FileSystem=materials/RTShaderLib/GLSL to resources.cfg it works fine.

Should I readd it then?
I will also have a look at how the mPluginsPath ResourceManager member is used. It must be linked somehow.

oyvindln:
> Btw, why is the plugins_d.cfg file only used on Windows? (I can't remember)

In windows there is one debug library and one release library, the debug one having a _d suffix. In Linux, the debug info and the library files are separate, so there isn't a need for loading different libraries depending on whether debug or release is used.

Bertram25:
> In windows there is one debug library and one release library, the debug one having a _d suffix. In Linux, the debug info and the library files are separate, so there isn't a need for loading different libraries depending on whether debug or release is used.

Ah, of course. Thanks for the reminder. :)

akien-mga:
> > Looks like ef8cbd8 is the culprit, if I readd FileSystem=materials/RTShaderLib/GLSL to resources.cfg it works fine.
> 
> Should I readd it then?
> I will also have a look at how the mPluginsPath ResourceManager member is used. It must be linked somehow.

Readding it would be a workaround, but it would hide the current issue which is that even though materials/RTShaderLib/GLSL is added to the resource manager in the code, it is not working similarly to having "FileSystem=materials/RTShaderLib/GLSL" defined previously in resources.cfg

So if we add the GLSL directory manually, then I wonder whether the GLSL/{120,130} folder would be properly taken into account or whether we would run into the same issue (probably the latter). So the issue is worth looking into anyway IMO.

Bertram25:
Ok I've fixed it IMHO. (My tests proved an installation running fine.)

@oyvindln :
BTW, shouldn't this:
https://github.com/OpenDungeons/OpenDungeons/blob/development/source/ODApplication.cpp#L93-113

Be moved into:
void ResourceManager::setupOgreResources() ? I'll handle it in any case. EDIT: Done. :)

I've added the mGameDataPath on the resource path given in the snippet present in ODApplication.cpp. That's what fixed it.

Bertram25:
Good to go.

oyvindln:
> BTW, shouldn't this:
> https://github.com/OpenDungeons/OpenDungeons/blob/development/source/ODApplication.cpp#L93-113
> 
> Be moved into:
> void ResourceManager::setupOgreResources()

According to the comment, it shouldn't. But if it works, I guess it can.

Bertram25:
> According to the comment, it shouldn't. But if it works, I guess it can.

Well, I moved the whole function call between root::Initialize() and initialiseAllResourceGroups() and it's working on my end. :)

hwoarangmy:
Seems like I should test this one after all ^^
I will try tonight

akien-mga:
Tested locally and it works fine, both in source and system wide, with my intel GPU (glsl130) and my nvidia GPU (it says shader 450, not sure what that corresponds to).

Bertram25:
@akien-mga Done. :)

Bertram25:
> (it says shader 450, not sure what that corresponds to).

The latest definition available atm. Must be linked to GL 4.1+

Bertram25:
@akien-mga This means you may be able to play Bioshock Infinite on Linux ;)

Bertram25:
@hwoarangmy I've added the windows part of #202 Feel free to tell whether there is something wrong. :)

@akien-mga What about the linux part, isn't it done already?

akien-mga:
> This means you may be able to play Bioshock Infinite on Linux ;)

\o/ Yeah I made sure to get a laptop with a middle to high-end GPU when I bought it 1½ years ago, so I should be able to play recent AAA games for the next couple of years at least (maybe not with the highest settings on in two years though ;-)).

Bertram25:
Ok, then our next resort is to look at how SDL does it, I guess?

akien-mga:
I guess you mean SFML? We're not using SDL AFAICT :)

akien-mga:
For me this can be merged, but most likely @hwoarangmy should test first the part about the windows icon.

hwoarangmy:
yep, I will tonight

hwoarangmy:
Testing now

Bertram25:
I just pushed, hope you had it.

hwoarangmy:
> I just pushed, hope you had it.

No problem, I was only running cmake ^^ I try now

Bertram25:
Cool :)

hwoarangmy:
Should I merge if ok?

Bertram25:
Yep. :)
I'll deal with the linux part (if it's feasible) in a second step.

hwoarangmy:
Works well ^^

PR #588, opened 2015-03-26
Make portals claimable + max creature number increases with portals + small fixes

hwoarangmy:
To test, the skirmish small test map (2v2 and empty) now start for seat 1 with 1 rogue portal that needs to be claimed. Note that the max creature number should increase when the portal is claimed (5 in 2v2 map and 10 in empty map).
I've not tested with 2 portals but it should work ^^

### DIFF
--- a/config/global.cfg
+++ b/config/global.cfg
@@ -96,7 +96,10 @@
 [GameConfig]
     NetworkPort	31222
     CreatureDeathCounter	30
-    MaxCreaturesPerSeat	20
+# Maximum creature number. This is used for lagging purpose and a seat cannot control more creatures
+    MaxCreaturesPerSeatAbsolute	30
+# Maximum creature number a seat can control without portal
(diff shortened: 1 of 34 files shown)


Bertram25:
I've got a few questions but overall it's worth being integrated. :) I'll test asap.

Bertram25:
Code looks good. Once it's tested, let's merge. :)

akien-mga:
It works great :) I've confirmed that you can even get your portal claimed by your ally if you prevent your own worker from claiming it. The AI does not focus on claiming portals for now though, so if you leave them be your worker will claim it first - and anyway that's not an issue because the setting that we have in the 2v2 map is only for testing :)

akien-mga:

akien-mga:
Four pull requests at once! We're only missing @Bertram25's ;-)

Code has been reviewed and I tested it live, so I'll merge this one.

PR #591, opened 2015-03-27
Remove unused models, materials and textures

### LINKED ISSUE #579: Spring cleaning in the materials/ folder

akien-mga:
As discussed extensively in #579.

This needs of course to be well tested to make sure it does add regressions if I removed textures that were needed in other materials, but normally I was careful.

I still have to check the logs for potential warnings or errors.


akien-mga:
For the reference, comparing gz-compressed archives of the source repo we go from 89 MB before this PR to 76 MB after, so that's quite some space saved for distribution and installation.

akien-mga:
Note that I kept local copies of all removed meshes and will make a PR to the media-source repo once we've settled on this one.

akien-mga:
I've compared the logs between my branch and the current development branch, and I don't see anything new introduced by this PR, apart from the scripts that I removed and that are therefore not parsed when starting OD.

For the reference, to be able to use `diff` to compare two logs, you need to remove the time info with something like `sed -i 's/^.\{10\}//' ~/.local/share/opendungeons/opendungeons.log{,.1}`

akien-mga:
I readded one of the materials in 54d21cf to fix a merge conflict with @Danimal696's PR, since I don't want him to panic over conflict resolution ;-)
Therefore I'll ask that my PR is merged first.

oyvindln:
This seems to work fine for me at least, merging.

PR #592, opened 2015-03-27
Abstracted logmanager for unit tests

oyvindln:
Made logmanager abstract
Started adding unit test for creaturedefinition


hwoarangmy:
The code seems good to me. The requested changes can be done later if you prefer. Just a few questions to answer ^^

oyvindln:
Added some fixes. I will fixe the merge conflict tomorrow.
Also, I found that the compiler happily accepts operator>> on a forward-declared pointer type, which caused me a bit of confusion. We should be wary of using and defining operator>> directly on pointers, and probably prefer references instead.

hwoarangmy:
Code seems good to me. I will test tonight with windows ^^

hwoarangmy:
Testing right now

hwoarangmy:
Works well. I merge

PR #593, opened 2015-03-28
Fixed materials - added/created normals- renamed multiple texture PNGs t...

Danimal696:
...o follow capital naming convention

Up to Portalvortex, tomorrow more ;-) , i hope there are no
discrepancies with Akien


akien-mga:
Awesome! I was about to suggest to rename some textures to make it easier to see which texture belongs to which material/model, but you did it yourself :)

I think the only conflict with my PR is that we both removed  materials/scripts/BeetleGolem_mat.material, but I can edit my branch to readd it (so that it won't conflict with yours).

akien-mga:
I tested it in game and there are some issues:
- Cavehornets have yellow wings :)
  cavehornet
- The gold full tile is now animated :-o I guess the normal is slightly offset and therefore it looks like the texture is moving when turning around it with the cursor (thus changing the lighting). It does not do it when the tiles are marked for claiming.
- The gold ground is not fully adjustable with the dirt ground tile, is it because of the normal?
  goldground
- Marked claimed walls are really bright, I don't know if that's intended:
  markedclaimed
- The (uncrafted) boulder is quite bright:
  boulder
  On the other hand it looks good once crafted and installed:
  boulder1
- Some models in the crypt are dull looking:
  crypt
  crypt1
- The portal vortex is now red. At first I thought it would be of the seat's colour (which could be nice), but it looks like all portal vortexes are red.
  vortex

I guess it could be nice to have a test map with all rooms built and all creatures spawned, to be able to quickly check all models.

Danimal696:
Lets go bit by bit:
-Dwarf 3 new name, i think it can be done without reexporting by writing the mesh file?
-Cavehornet wings- ill check, probably my fault
-Animated gold, actually that happens with all fixed models, but tiles are more visible, what should i   do, no normals for gold?
-gold ground, its because its reflecting and emiting more ligth than common dirt, ill make it duller
-Marked walls, i just renamed its textures, i dont think that should have that effect?
-The (uncrafted) boulder - something to do with the transparency code?
-Statues, they need more work, at tleast they are not metallic shiny anymore
-portal vortex - yes, red for bad guys, but the team colour sound interesting, can it be done by code? if not i dont know how (same as new bridge)

Danimal696:
Fixed cavehornet and ground gold, the other issues still remain

oyvindln:
The tile marking colour is not from this PR. I get it too without this. I think it's from the earlier pr @Bertram25 to fix marked tiles far away from lights not being shaded. The boulder is probably due to opacity, as @Danimal696 has mentioned.

oyvindln:
> -Animated gold, actually that happens with all fixed models, but tiles are more visible, what should i do, no normals for gold?

The normals for dirt tiles are not really the correct texture, maybe making a new normal that follows the texture would look better. You may also want to check if the normal maps are in object or tangent space, if it's defined the wrong way in the material file that may also cause it too look weird.

akien-mga:
> Fixed cavehornet and ground gold, the other issues still remain

Yep, I confirm the fixes.

Two new issues:
- Wizard:
  wizard
- SpiderBed:
  spiderbed

I've made a test level that shows most models available in game, it's in my branch for now and in a PR, but you can get it here and put it in levels/skirmish to use it. It's pretty useful to spot regressions or see which materials would need more/less lighting: https://raw.githubusercontent.com/akien-mga/OpenDungeons/3b9854814f25ca96dbdff87f67f77e70b2af3951/levels/skirmish/TestModelsFair.level

Danimal696:
> The normals for dirt tiles are not really the correct texture, maybe making a new normal that follows the texture would look better. You may also want to check if the normal maps are in object or tangent space, if it's defined the wrong way in the material file that may also cause it too look weird.

The normal is custom made for gold, so thats not the problem, can you check the orientation on the materials, please? i dont know how to.

Im reverting wizard, no point to put lots of work into a model that will dissapear
My mistake into SpiderBed

akien-mga:
For the Spiderbed and SpiderBedWeb there is still an issue (supposedly only on Linux as it's a case-sensitiveness issue).

The texture is named "Spiderbed.png" but you use "SpiderBed.png" in the materials files, you should probably rename the texture to SpiderBed.png.

akien-mga:
So apart from that, there are only two new issues linked to this PR:
- The boulder trap does not take into account the opacity parameter set in RenderManager::setMaterialOpacity. @hwoarangmy, @Bertram25, @oyvindln Any clue about this one? Is something wrong in the new format of Boulder.material, or is the issue in the way the code handles (or does not handle) normal maps?
- The "animated gold" issue discussed above, that indeed seems to happen for more models than just gold, it's just more visible on gold. By looking closely to claimed walls you'll also see that it seems to be moving with the light. I can't tell if it's an optical illusion or if their is really a normal offset issue.
  For the gold normal I noticed that there is indeed a 1px offset issue, but even after fixing it, the gold still seems to be moving.

akien-mga:
> The "animated gold" issue discussed above, that indeed seems to happen for more models than just gold, it's just more visible on gold. By looking closely to claimed walls you'll also see that it seems to be moving with the light. I can't tell if it's an optical illusion or if their is really a normal offset issue.
> For the gold normal I noticed that there is indeed a 1px offset issue, but even after fixing it, the gold still seems to be moving.

I've been playing a bit with the textures and I confirm that there are some distortion effects in the normal textures, the pixels do not match the ones of the basic texture 1:1. For example I took ClaimedWall.png, then did some colour and lighting modifications in GIMP, and exported it as ClaimedWallNormal.png: my claimed wall does not look as good as the current one since I'm bad at making normal maps, but at least it does not move anymore with light. Same thing with the gold walls.

So there seems to be something wrong with the way you create normal textures @Danimal696. You should check your script/filter and make sure that it does not distort the original image.

For example this one does not create the "animated gold" bug:
goldnormal
It looks like crap in game though, I don't know how to do a normal map :-P It does exhibit some lighting issues depending on the orientation of the tiles though :)

Danimal696:
surely you made it based on alpha channel rather than color height, the results are pretty bland and almost invisible... Im at a loss here, good-looking but moving or barely looking but in-place? Ill vote for leaving them as if, and waiting for the new ogre to fix it. Its nothing so noticiable imo and adds a lot eye candy that would be lost.

akien-mga:
> surely you made it based on alpha channel rather than color height, the results are pretty bland and almost invisible... Im at a loss here, good-looking but moving or barely looking but in-place? Ill vote for leaving them as if, and waiting for the new ogre to fix it. Its nothing so noticiable imo and adds a lot eye candy that would be lost.

You misunderstand me, this is not an OGRE issue. The issue is that the motifs on your normal maps are not well aligned with the pixels of the original texture, so the motifs are "moving" when the normal map is applied. My example looks bland but it was just given to show that it works fine if the motifs are not altered, but only the colour/saturation/luminosity.

Can you tell me how you create normal maps so that I can try creating a good looking normal map where the motifs would not move?

Danimal696:
Normally by photoshop plugin, those are the settings i commonly use; animate ligth is for the real time rigth display
normal-maker

akien-mga:
I found a clone of your plugin for GIMP [1] so I gave it a try and generate my own normal map for the gold texture, and I get the same output as you do. So I guess this "offsetting" is somehow related to the nature of normal maps.

I've also looked closely at some other materials, and indeed this "animation" bug happens for most of those that have normal maps. It's just more visible/annoying for the gold due to its texture, when it's moving it looks like lava more than gold. IMO we can merge this change and see with the rest of the devs what can be done about it; there are other problems linked to the gold (the tileset does not fit the various orientation needs).

So IMO once you've fixed the Spiderbed.png texture filename, we can merge this one and open bug reports for the boulder issue and the normals issue.

[1] https://code.google.com/p/gimp-normalmap/

oyvindln:
It looks like the plugin makes tangent space normal maps, so try changing "object_space" to tangent_space" in the material file and see if it looks better. (I don't know if all the current normal maps are though.) Also note that normal mapping doesn't give perfect results, it does result in some visual distortions particularly when viewing from very acute angles. Normal mapping only adjusts the lighting to fake a depth effect, it doesn't actually adjust any vertices.

akien-mga:
> It looks like the plugin makes tangent space normal maps, so try changing "object_space" to tangent_space" in the material file and see if it looks better.

I've tried that but it doesn't really look better.

@Danimal696 I found this software that might interest you btw: https://github.com/Theverat/NormalmapGenerator/releases
It seems to be able to do the same as the Nvidia normal map plugin, but also handles specular/displacement/ambiant occlusion maps and does batch processing.

Danimal696:
looks nice, ill try, ill make the last fix later

oyvindln:
I think the "animated gold" is just visual distortion from the normal map process, rather than anything wrong. Maybe the light is too strong. It looks more pronounced when the texture has a normal map that corresponds to the texture, than when using a normal map that is different from the texture, like on the dirt tiles.

As for the transparency, I'm not sure, there may be some material settings that has to be set right for it to work correctly.

akien-mga:
> I think the "animated gold" is just visual distortion from the normal map process, rather than anything wrong. Maybe the light is too strong. It looks more pronounced when the texture has a normal map that corresponds to the texture, than when using a normal map that is different from the texture, like on the dirt tiles.

That sounds sensible. I just did a normal map for the dirt based on the actual dirt texture, and indeed the distortion is then more visible. The dirt actually looks way better when using the wrong normal map as currently :)

oyvindln:
I think the opacity issue is a bug in the code, it only sets opacity for the latest pass, and since the normal mapped material has more than one pass it ends up looking wrong.

oyvindln:
I've tried various things, but I can't get multi-pass rtshader normal map materials to appear transparent, so maybe we should drop normal maps on the traps for now. Single-pass normal maps seem to work, but that may cause problems if there are too many lights.

Danimal696:
Akien, the png file was already fixed, can you have a look and confirm?

Danimal696:
so do i restore the boulder amterial for the time being?

Danimal696:
My god, close and comment button are too near each other...

oyvindln:
> so do i restore the boulder amterial for the time being?

Yeah let's do that for now.

akien-mga:
> Akien, the png file was already fixed, can you have a look and confirm?

It's not fixed because the materials reference SpiderBed.png, but the texture is named Spiderbed.png. On Windows it doesn't matter, but on Mac and Linux the capital B is important.

So you should rename materials/textures/Spiderbed.png to SpiderBed.png.

akien-mga:
I'll merge and fix the texture name in another PR.

PR #594, opened 2015-03-28
Added RoomPortalWave that spawns creature waves

hwoarangmy:
Note that the portal will automatically mark tiles to dig if a worker is in the same seat as the portal (or it is spawned by it).

Also note that the AI uses GameMap::path to choose tiles to dig. That means that it will always go straight to the wanted dungeon temple if it can (ie if it is not perfectly walled). Moreover, if a dungeon is perfectly walled, the AI won't try to attack it at all (the expected behaviour would be to dig around the walled tiles.
This will probably need to find another algorithm than GameMap::path but I don't want to mess with it for now. Maybe next week. For now, it is more than enough ^^

To test the portal, as always, you can try the 2v2 skirmish smell test map. Beware, the portal is near and spawns pretty powerfull creatures. If you don't wall soon enough (or you don't claim the portal at start), you are done ^^


akien-mga:
Works great locally :-) Those big dragons are fearsome :-O
Is it intended that the PortalWave uses one of the crypt's tomb models?

hwoarangmy:
> Is it intended that the PortalWave uses one of the crypt's tomb models?

Yes. It will usually be 3x3 but it can have any size as the other portals. But to make the difference between normal portal and wave portal, I've used one of the crypt mesh.

hwoarangmy:
Rebasing

hwoarangmy:
Rebased

oyvindln:
This PR makes tile selection a bit laggy for me, there is a delay between selecting tiles, and them being marked. Otherwise it seems to work fine.

hwoarangmy:
I probably messed something while merging the tile selection function for players with the tile selection for portals. I will have a look tonight

hwoarangmy:
@oyvindln As you have seen, there was a bug in tile marking. It was even worst than what you have seen since tiles not claimable on the server were not marked on client ^^
It is fixed now

oyvindln:
Works fine now.

PR #597, opened 2015-03-29
New test level to show models + template update

akien-mga:


hwoarangmy:
This works well. I would say that 2x2 tiles per creatures would be better to better see them but it is already good. I merge

PR #601, opened 2015-03-30
Chat and event handling overhaul + Misc fixes and cleanups.

Bertram25:


akien-mga:
I've tested the PR and it's working great :-D
Here's an example:
chat

Only issue I noticed so far is that the scrollbar does not work as expected when there are more lines to display than the height of the widget:
scroll
It's not possible to scrollbar by dragging the slider or by clicking on the top arrow.

akien-mga:
The small bug I reported above is not blocking IMO, feel free to merge this one and we'll open a new issue for the non working slider.

akien-mga:
Merging, as discussed above this is a first (and big) step, if things need to be improved this can be done later on.

PR #605, opened 2015-04-01
fixed grimoire animation

Danimal696:
since i coulndt test it ingame before it looked very ugly


akien-mga:
I confirm it looks better now :-)

PR #607, opened 2015-04-01
Improved path finding algorithm for portal waves

hwoarangmy:
To test it, you can use the same seat id for the portal wave as for the player. This way, you will see the marked tiles.

The used algorithm is closer to what would do a human player: going around the claimed walls.
Note that the choosen path will not always be the shortest.


Danimal696:
That is pretty cool :)

akien-mga:
I got a segfault with this PR:

```
Computing turn 71
ERROR: Assert failed file /home/akien/Projects/opendungeons/OpenDungeons/source/game/Seat.cpp line 1876 info : Tile=[15,4]
signal 11 (Segmentation fault), address is 0x168 from 0x4dc24a

[bt]: (1) ./opendungeons : Seat::isTileDiggableForClient(Tile*) const+0x4a [0x4dc24a]
[bt]: (2) ./opendungeons : Seat::isTileDiggableForClient(Tile*) const+0x4a [0x4dc24a]
[bt]: (3) ./opendungeons : Player::markTilesForDigging(bool, std::vector<Tile*, std::allocator<Tile*> > const&, bool)+0xb4 [0x4d20a4]
[bt]: (4) ./opendungeons : RoomPortalWave::handleDigging()+0x370 [0x5d8450]
[bt]: (5) ./opendungeons : RoomPortalWave::doUpkeep()+0x33d [0x5d93dd]
[bt]: (6) ./opendungeons : GameMap::doMiscUpkeep()+0x245 [0x4f9815]
[bt]: (7) ./opendungeons : GameMap::doTurn()+0x7e [0x4f9a7e]
[bt]: (8) ./opendungeons : ODServer::startNewTurn(double)+0x359 [0x5894f9]
[bt]: (9) ./opendungeons : ODServer::serverThread()+0x27c [0x58a8cc]
[bt]: (10) /lib64/libsfml-system.so.2.2.0 : +0x9478 [0x7f35a4ae6478]
[bt]: (11) /lib64/libpthread.so.0 : +0x75bd [0x7f35a32b35bd]
[bt]: (12) /lib64/libc.so.6 : clone+0x6d [0x7f35a3ddb5ed]

signal 11 (Segmentation fault), address is 0 from 0
```

Note that I had disable the FOW with the "togglefow" command, to be able to see the portal wave creatures. The rogue creatures had a dwarf worker unclaiming my tiles btw and claiming them for seat 0, if that might be related to what the backtrace shows.

akien-mga:
I just got the same segfault with using "togglefow", just starting the 2v2 small test map.

hwoarangmy:
Yes, I've seen and fixed that (and other stuff too). In any case, from the tests I've done yesterday, I've seen a case that do not work with the current algorithm. I had no time to investigate but we shouldn't merge for now.

Danimal696:
you could create a "tunneler dwarf" with 0 claiming instead of using the dwarf worker, anyways, its not a real solution for the problem

hwoarangmy:
No, it has nothing to do with the dwarf claiming tiles. It is because of my mistakeof calling client functions with portal (that can be owned by an AI). Because I did my last tests using the player seat to see which tiles were marked, I didn't see the bug. But it's not a big deal and I have already fixed it.
Changing the path finding algorithm is a little bit more complicated. Once I will be done, I will update this PR

oyvindln:
Is this segfault from this PR, or is it from earlier? I get the same thing without this PR (client functions being called for ai players and failing because mTilesStates is empty). I've done some changes locally, and while it could be me having done something wrong,  I haven't really touched this area.

hwoarangmy:
AFAIK, it is from this PR. Do you have a backtrace of such a segfault ?

oyvindln:
I get the same backtrace as akien. Using the map Test map small (2 vs. 2).

oyvindln:
I get the same crash when the current development branch without my changes, so it's probably from the portal wave PR.

hwoarangmy:
@oyvindln I just had a look and you are right, it comes from my previous PR (sorry, i couldn't check before). It is fixed now (it was in Player.cpp).
This PR should be mergeable now if nobody sees a problem ^^

Bertram25:
The overall change looks fine, even if I must say I spent 1/2h reading the new stuff :D 

I've got a question though, isn't it more a WavePortal than a PortalWave? For the room has got its own mesh anyway, right?
Also, as I've told somewhere else, do you need an icon to be able to place a WavePortal from the editor? (for 0.6.0 maybe)

EDIT: Testing the new level right now. :D

hwoarangmy:
> isn't it more a WavePortal than a PortalWave?

I usually prefer to concatenate things from the most important to the less (so that they are sorted alphabetically). In this case, portal seems more important than wave. For example, if we create a portal that does something else, if we call PortalUnique, it will be next to the other portals. If we call them WavePortal and UniquePortal, they will be split.

> do you need an icon to be able to place a WavePortal from the editor?

If you have something in mind, please do not hesitate ^^

Bertram25:
I love that one:
http://game-icons.net/delapouite/originals/dwarf-face.html
But maybe one of those ones will do? ;)
http://game-icons.net/lorc/originals/guarded-tower.html
http://game-icons.net/lorc/originals/stone-tower.html

hwoarangmy:
I don't like the guarded tower. But the dwarf face and the stone towers are good IMHO. I vote for the dwarf ^^

Bertram25:
Btw, I tested it and died surrounded by level 15 creatures :E Is good to be merged IMHO. Ah, and absolutely no crashes at all.

hwoarangmy:
I think the level will be pretty hard (maybe not possible to win at all). We will see. From the tests I've seen, the enemy dungeon gets pretty strong creatures from the start and the hero waves are also pretty powerfull with time.
We will balance it if needed later ^^

Bertram25:
> I think the level will be pretty hard (maybe not possible to win at all). We will see. From the tests I've seen, the enemy dungeon gets pretty strong creatures from the start and the hero waves are also pretty powerfull with time.
> We will balance it if needed later ^^

Np, it was fun. :D Let's balance all that at the end. Research and creature spawn conditions need balancing too. All that is more mid-term effort, anyway. :)

Bertram25:
I'll merge. It's working fine. I don't think I'm taking a big chance here. ;)

PR #609, opened 2015-04-01
New model for workshop

Danimal696:
It has an animation called "Loop", for the walls we can use the anvil and whetstone at ramdon, i think they fit both rooms


akien-mga:
Nice! I guess it is meant to be used together with the current model, chosen randomly as for e.g. the library or the training room?

Danimal696:
yes, its a different kind of workbench
cuttingbench2p

PR #616, opened 2015-04-04
More work towards unit tests + minor cleanups

oyvindln:
I found that the most straightforward way to be able to create unit tests using the existing classes was to split up the functions in two files. This allows us to link to only the functions which doesn't have too many dependencies until the code is made more testable. I've done this for seat, tile and player now, so that tiles can be tested. Hopefully I didn't mess up anything when fixing merge conflicts.

Some request to make the code more testable:
- Try to make objects abide by the [Single responsibility principle](https://en.wikipedia.org/wiki/Single_responsibility_principle), i.e objects should only be responsible for one thing.
- Don't make functions member functions unless access to class internals or virtual dispatch is needed. 
- Don't use singletons, send in what the function needs as parameters instead. (Except for ogre stuff, as the API leaves no choice.)
- Try to use as parameters only what a function needs, e.g., don't do things like this unless you have some good reason for it: 

``` c++
int x function(int x, bool y, MassiveObjectWithLotsOfMethods* m)
{
      ... //code that doesn't use m
      if(m->isX());
      ... //code that doesn't use m
}
```
- Don't create half-initialised objects. Objects should be fully initialised when created. If creation can fail, a factory function can be used.

Other changes:
- Moved the player selection info out of the player class as it's only used in the gui.
- Removed an unused old spell class that for some reason still existed.
- Gave the player more gold and research on 2 of the test levels.


hwoarangmy:
There are many interesting things here. However, there are also lots of changes. Do we want to integrate it before or after the 0.5.0 ?

akien-mga:
If there is no direct influence on the gameplay and the performance, I'd say we should merge it after 0.5.0, to avoid potential regressions.

oyvindln:
There isn't much in the way of end-user changes so it can wait for after 0.5.0 yes.

Bertram25:
The changes are quite big @oyvindln, it's an impressive amount of work. I'll happily read all of them with time. In the meanwhile, and regarding the next release, should I create a 0.5 branch now or after @akien-mga latest PR?

Best regards :)

Bertram25:
Btw, while we're speaking about the SR principle, I guess it was mainly due to the current code state of entities and the gamemap?

I'd still like to add one thing about that principle though: Defining each object responsibility area is an everchanging thing and can sometimes be somewhat arbitrary from experience.
What I'd like to point out is that I'd more have us focus on taking care about having objects without intertwined roles than splitting objects for the sake of pure conceptuality even if you're right imo about pointing it out.
Nice point about factories, too :)
I'll make my usual comments about the PR tomorrow. ;)

Best regards,

oyvindln:
> Btw, while we're speaking about the SR principle, I guess it was mainly due to the current code state of entities and the gamemap?

Yes. GameMap in particular. Part of the problem is that it is included almost everywhere, and the header is massive, which slows down compile times (I have a slow computer), and makes it hard to create tests using it. Many of the functions in the object don't really need to be there, as they only call public functions anyway. There are several which doesn't even use the GameMap object.

Another problem with both is that they are not very cleanly initialised, which was quite confusing. E.g, for the creatures, some of the setup that was required for the object to work on the server side was inside the "createMesh" function.

> What I'd like to point out is that I'd more have us focus on taking care about having objects without intertwined roles than splitting objects for the sake of pure conceptuality even if you're right imo about pointing it out.

Yeah, I agree, I'm not saying we should immediately focus all efforts on splitting up all objects and be super strict about it, just that it should be kept in mind when adding new functionality (I'm guilty of being too lazy with this too). I am planning to do a bit of cleanup though, particularly of GameMap.

The changes are not really all that large, most of it is just splitting the implementation of certain objects into two files.

oyvindln:
Also, I think some of this will come in more naturally if we get more into a habit of writing tests for new functionality.

hwoarangmy:
> For the creatures, some of the setup was inside the "createMesh" function which .

It was done in createMesh because of the default worker class and the HP that can be set to "max" which has to be computed only once stats are built. To be honest, I'm curious to know how you could get ride of this (I would have put the code elsewhere if I had seen a better way).
I will see tomorrow if I can ^^

oyvindln:
Yeah I could see why it was put there, it was just very confusing because I thought that createMesh wouldn't need to be called on the server ^^. I already moved it out of there in this PR.

Bertram25:
@oyvindln I'll review your PR asap. Sorry to keep you waiting. :)

Bertram25:
The stable and 0.5 branch have been created, we can resume with crazy new features. ;) And review this PR as a nice side-effect.

hwoarangmy:
I tried this PR but strangely, when I rebase main stream, I get a conflict even if this PR is said mergeable...

oyvindln:
Maybe it will work better when I have rebased.

oyvindln:
rebased

hwoarangmy:
I have a segfault in the tile test case. Which happens because there is no gamemap (which is something most entities will segfault if created with nullptr gamemap). Honestly, there is a huge amount of work in this PR but I'm not sure testing without gamemap is really relevant. Even if we patch everywhere to test if gamemap is null, the behaviour will be so different between normal use and test use that the test will only be usable for very basic stuff and, thus, not really usefull. IMHO, the best would be to try decoupling the rendermanager and everything else to allow testing without ogre. But testing without gamemap is not relevant enough considering the amount of work required.

oyvindln:
I'm not intending to test without gamemap. It's just not gotten to the point where it's decoupled enough yet. I didn't notice the tile test was failing, I must have forgotten to check all the tests.

Bertram25:
@oyvindln Can you join us on: :) http://webchat.freenode.net/?channels=opendungeons

source/gamemap/GameMap.cpp
```diff
@@ -905,68 +907,122 @@ void GameMap::createAllEntities()
...
-    for (MapLight* mapLight: mMapLights)
+        for (Trap* trap : mTraps)
+        {
+            trap->updateActiveSpots();
+        }
+
+        for (Creature* creature : mCreatures)
+        {
+            //Set up definition for creature. This was previously done in createMesh for some reason.
+            creature->setupDefinition(*this, *ConfigManager::getSingleton().getCreatureDefinitionDefaultWorker());
```
hwoarangmy, line 944: Why are you calling setupDefinition so many times ?

source/gamemap/GameMap.cpp
```diff
@@ -905,68 +907,122 @@ void GameMap::createAllEntities()
...
+        for (RenderedMovableEntity* rendered : mRenderedMovableEntities)
+        {
+            rendered->createMesh();
+            rendered->setPosition(rendered->getPosition(), false);
+        }
+
+        // Create OGRE entities for the creatures
+        for (Creature* creature : mCreatures)
+        {
+            creature->setupDefinition(*this, *ConfigManager::getSingleton().getCreatureDefinitionDefaultWorker());
```
hwoarangmy, line 973: and again for client ?

source/entities/Creature.cpp
```diff
@@ -155,6 +160,8 @@ Creature::Creature(GameMap* gameMap, const CreatureDefinition* definition) :
 
     if(mDefinition->getWeaponSpawnR().compare("none") != 0)
         mWeaponR = gameMap->getWeapon(mDefinition->getWeaponSpawnR());
+
+    setupDefinition(*gameMap, *ConfigManager::getSingleton().getCreatureDefinitionDefaultWorker());
```
hwoarangmy, line 164: Couldn't we find some way to call it only once ?

source/entities/Creature.cpp
```diff
@@ -559,6 +519,7 @@ void Creature::importFromPacket(ODPacket& is)
         mWeaponR = getGameMap()->getWeapon(tempString);
         OD_ASSERT_TRUE_MSG(mWeaponR != nullptr, "Unknown weapon name=" + tempString);
     }
+    setupDefinition(*getGameMap(), *ConfigManager::getSingleton().getCreatureDefinitionDefaultWorker());
```
hwoarangmy, line 522: Creatures loaded from stream will be loaded before calling GameMap::createAllEntities. That means that calling it here may not be required (not sure, need testing).

hwoarangmy:
After looking a little bit deeper into setupDefinition, I don't see problems with your implementation (I've not tested but I guess you did).
The only thing that may have been forgotten is the console command that allows to create a creature from a string. However, I'm not really sure this command is really usefull anyway. It is an old command from the time when the map was to be created by console command. IMHO, if we really want to let it, we should simplify by asking only the class and spawn position. Otherwise, the complete line is too complex (not talking about the fact we cannot copy/paste).

IIRC, the reason why I put the code into createMesh was to avoid calling this setup from too many places. But after all, it is probably better as you did. With a bit of luck, we can even remove some of the setupDefinition calls.

source/gamemap/GameMap.cpp
oyvindln, line 944: It's called once for each creature, as it has not been called for creatures that have been loaded from the map. For creatures spawned later it's done in the constructor.

source/entities/Creature.cpp
oyvindln, line 164: Ideally it would be always done by the constructor, but I couldn't get that to work for the map loading when I tried.

oyvindln:
> The only thing that may have been forgotten is the console command that allows to create a creature from a string. 

That command was already broken I think. It calls getCreatureFromStream which creates an empty creature object (disregarding the stream).

hwoarangmy:
> That command was already broken I think. It calls getCreatureFromStream which creates an empty creature object

You are right, the call to importFromStream is missing. Anyway, without an easy way to copy/paste, I don't think this command will ever be used...

source/entities/Creature.cpp
hwoarangmy, line 522: I misread. You already don't call setupDefinition in importFromStream. It would be nice to avoid this call here but I don't think it is possible...

hwoarangmy:
I've tested and seen no problem. The code looks fine. IMHO, we should merge this PR as soon as possible because it will conflict with nearly all commits we might do. There are nice things here and, if we miss some regressions, we can always fix them later.

@Bertram25 feel free to merge when you are ok

oyvindln:
So is this fine to merge now, @Bertram25 ?

Bertram25:
> So is this fine to merge now, @Bertram25 ?

Yes, now it is. :) Yet, you'll have to rebase once again, sorry about that.

oyvindln:
Rebased.

Bertram25:
I quick tested it again after the rebase to make sure. It is working fine. I merge :)

Issue #615, opened 2015-04-04
Alpha value for lighting parameters is often wrong, creatures change opacity when getting superposed


akien-mga:
The recent changes on materials introduced another regression (I guess I was a bit quick to merge them without checking the syntax), as the 4th setting for ambient, diffuse, specular and emissive should be the alpha parameter (i.e. transparency/opacity), but @Danimal696 often used the exact same value as for the first three parameters (red, green, blue).

akien-mga:
Unless otherwise told, I'll be on setting 1.0 everywhere for this alpha parameter.

Danimal696:
please do it, my sanity is going down with each new material issue

Bertram25:
Yes, 1.0 is the default value anyway. Thanks for catching up this one.

akien-mga:
Reference doc: http://www.ogre3d.org/docs/manual/manual_16.html

PR #612, opened 2015-04-02
New workbench

Danimal696:
i messed up my repository so i had to close the last commit, this should
be fine

PR #617, opened 2015-04-04
Some more work on the material scripts

akien-mga:
As discussed in #612 and #615, I've tried to correct some mistakes introduced previously due to a wrong understanding of what the arguments mean for the ambient, diffuse, specular and emissive lighting parameters.

I enforced the default values for those that were not already custom, and used a minimal shininess value of 2.0 for the specular parameter, since Ogre advises against uses values ranging from 0 to 1.

Some values were also rounded to decimals with one (or two in some cases) significant number, and I tried to fix the indentation (with a "tabulation" of 4 spaces).


hwoarangmy:
Nice cleaning. I will have a look if this fixes the segfault when some creatures are removed

akien-mga:
BTW I added the name of the material as a comment in the files where it was missing, because it's what my text editor seems to rely on to choose the syntactic colouring ;-)

akien-mga:
Some more cleaning coming soon, I'm trying to put even more order in the filenames (@Danimal696 already did a great job in his earlier PR, I'm just finishing the job).

hwoarangmy:
> (@Danimal696 already did a great job in his earlier PR, I'm just finishing the job

No worries, I don't think he will get angry at you if you spare him having to deal again with materials ^^

Bertram25:
Impressive :D This an awesome cleanup! :+1: 
For my own culture, I guess you turned the materials into XML, changed the material used there and turned them into mesh again, or did you use blender?

I'll test now.

Bertram25:
Looks fine after a full game played with several AI. IMHO good to merge if it's ok on Windows.

hwoarangmy:
testing right now

hwoarangmy:
> For my own culture, I guess you turned the materials into XML, changed the material used there and turned them into mesh again, or did you use blender?

I guess he did since AFAIK, OgreMeshUpgrader do not work for him. But note that it works for me (and maybe for you). That's easier to use.
You may already know but if you never tried, when you export from blender, you get an xml file. Then, you need to turn it into a mesh using OgreXMLConverter.
BTW, if you want to play with it, there is a very nice soft called "ogre meshy" that allows to open mesh files, see how they look like, play their animations...

hwoarangmy:
I got no error in the logs while opening the Model's fair. Merging

Bertram25:
Yeah, even if I did know about those tools, I must say I didn't play much with the XML content. :)
The OgreMeshUprader is broken for me too on meshes and last time, we already had to turn those back to xml first.
I guess we need to document all that.

Ah and yes, I need to try meshy asap. It looks cool. :D

akien-mga:
I've been playing around with meshy, the Linux buildsystem is somewhat messy (sic ;)). It's not really meant to be run in source, and the system wide installation isn't pretty either :)

Now I've managed to install it properly I think, but it can't seem to find its own Grid.png texture to display the grid. Same issue with the Kubuntu binary release from the official website.
But at least I can see the models :)

akien-mga:
I've now managed to get Meshy to load our resources.cfg, after I removed some stuff from our materials/RTShaderLib/materials (basically everything apart from RTShaderSystem.material, the rest throw errors when getting loaded in Meshy).

So now Meshy manages to locate my material files if I load a mesh, but it can't load the textures either, so it looks like this texture problem is a real bug.

PR #619, opened 2015-04-05
Stop portal spawning when the dungeon temple is destroyed

hwoarangmy:


akien-mga:
Works fine, thanks to the new map it's fairly easy to get one's dungeon temple destroyed :-P

hwoarangmy:
> What does it fix?

Nothing. I just saw that when using an older version of the keeper hand (while testing older materials) and realized we didn't check the animation exists before launching it.

PR #621, opened 2015-04-05
Add tooltips in research tree + rename Tentacle{,2} + enable WorkshopMachine2

akien-mga:
For the tooltips, I used a "Type: Name" format to make it clearer what is a room, a spell or a trap.

Because of a typo in creatures.cfg, the two tentacles were using the same mesh. This commit fixes it and also renames the tentacles meshes and creature definitions for more clarity. "Tentacle" becomes
"TentacleAlbine" and "Tentacle2" becomes "TentacleGreen". The skeleton used by the two meshes is exactly the same, so the duplicate one is removed and both use Tentacle.skeleton.


Bertram25:
Change looks good. :) I'll test asap.

akien-mga:
I've updated the CREDITS (phew!), and left blank lines for files that we missed previously or that were added recently. Some research will be needed to fill in the blanks.

I've also dropped some more unused textures:
- GoldGroundNormal is not used since it's replaced by DirtNormal
- DungeonTemple_glow, Keeperhand_colD and Keeperhand_spec are not used either, and IINM @Danimal696 does not plan to use them (the current workflow is more basic texture + normal + proper parameters). @Danimal696, if you want to keep those in the repo, I can revert my commit :)

akien-mga:
Small treat for @Danimal696, I added the new WorkshopMachine2 as a random active spot in the workshop, and I found a way to enable its animation :-)

Danimal696:
> I added the new WorkshopMachine2 as a random active spot in the workshop, and I found a way to enable its animation :-)

Nice :) , please erase those unused textures

hwoarangmy:
Works well. Can be merged for me

Bertram25:
Tested as well, looks good. :) Didn't @akien-mga just expanded a functionality? ;)

PR #626, opened 2015-04-08
PNG optimisations with OptiPNG

akien-mga:
Thanks to @oyvindln for the reference, OptiPNG does quite a good job.

I used the 5th optimisation level (out of 7, 7 being the best and slowest), and we gained up to 20% on some pictures (sometimes optipng also removed alpha layers when there were no alpha data, so we go from 4x8 bits/color (RGBA) to 3x8 bits color (RGB)).

I checked the diffs with GitHub, there is no visible alteration.


akien-mga:
With the source repo archived as zip, we go from 83 MB before this PR to 76 MB afterwards. I know disc space is not that much of a concern nowadays, but bandwidth still is for some people, and until we have really badass graphics, I think it's best for us to keep a reasonable size (of course without limiting our creativity, if we include 20 new models with awesome textures for the next release, I don't have anything against a 200 MB repo ;-)).

Bertram25:
This is looking cool. :) I'll test asap. IMHO, time for a lan asap. ;) And then release.

akien-mga:
> IMHO, time for a lan asap. ;) And then release.

Agreed :-) The Windows and Linux binaries from yesterday are the RC IMO, let's try to organise some LAN with the guys from IRC (the community builds up slowly but surely!).

Bertram25:
I should be available tonight for a first quick try if you want. In the meanwhile, I'll test your PR and make tries on the socket problem. But in any case, I don't see any known blockers left. :D

Bertram25:
Is good! Can be merged IMHO.

hwoarangmy:
I could see no visual difference and there are no error or failing assert in the logs. I merge

Issue #624, opened 2015-04-06
Final balancing needs before release?


Bertram25:
@akien-mga @hwoarangmy @Danimal696 @oyvindln 

Do we need some balancing obvious fixes before release and what would they be?

IMHO, the research points need to be 50 on first row, 75 on second, 87 on third, ... as a first start on tweaking.

The waves in Falling Keeper need tweakin even if I've got no idea what. ;)

Apart from that, and the last bugs to fix, we can release, IMHO.

Note:
Transfered all stuff to #418 to make sure most changes are gathered there (so it is easier to make the change log)

Danimal696:
i would say: cheaper wages, less gold stored in each treasure room, reduce all creatures walking speed by a 10%, rise their working times in rooms, set armor increase value to 0.1 for all creatures and raise experience got in battle to be higher than in training room

hwoarangmy:
Increasing research for higher level things makes sense.
I agree wages are too high and treasure rooms can have too much gold (almost no need to build more than 1 treasury).

Concerning waves in falling keeper, it may be too high but I've never really tried the level (unless playing alone to test the portal wave).

IMHO, we should try to set up a lan and try multiplayer ^^

akien-mga:
So to sum up, we should:
- [x] Increase research points needed for higher level researches (50 for first row, 75 for second, 90 for third, 100 for fourth as a first try).
- [x] Decrease wages (say by 30%?)
- [x] Decrease amount of gold stored in each treasury tile (currently 5000, we can make it 2500)
- [x] Reduce creature walking speed by 10%
- [x] Set armor increase/level to 0.1 for all creatures
- [x] Raise gained experience from battle
- [x] Increase mana cost of Call to War
- [ ] Do something for falling keeper waves

About @Danimal696's "rise their working times in rooms", what do you mean exactly? Should we lower creatures' "working efficiency" (if we have such a parameter) or increase the needed research points and crafting points?

Danimal696:
Hi, what i mean is that currently the game plays like its on crack/cocaine, everything is way too fast paced, its good for development but an actual gameplay will suffer from that pace.

There are some points to refine in there:
Walking speed -> too fast in general
Portal spawing time -> you get your dungeon full of critters in no time, increasing the time would be good so we can build some rooms and stop getting flies, also to value a bit more whatever creature you get - mitigated by deciding when you claim portal
Walls -> increase walling time, it should take way longer taking into account its a perfect fortress
[...]

PR #629, opened 2015-04-08
Made sure to disconnect the clients before the server + Parts of #624

### LINKED ISSUE #600: Exiting with window button does not clean up sockets properly
If I exit the game and start it again right away a new game can't be started as the port is in use (On my system (Ubuntu 14.10) at least). I don't know if there is an issue of ports not being closed, or if the OS reserves the port for a while after the application is done. This is frustrating when working with materials or other things where the game has to be restarted to test the effects.

If it is not a bug, a single player game should try a different port if the first one fails when starting.

Bertram25:
I also removed a duplicate call to ODSocketServer::stopServer()
in ODServer::notifyExit()

Fixes #600

Thanks to @hwoarangmy for the hints about this. :)

### DIFF
--- a/config/spells.cfg
+++ b/config/spells.cfg
@@ -14,6 +14,6 @@
 [Spells]
     SummonWorkerNbFree	1
     SummonWorkerBasePrice	1500
-    CallToWarPrice	3000
+    CallToWarPrice	6000
     CallToWarNbTurnsMax	50
 [/Spells]
--- a/levels/skirmish/ScreamInTheDark.level
+++ b/levels/skirmish/ScreamInTheDark.level
@@ -40,7 +40,7 @@ trapBoulder
 [/Seat]
 [Seat]
 seatId	2
-teamId	2
+teamId	1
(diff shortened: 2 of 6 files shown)


akien-mga:
It works fine. There is a small glitch in the logs but it's not related to this PR, it was already there before and it's probably the designed behaviour, but I just want to check:

```
Disconnecting client...
ERROR : Could not receive data from client status=2
Client disconnected state=ready
Stopping server...

Computing turn 16
During this turn there were 0 calls to GameMap::path().miscUpkeepTime=1166
SERVER - Removing Creature BossDragon1
SERVER - Removing Creature BossDragon2
```

Should we do something to prevent this error to happen when the client is disconnected and the server continues polling it for a turn, or is status=2 the signal that tells the server that the client has been disconnected on purpose?

akien-mga:
It's good to merge though, my above comment is more to get some explanations from one of you devs about the error log and if we should do something about it.

PR #631, opened 2015-04-09
Rename music files as TrackName_Author.ogg

akien-mga:
As the TODO says it well, the OpenDungeons Main Theme track really needs to be made configurable, and if we could avoid the need to hardcode it in all menu modes, that would be nice. I'll open a ticket for 0.6.0 so that we keep it in mind (#632).


Bertram25:
Looks quite straight-forward. Feel free to merge if it's working for you. :)

akien-mga:
@Bertram25 It doesn't have much to do with this PR, but could you change the URL in the github repo description to point to opendungeons.github.io?

Bertram25:
Sure. :)

PR #644, opened 2015-04-09
Add multiplayer version of Scream in the Dark

akien-mga:
For our next multiplayer game :-)

I've edited the north part to make it more similar to the south part (also with a river joining the two allied dungeons and a bit more gold from scratch), and removed all custom buildings and claimed tiles from seats 2, 3 and 4. Seat 5 (the middle seat) keeps it's base since it's meant to play alone (most likely AI as a rogue keeper, but some adventurous users might want to play it too).

I've set default teams (1 in the south, 2 in the north and 3 in the middle), but they can be modified at will by the players (up to five teams - free for all).


Bertram25:
Looks good so far. :)

I merge, we can always improve later.

PR #653, opened 2015-04-10
Small fixes

hwoarangmy:


Bertram25:
Tested. I could load games fine, even if I don't know how to precisely fix it.
The mood modifier is working as I got a few PitDemon quickly becoming angry when not fighting.
The game is becoming interesting. :>

Bertram25:
Ah, good to merge, IMHO.

hwoarangmy:
Note that I've raised base mood points. Now, creatures that come to your dungeon are happy. With the default modifiers, if not fighting, the creatures level 2 will be slightly happier than before, level 3 will be less happy and level 4 will be more angry.

Bertram25:
Yeah, never let Level 4 rule... ;)

akien-mga:
Testing now

hwoarangmy:
> How does one select the player able to do the config? Is this only the backend, and the frontend has yet to be implemented?

That's for later. We will see if we want to allow to change.

PR #656, opened 2015-04-10
Walls models for training hall

Danimal696:
these will replace the punching dummy, that should be converted to central
spots along the others.
Also its a sneak-peek at the weapons we will get when forge is coded.


akien-mga:
It works great, I merge and I'll update the mesh format.

Danimal696:
im using scite, i suppose i havent cheked all options needed

akien-mga:
IIRC my Windows days 10+ years, Scite was quite good to edit code files, so it should have an option to use a monospace font indeed.

Bertram25:
@Danimal696 Yes, there is an option to use a monospaced font, the option to use spaces instead of tabs is next to it, btw:
scite
Scite does exist for linux, while we're speaking about it.

PR #667, opened 2015-04-12
Fixed segfault when client disconnects + nickname non ascii chars counting for 2 letters

hwoarangmy:


akien-mga:
Works fine and the changes are quite simple, I merge.

PR #680, opened 2015-04-15
Display backtrace when a segfault happens with MinGW

hwoarangmy:


akien-mga:
I guess I'll have to sit this one out and let @Bertram25 do the testing ;)

Bertram25:
Sure. :)

hwoarangmy:
> I guess I'll have to sit this one out and let @Bertram25 do the testing ;)

I guess I am the only one to use MinGW so it won't be easy to test ^^

@Bertram25 If you want, I can provide an exe to test so you don't have to install everything

akien-mga:
>  If you want, I can provide an exe to test so you don't have to install everything

But then you should also add a bug to trigger a segfault :-p

Bertram25:
It's more than time that I setup a MingW build environment, so I'll start that tonight and test.
In the meanwhile, compiling on Linux to check it's still alright should be ok.
As for creating a segfault, nothing easier. ;)

hwoarangmy:
If you setup a MingW build environment, note that you will need for this PR to include "libz" (it's zlib)  if you use the standard MinGW installer (which is not installed by default).
But I agree that's even better ^^

oyvindln:
If this requires non-standard libraries, shouldn't this be off by default?

hwoarangmy:
Well, it's on only for win32 mingw. But it's such a usefull feature that I'm not sure we should disable it...

ATM, the problem is that the exe is 45 Mb when I compile it in release with debug info (versus 10 Mb in release without symbols). I'm currently having a look if it can be reduced. If any of you know how to do that, I'm interested.

Bertram25:
Apart from using shared libs + optimization for size + strip in the compiling flags, I don't know more.

oyvindln:
> Well, it's on only for win32 mingw. But it's such a usefull feature that I'm not sure we should disable it...

Well it means that anyone that wants to compile it on windows needs to install additional libraries. If we want to always enable it we need to add it to the build tutorial. I think we should have an option to disable it on linux as well though, I would like to turn it off when developing so I can get clean backtraces from the debugger. Anyway, that can be sorted later, doesn't have to be dealt with in this PR.

hwoarangmy:
> Well it means that anyone that wants to compile it on windows needs to install additional libraries

Note that this is true only for people using MinGW (all the code and added libraries are under flag). Moreover, the concerned additional libraries can be easily installed with MinGW installer. I would agree if it would require download and compile a library like SFML or Ogre. But in this case, it doesn't sound like a big issue to me.
Is it such a problem for you ?

oyvindln:
I thought it meant downloading something extra. If it comes with MinGW, it's probably okay.

Bertram25:
libz is one of the most common lib out there. IMHO, it's ok and yes, it can be obtained with mingw installer. Maybe this should be added in the compile how-to?

Bertram25:
Btw, libz is already necessary to build boost for Mingw, so nothing new, btw.

hwoarangmy:
> Btw, libz is already necessary to build boost for Mingw, so nothing new, btw.

Seems not as I didn't had it (I installed it for this PR) and I was able to compile boost. But it is also in Ogre dependencies so, no, that's nothing new. The biggest issue so far IMHO is the exe size. With the actual release executable, you only get the cpp file where the crash occured (not the function nor the line) and the executable size comes from 4Mb to 10 Mb. I could only get a real callstack by compiling in RelWithDebInfo but I get a 45 Mb executable

Bertram25:
> Seems not as I didn't had it (I installed it for this PR) and I was able to compile boost. But it is also in Ogre dependencies so, no, that's nothing new.

You are right after all. Boost complained about zlib not being there, but it actually failed because of the gcc version that I had to update. (from 4.7 to 4.8.1)

Bertram25:
Got a nasty one: If CMAKE set DIRECTX_INCLUDE_DIR to Blabla/Windows Kits/8.1/Include/shared
you have to make it point back to Microsoft DirectX/Include or it won't compile with rather strange errors.

Bertram25:
Ah and if it fails telling it doesn't find dinput.h, you can also change the DirectX_D3D11_INCLUDE_DIR from the Windows Kits to the DirectX folder as well.

hwoarangmy:
I've not had this probably because I didn't compile all the directx renderers

Bertram25:
Note to self:
Fix a few newlines in the How-to.
Indicate the makefile type is MinGW-makefiles in ogredeps
Indicate to set the freetype and OIS include and libs when configuring Ogre builds.

Bertram25:
Allright, boost, ogredeps + ogre compiled, I'll do the rest tomorrow evening. Good night. :night_with_stars:

Bertram25:
It is working fine as far as I tested. It's not often you delete a valid pointer before trying to access it on purpose. ;)
I've also compiled on linux to make sure it doesn't get in the way somehow, even if I had no actual fear from looking at the code. I'll update the wiki how to compile section afterwards with the few bits found along the way.

Well done on that one, it will be really useful for feed-back! :D

Bertram25:
@hwoarangmy Maybe we'll have to bundle zlib.dll or something for the next binary?

hwoarangmy:
Yep. That will be worth checking

Issue #666, opened 2015-04-12
Spectator mode for multiplayer games (or skirmish with only AI players)


akien-mga:
That would be a nice addition for a later milestone, especially for our shy players who don't want to compete against us because we know the core elements on the gameplay and how to exploit balancing issues :-p

Setting to "Later" for now, though I wouldn't be surprised to see @hwoarangmy grabbing this one for 0.6.0 together with the dedicated server stuff ;)

hwoarangmy:
> Setting to "Later" for now, though I wouldn't be surprised to see @hwoarangmy grabbing this one for 0.6.0 together with the dedicated server stuff ;)

Depends. That would need more specifications. First, do we want to use specific seats (ie need to define observers in the level) ?
Second, what do an observer sees (everything? Only what one of the player sees) ?

akien-mga:
I would make observers all-seeing, and not using seats defined in the level. IMO they should not be able to participate in the player chat (if they are all-seeing they could spoil the game), but maybe they could have a separate spectator chat to discuss between observers.

Bertram25:
> I would make observers all-seeing, and not using seats defined in the level. IMO they should not be able to participate in the player chat (if they are all-seeing they could spoil the game), but maybe they could have a separate spectator chat to discuss between observers.

That point of view is relevant in the player's point but less technically, Eh eh.
This means the server will have to send every update data to every spectator and that may (or may not) kill performance. This needs investigating.

I'm still convinced it is something we should add, but maybe not until we have an option to disable them when hosting.
On the client side, the client would simply tell the server it connects as a specator and thus has no interaction whatsoever in game, except for chat. The game mode would also have to handle it by hiding the relevant widgets.

Well, definitely something for later, IMHO.

hwoarangmy:
I agree it is for later. Moreover, that may also imply to change the server logic. ATM, events are only sent to relevant players. If observers can see everything, they should be able to see traps and everything. Well, might not be too difficult but as @Bertram25 pointed, it may hurt performances pretty much since observers would have vision on the whole map. It will be as long as opening the map in the editor...

PR #681, opened 2015-04-16
Drop FAQ, multiplayer The Bridge, improve seat colour

akien-mga:


Bertram25:
I'll try asap.

akien-mga:
To "test" the seat colours in game, I modified Scream in the Dark to use the colorIDs 1, 5, 6, 7 and 8, and then used togglefow to see everything.

BTW I've noticed that togglefow only takes like 1 s to load the entire map on my i5 CPU with Intel HD 4000 GPU, so the performance concerns that we discussed regarding #666 are maybe not that bad.

akien-mga:
Feel free to tell me if there is stuff from FAQ.txt that I haven't ported to the wiki and that we should keep. I left on purpose the "Help!" stuff as I think it's mostly irrelevant now, CMake should make sure people don't build against the wrong Ogre or CEGUI version or stuff like that.
As for the gameplay description, there is already the in-game help and the gameplay wiki page, I did not check whether the FAQ.txt had info that those might be lacking though.

Bertram25:
About the FAQ, I think you drop the obsolete chunks so I'm fine with what was added in https://github.com/OpenDungeons/OpenDungeons/wiki/FAQ

The rest was already part of the wiki anyway.

akien-mga:
I've reworked the README a bit too.

Bertram25:
Is fine. Can I merge?

akien-mga:
Maybe we should add this to the coding guidelines?
http://xkcd.com/1513/
:-D

Bertram25:
Eh eh. @hwoarangmy . Feel free to merge when you're ok with it. :)

akien-mga:
I'll merge myself as I think it's not too disruptive. @Bertram25 Can you merge the current development branch in 0.5? Since neither @hwoarangmy nor @oyvindln's branches have been merged, there is no need to cherry-pick yet.

PR #685, opened 2015-04-19
Fix misspellings using codespell

akien-mga:
Done with the following instructions: codespell --quiet-level=3 --skip=*.cg --write
And then reverting the changes to the files we do not want to touch (AngelScript source, licenses).

Cf. https://github.com/lucasdemarchi/codespell


Bertram25:
I'll move all that to 0.5 as well.

hwoarangmy:
Nice tool, indeed ^^

PR #687, opened 2015-04-20
Fix doc installation

akien-mga:


akien-mga:
I've tested the change and it works fine, and since it's a trivial change I merge it myself :)

PR #693, opened 2015-04-24
Tileset fix

Danimal696:
Fixed a few tiles but mostly someone used wrong tiles, he knows who he is, he
better guards his back from now on...


akien-mga:
Nice! That's a great improvement already!
Feel free to play with config/tilesets.cfg if you need to change what mesh is used in what configuration.

akien-mga:
As usual, the new meshes produce a warning saying that they are "in an older format [MeshSerializer_v1.41]". But IIRC it's not something that you can fix yourself, so I'll handle updating the meshes format once the PR is merged.

Bertram25:
Will test in a few sec. :)

Bertram25:
Excellent fix :D :+1: 

The tile corners are fixed. Only the lava and water animation is yet to be fixed now. I'll merge now and I'd like to move that to the stable and 0.5 branch once the mesh models version is updated. :)

WDYT?

Bertram25:
> he knows who he is, he better guards his back from now on...

Making threats to yourself @Danimal696 ? ;P

akien-mga:
> I'll merge now and I'd like to move that to the stable and 0.5 branch once the mesh models version is updated.

That's a good idea IMO. If we are careful to backport such interesting fixes to the 0.5 branch while we develop 0.6.0, we might end up having a 0.5.1 to release without having to really "work" on it :-)
I do wonder if we're ever going to make use of the stable branch itself though ;-)

Bertram25:
> That's a good idea IMO. If we are careful to backport such interesting fixes to the 0.5 branch while we develop 0.6.0, we might end up having a 0.5.1 to release without having to really "work" on it :-)

Yeah, exactly. :)

> I do wonder if we're ever going to make use of the stable branch itself though ;-)

In the pure theory, stable is representing the latest stable version possible, but as we have 0.x branches that have the same role, it may indeed be redundant. The only use I can see to it is that it will help a newcomer when he is searching for the latest stable version and doesn't know/care about 0.x branches, but again, it's not a strong point.
I'll open an issue about whether we should remove it.

Danimal696:
im a bit concerned about any possible performance hit, since some of the tiles have parts that get hidden by other tiles around when not exposed, if the culling is working well there shouldnt be a problem, can someone try?

Bertram25:
> im a bit concerned about any possible performance hit, since some of the tiles have parts that get hidden when by other tiles around when not exposed, if the culling is working well there shouldnt be a problem, can someone try?

I'm not sure to get what you mean. Are you still speaking about corners or about other changes?
I didn't notice any FPS drop while playing with it, btw.

Danimal696:
yes, for example the full dirt tile was before just a plane, now it has four small "legs", they cover the empty corners when exposed, but they are also present on each full dirt tile on map

untitled

Was Akien or Hmow the one who simplified the set? i cant barely understand how the config file works...

Bertram25:
> yes, for example the full dirt tile was before just a plane, now it has four small "legs", they cover the empty corners when exposed, but they are also present on each full dirt tile on map

Ah, I see. I'm sure we can use the plane for the case where every corner is filled, and the other one for all the other cases.

> Was Akien or Hmow the one who simplified the set? i cant barely understand how the config file works...

It is @hwoarangmy 's work.

The config should be simple though. here is an example:

0000    Dirt_0.mesh GoldGround  0.0 0.0 0.0
neighbors, model, material or empty for default one, X, Y, Z rotation in degrees.
(We only use Z rotation for tiles for now.)

As for the neighbors, a value 0 means ground, 1 means same tile type wall. See those 4 figures as a value for each neighbor clockwise starting on the left neighbor.

I hope it will help you. In your case, for instance, the tile to replace by a plane is a dirt tile with all neighbor being at 1

Danimal696:
Edit: i now fully understand it, since we use 4 comparations instead of the previous 8 there is no other way to do it different to my own. If there is no perfomance impact, there is no reason to worry about this imo

Bertram25:
@Danimal696 

> Edit: i now fully understand it, since we use 4 comparations instead of the previous 8 there is no other way to do it different to my own.

IMHO, it's not completely true. While you can't handle all the case, you can change the most needed.

For instance, in the dirtFull tags, the model value of this line could be changed to use the plane (and maybe remove the rotation):
1111    Dirt_11111111.mesh  Dirt    0.0 0.0 90.0

akien-mga:
> For instance, in the dirtFull tags, the model value of this line could be changed to use the plane (and maybe remove the rotation):
> 1111 Dirt_11111111.mesh Dirt 0.0 0.0 90.0

I'm not sure because that's precisely what caused the corner issue.
corner
This tile has its four neighbours who are wall types, so it uses the plane mesh, and therefore it's lacking the "leg" that should cover the missing part.

akien-mga:
I wonder if we could not put the "leg" on the ground dirt tile actually.

The ground tile next to the corner would have those neighbours: 0011 (different fullness on the left and the top, same fullness on the right and the bottom). So it could be configured to point to a mesh with a "leg" in the top-left corner.

There would be a new problem though, as it wouldn't work if the neighbouring wall tiles were rock or gold for example...

Danimal696:
that was the only solution i could think while using so few tiles that would cover all cases

akien-mga:
> that was the only solution i could think while using so few tiles that would cover all cases

Yes your solution is quite good IMO. It was a valid point to raise the question of performance though, let's see how it behaves on the long run :)

hwoarangmy:
If there is no performance/graphical glitch, that's a good workaround. However, the best solution would be to re-work the tileset to make sure the used tile types finishes at coordinates 0:0.
This way, full tiles would only be a square (as it used to be).

> As for the neighbors, a value 0 means ground, 1 means same tile type wall.

Actually, that's not exactly the same wall type. That means a tile that should be linked to it. Tile links are defined within the "TileLink" tag in the tileset config file. The difference is that is allows to link dirt ground, water or watever as we want. For example, we could have a waterfall when a full dirt tile is next to a water tile ^^

Bertram25:
> Actually, that's not exactly the same wall type. That means a tile that should be linked to it. Tile links are defined within the "TileLink" tag in the tileset config file. The difference is that is allows to link dirt ground, water or watever as we want. For example, we could have a waterfall when a full dirt tile is next to a water tile ^^

Ah, my bad. Guess it's my turn to get whipped for bad doc. XD

> If there is no performance/graphical glitch, that's a good workaround. However, the best solution would be to re-work the tileset to make sure the used tile types finishes at coordinates 0:0.
> This way, full tiles would only be a square (as it used to be).

@Danimal696 What's your point of view about this? Are you ok redoing the tileset, even in little by little? (We support different tilesets so adding a new one would be easy.) :)

Danimal696:
Ok, once i bite the creation of a new tileset fixing this will go in

Bertram25:
Ok, in the meanwhile, the current tileset fix is great. Once @akien-mga has upgraded the mesh versions, I'll move those to 0.5 as discussed.

PR #706, opened 2015-04-29
Implemented spell logic + heal spell + explode spell

hwoarangmy:
To test, the skirmish test map 2v2 have the 2 spells available from the start (as well as test very small)


hwoarangmy:
Rebased

gui/OpenDungeonsIcons.imageset
```diff
@@ -23,6 +23,8 @@
     <Image height="60" name="PortalButton" width="60" xPos="386" yPos="194" />
     <Image height="60" name="SummonWorkerButton" width="60" xPos="2" yPos="322" />
     <Image height="60" name="CallToWarButton" width="60" xPos="66" yPos="322" />
+    <Image height="60" name="CreatureHealButton" width="60" xPos="192" yPos="194" />
```
Bertram25, line 26: I seems you didn't add a shadow to spell icons. I can handle that if you want. Shouldn't prevent a merge IMHO.

levels/multiplayer/GreedOrMight.level
```diff
@@ -36,6 +36,8 @@ trapCannon
 trapSpike
 roomCrypt
 trapBoulder
+spellCreatureHeal
+spellCreatureExplode
```
Bertram25, line 40: New techs :D

source/ai/KeeperAI.cpp
```diff
@@ -757,10 +757,12 @@ bool KeeperAI::handleWorkers()
 
     mCooldownWorkers = Random::Int(3,10);
 
-    Tile* central = getDungeonTemple()->getCentralTile();
-    std::vector<Tile*> tiles;
-    tiles.push_back(central);
-    int summonCost = Spell::getSpellCost(&mGameMap, SpellType::summonWorker, tiles, &mPlayer);
+    // We want to use the first covered tile because the central might be destroyed and enemy claimed
```
Bertram25, line 760: Nice catch.

Bertram25: Could you use the first tile as a fallback? So we keep the central position whenever possible? Looks better IIRC.

source/creatureeffect/CreatureEffect.h
```diff
@@ -0,0 +1,89 @@
...
+std::istream& operator>>(std::istream& is, CreatureEffectType& type);
+
+class CreatureEffect
+{
+public:
+    // If particleEffectName is not empty, the corresponding particle effect will be
+    // sent when the effect is added to the creature for nbTurnsEffect turns.
+    CreatureEffect(uint32_t nbTurnsEffect, const std::string& particleEffectScript) :
+        mNbTurnsEffect(nbTurnsEffect),
+        mParticleEffectScript(particleEffectScript)
```
Bertram25, line 42: Am I dreaming??? :D

source/entities/Creature.cpp
```diff
@@ -296,10 +310,6 @@ void Creature::removeFromGameMap()
         home->releaseTileForSleeping(getHomeTile(), this);
     }
 
-    Tile* posTile = getPositionTile();
```
Bertram25, line 299: Is this a potential crash fix?

source/entities/Creature.cpp
```diff
@@ -2967,6 +3040,38 @@ void Creature::refreshCreature(ODPacket& packet)
     OD_ASSERT_TRUE(packet >> mLevel);
     OD_ASSERT_TRUE(packet >> mOverlayHealthValue);
     RenderManager::getSingleton().rrScaleEntity(this);
+
+    uint32_t nbEffects;
+    OD_ASSERT_TRUE(packet >> nbEffects);
+
+    // We copy the list of effects currently on this creature. That will allow to
+    // apply only effects we were not allowed of
```
Bertram25: I guess you meant "we are allowed to"?

source/game/Player.h
```diff
@@ -185,6 +183,16 @@ class Player
     //! Marks the tiles for digging and send the refresh event to concerned player if human
     void markTilesForDigging(bool marked, const std::vector<Tile*>& tiles, bool asyncMsg);
 
+    uint32_t getSpellCooldownTurns(SpellType spellType);
+
+    void setSpellCooldownTurns(SpellType spellType, uint32_t cooldown);
+
+    //! Called each turn, it should handle Seat upkeep
```
Bertram25: I guess either the doc or the function name is wrong.

source/gamemap/GameMap.cpp
```diff
@@ -306,6 +303,65 @@ void GameMap::clearAll()
     resetUniqueNumbers();
     mIsFOWActivated = true;
     mTimePayDay = 0;
+
+    processActiveObjectsChanges();
+    processDeletionQueues();
+
+    // We check if the different vectors are empty
+    if(!mActiveObjects.empty())
```
Bertram25, line 311: Nice! :D

Bertram25:
This sounds awesome, I can't wait to test. :D 
Ah, while I'm thinking about this, could you update the CREDITS file with the new textures?

Bertram25:
Small compile error:

```
opendungeons/source/gamemap/GameMap.cpp:313:108: error: call of overloaded ‘toString(std::vector<GameEntity*>::size_type)’ is ambiguous
         OD_ASSERT_TRUE_MSG(false, "mActiveObjects not empty size=" + Helper::toString(mActiveObjects.size()));
```

To avoid such things, do you think we could templatize toString()?
nevermind: the Ogre function parameters make it useless.
To me, `size_t` is `long unsigned int`. Adding a new toString() overloaded function fixes the compilation.

Bertram25:
OD doesn't run with this change because it seems you didn't uncomment the particleFX plugin in the plugins.cfg(.in) file. Same for plugin_d.cfg.in for windows.

Bertram25:
An assert failure when trying Explode for the second time on one of the big dragons in the 2vs2 map:

```
Computing turn 118, timeSinceLastTurn=0.687013
> 20 loops in Creature::doUpkeep name:TentacleAlbine1 seat id: 3. Breaking out..
> 20 loops in Creature::doUpkeep name:TentacleAlbine2 seat id: 3. Breaking out..
> 20 loops in Creature::doUpkeep name:TentacleAlbine3 seat id: 3. Breaking out..
> 20 loops in Creature::doUpkeep name:TentacleAlbine4 seat id: 3. Breaking out..
> 20 loops in Creature::doUpkeep name:TentacleAlbine5 seat id: 3. Breaking out..
> 20 loops in Creature::doUpkeep name:BigDragon17 seat id: 0. Breaking out..
During this turn there were 98 calls to GameMap::path().miscUpkeepTime=13814
ERROR: Assert failed file opendungeons/source/game/PlayerFunctions.cpp line 101 info : playerNick=Keeper AI 2, index=0
ERROR: Assert failed file opendungeons/source/ai/KeeperAI.cpp line 742
opendungeons: /build/libsfml-9p5R33/libsfml-2.1+dfsg2/src/SFML/System/Unix/ThreadImpl.cpp:54: void sf::priv::ThreadImpl::wait(): Assertion `pthread_equal(pthread_self(), m_thread) == 0' failed.
Abandon
```

The explode particle effect looks nice. :) But it seems to never end (loop) on the creature once activated.
Or maybe the assert was not related?

Bertram25:
The Heal spell doesn't seem to work for me. It doesn't appear, and once it has been triggered there are Z-buffer problems when the target creature is south of the camera position (The gui flickers):
odscreenshot_2015-04-30_020354_0
No errors in the logs. I'm using mesa 10.4.2 with an intel chipset and the following GLSL version:

```
02:03:04: Added resource location './materials/RTShaderLib/GLSL' of type 'FileSystem' to resource group 'Graphics'
02:03:04: Added resource location './materials/RTShaderLib/GLSL/130' of type 'FileSystem' to resource group 'Graphics'
```

gui/OpenDungeonsIcons.imageset
hwoarangmy, line 26: no I didn't. Is it in your wiki ? I missed it.

source/ai/KeeperAI.cpp
hwoarangmy, line 760: yes. Even if usually, when your dungeon temple central tile is enemy claimed, it is not a good sign ;-)

source/entities/Creature.cpp
hwoarangmy, line 299: No. That's because entities didn't use to be added on tiles on client side. Now, they are because we have to know which creature is on which tile to compute spell cost on client side.
Actually, that's the potential game breaker in this PR if I forgot to release something. But that's the good timing to do such things since we will have time to test and check for side effects

source/entities/Creature.cpp
hwoarangmy: I meant "we are not aware of" (ie since all effects are sent when refreshing, we only create the ones that are not already created on client side - to avoid having to know for each client if the effect is known or not)

source/game/Player.h
hwoarangmy: Yep, it's the player upkeed for the player class ^^

hwoarangmy:
> To avoid such things, do you think we could templatize toString()?

I usually cast them to uint32_t since I do not expect more than 2e9 active objects in OD ^^

> OD doesn't run with this change because it seems you didn't uncomment the particleFX plugin in the plugins.cfg(.in) file. Same for plugin_d.cfg.in for windows.

I used to change manually the generated config file (plugin.cfg). I had not seen the .in. I will change that.

> An assert failure when trying Explode for the second time on one of the big dragons in the 2vs2 map

The assert log is not related. It sometimes happens when AI tries to pickup and release a creature. But I couldn't find why yet (I didn't search much, either). That will need more logs because I do not reproduce that often (but often enough for debugging if time is taken).
Concerning the assert in SFML, that's very interesting because I had something similar (but without message), only the callstack. Since I was doing changes, I thought it was because I had broken something. I have kept the callstack, I will have a look (I hope it is also related to the graphic drivers card crash).

> The Heal spell doesn't seem to work for me. It doesn't appear

Well, I've tested on 2 different computers (both windows 7 64bits - Intel GC and nvidia). For both, it worked. I fear I won't be of much help on that.
But firstly, the effect comes from Ogre samples. When you open SampleBrowser, choose "Particle effects" sample, and "Nimbus" effect, does that work ? I've used the same effect.
Secondly, could you try to manually add an effect when a creature is created and open "test very small map" ? That would allow to easily test without having to cast spells (waiting to have enough mana and all).

To do that, you can just add in RenderManager::rrCreateCreature

```
Ogre::SceneNode* node = creature->getEntityNode();

ParticleSystem* ps = mSceneManager->createParticleSystem(creature->getName() + "ps", "SpellCreatureHeal");
node->attachObject(ps);
```

Note that if you do that on a map where you can loose vision on creatures, it will crash when a creature is re-created because the particle effect will be recreated. But it is pretty nice to test.
Another way, if you prefer, is to edit a level and replace the 0 at the end of a creature by "1 100 SpellCreatureHeal 0.01" (note that if you copy/paste that, you will have to replace spaces by tabs).

hwoarangmy:
> Ah, while I'm thinking about this, could you update the CREDITS file with the new textures?

I don't really know what to say since the png is from Ogre Samples.

hwoarangmy:
> The explode particle effect looks nice. :) But it seems to never end (loop) on the creature once activated.
> Or maybe the assert was not related?

I've had a look at the code and, indeed, I made a stupid mistake explaining why spells doesn't disappear (I push the particle object in the vector before adding the Ogre::ParticleSystem to the object through the RenderManager. Thus, the particle system is always null when released and it crashes when you try to cast the same spell on the same creature.
Will fix that tonight

source/creatureeffect/CreatureEffect.h
akien-mga, line 42: \o/

akien-mga:
Same compile error as Bertram with gcc (+ a ton of "note" messages, that are meant to explain the error AFAICT).
I'll try the beast once this is fixed, I'm not really sure how to fix it myself :)

hwoarangmy:
if you want to try, the fix will be to replace

```
Helper::toString(vectorXXX.size());
```

per

```
Helper::toString(static_cast<uint32_t>(vectorXXX.size()));
```

Anyway, I will post a fix tonight

hwoarangmy:
Note that the spells are currently pretty powerfull (too much ?). With them, I could manage to win the test 2v2 skirmish map only by healing the BigKnight and let him kill all the waves until I could claim the wave portal (of course, without dropping a worker next to the portal when there was no enemies) ^^

gui/OpenDungeonsIcons.imageset
Bertram25, line 26: Yep, see  "Create the shadow". Feel free to upgrade the page if it feels vague or inadequate to anything.

source/entities/Creature.cpp
Bertram25, line 299: Indeed. Cool.

gui/OpenDungeonsIcons.imageset
hwoarangmy, line 26: ok. I also missed that ^^

Bertram25:
> > Ah, while I'm thinking about this, could you update the CREDITS file with the new textures?
> 
> I don't really know what to say since the png is from Ogre Samples.

I can understand that, Ogre texture licensing is a real mess. Since the flare is quite simple, I'll recreate it myself: https://github.com/OpenDungeons/OpenDungeons/issues/709

> > Helper::toString(static_cast<uint32_t>(vectorXXX.size()));
> 
> Anyway, I will post a fix tonight

N.B.: There might more than one of such thing. When I quick-fixed it, I lazily added an overloaded toString() function to cope with that one.

> [...] To do that, you can just add in RenderManager::rrCreateCreature [...]

Thanks a lot! :D I'll sure test all that. I shows all the preliminary work you had to do before seeing particles in game. :) Note that might well be not your fault at all. Drivers are usually a fragile thing.

Bertram25:
> Note that the spells are currently pretty powerfull (too much ?). With them, I could manage to win the test 2v2 skirmish map only by healing the BigKnight and let him kill all the waves until I could claim the wave portal (of course, without dropping a worker next to the portal when there was no enemies) ^^

Only you can create stuff to make BigKnights even more powerful. ;P
Don't worry, we'll balance all that afterwards.

hwoarangmy:
Actually, that's the same for the spells/icons and nearly everything here. I've just created that to not have an empty shell but if you have a prefered icon or you want to change the duration/value, please feel free (and it's easy since in the configuration files).

Bertram25:
Newest changes look good.
I'll check about the heal particle problem when testing. :)

hwoarangmy:
@Bertram25 @akien-mga I've fixed what you've seen (excepted the icons, I will let you finish the job so the look and feel stays the same).

In any case, I confirm the particles show on my side and I have no issue at all. @akien-mga Could you please confirm if you have the same bug as @Bertram25

hwoarangmy:
> Newest changes look good.

BTW, I've also changed the plugins path on windows so that it works with how we package the release. That's cool, I won't need to copy the plugins.cfg files I had fixed anymore ^^

akien-mga:
Tested the new spells and cooldown:
- Explode works fine, it explodes a first time when I click on the creature with a nice particle effect, then a second time a couple of second later. Is that expected or should the animation play only once?
- Same issue as @Bertram25 with the heal spell on Linux with my Intel chipset. With my Nvidia GPU, the animation plays fine, so that's a driver issue. I have the latest driver for the Intel HD Graphics 4000 chipset though, so it's probably safer to assume that it's not a drivers bug, but we are using a Nvidia feature that is not compatible with Intel.
- Icons are great :)
- IMO misclicking should not trigger the cooldown, at least not with the current state of the game. If there was some visual feedback when you hover a creature (like a halo), why not, but now it's would mostly create frustration for little added value IMO.
- Should there be a cooldown for the Summon worker spell? Basically you can invoke 3 workers at a time if you do a 3x1 tiles invocation, but not by clicking three times. I don't have a fixed opinion, and anyway this kind of things can be fixed later on in the release cycle (same thing for the misclicking triggering the cooldown).

Good to merge already IMO, then we should just open issues for what has been discussed in this PR.

hwoarangmy:
> Is that expected or should the animation play only once

I've only used the particle effects as examples. Actually, it repeats itself endlessly but because it last only 5 turns, it plays twice.

> Same issue as @Bertram25 with the heal spell on Linux with my Intel chipset.

I tried on a Windows 7 with Intel HD 4500 and it works fine. Could you please try Ogre samples? You have to open the SampleBrowser, choose "Particle effects" sample, and "Nimbus" effect. Does that work ? I've used the same effect.

> IMO misclicking should not trigger the cooldown

That was a requirement in the Spell logic issue ;-) But it can be changed, of course.

> Should there be a cooldown for the Summon worker spell?

All the cooldowns are configurable in spells.cfg so I would say yes and we can set it to 0 if we want. Concerning the tile invocation on many tiles vs only 1 tile, I would say that it encourages players to not spend their mana (to allow to cast spells on as many creatures as they can). So, it adds something to the gameplay. However, as everything else, it can be changed later. I will probably give a try in some time to a 1 creature target spell (that doesn't display the tile selector).

Bertram25:
> I've only used the particle effects as examples. Actually, it repeats itself endlessly but because it last only 5 turns, it plays twice.

Indeed, this is IMHO easily fixable, even afterwards.

> Should there be a cooldown for the Summon worker spell?

+1 Let's put it back to 0. I wouldn't add a space between the cooldown value and the "s" as well, but all that is details non blocking a merge.

> I tried on a Windows 7 with Intel HD 4500 and it works fine. Could you please try Ogre samples? You have to open the SampleBrowser, choose "Particle effects" sample, and "Nimbus" effect. Does that work ? I've used the same effect.

@akien-mga Could you also try the most particle effect examples possible and see whether there are other ones failing? This will help us determine which keywords in the particle files and/or the corresponding material files are leading to a driver or a gl version supported failure. :)
In the mean time, we can use another working example btw.
Pinging @oyvindln as he has knowledge in this area.

Bertram25:
> excepted the icons, I will let you finish the job so the look and feel stays the same

I'll sure do. :) Non-blocking for this PR anyway.

Bertram25:
> BTW, I've also changed the plugins path on windows so that it works with how we package the release. That's cool, I won't need to copy the plugins.cfg files I had fixed anymore ^^

Yeah, I've just seen that. One good thing fixed. :)

With latest changes, after casting the heal spell, the game doesn't have any GUI flickering .
Casting a spell several time isn't making the game crash anymore. :)

When playing on the skirmish test 2vs2 players, it seems one can't cast a spell on ally claimed tiles, btw.

I'd thus request that the following gets fixed before merging if possible, or asap by anyone after merging since I know @hwoarangmy has got some other improvements waiting (I have some, too, based on his changes. ;]):
- Make the the spawn worker spell have a 0 cooldown value.
- Make the heal, explode, and spawn worker spells castable on creatures whatever their tile type is.
- Make the spell be casted only if it's relevant. I mean if it has no effect, it shouldn't trigger.

The problem with the particle effect not visible with intel chipset should become an issue and be investigated there. (I'll be on that.)

@hwoarangmy Are you willing to make the changes before merging, or would it be best for your own repository management if we delay those changes to another PR?

@akien-mga Can you confirm the crashes are gone?

Anyway, it's a great step forward. :D Thanks @hwoarangmy

Bertram25:
Ah, and now we have more spells and more rooms, it would be better to make the text overlay display the unit along with the cost.
Example:
Training Hall [150 Gold]
Heal [5000 Mana]

The spells and rooms overlay texts shouldn't show a 0 cost when unspawnable/uncastable, just the text in red.

WDYT? If you're ok with it, I'll open an issue about that.

hwoarangmy:
> Are you willing to make the changes before merging, or would it be best for your own repository management if we delay those changes to another PR?

Yep, that's improvements so I would prefer to open an issue and do it later. What I have currently are only small fixes (some are interesting for the game but code talking, its pretty simple).
So, please merge if it works for you.

I am also currently having a look on a Spell manager (based on the old AI factory) that will allow to register more easily spells (and allow @oyvindln to not have to include all of them if he wanna go for a unit test). So, in my current repository, Spells are a bit changed...

> it would be better to make the text overlay display the unit along with the cost.

Agreed

> The spells and rooms overlay texts shouldn't show a 0 cost when unspawnable/uncastable

ok

akien-mga:
Merging then, I'll open tickets for the things that are left to be done. We're in the development branch after all ;-)

akien-mga:
> Can you confirm the crashes are gone?

No crash for me btw, but I did not try the first version of this PR as it wasn't building.

Bertram25:
> Merging then, I'll open tickets for the things that are left to be done. We're in the development branch after all ;-)

Yeah, so true! :) Thanks @akien-mga

PR #708, opened 2015-04-29
Made the Editor flag icon colored with the currently selected seat.

Bertram25:
And factored some code. :)


akien-mga:
It works great, and that's awesome :-D

Issue #713, opened 2015-05-03
Improvement to the text overlay


Bertram25:
- [x] Now we have more spells and more rooms, it would be better to make the text overlay display the unit along with the cost.
  Example:
  Training Hall [150 Gold]
  Heal [5000 Mana]
- [x] The spells and rooms overlay texts shouldn't show a 0 cost when unspawnable/uncastable, just the text in red.

hwoarangmy:
true

Bertram25:
Ah, btw, while we're speaking about it. 
Both debug and room/spells/trap text overlay should be hidden when leaving the game/editor modes.
Atm, if you leave the game with those shown, they'll stay shown on the main menu.

Bertram25:
Done with commit b96b57e7821c16bced7ed6881987a74c7ac42694

PR #727, opened 2015-05-04
Added a spell manager to handle spells + small fixes

hwoarangmy:
@Bertram25 @oyvindln The spell manager allows to add spell just by adding the corresponding source file (and remove them by not compiling it).
Moreover, it won't compile if the spells do not define the required static functions.

I think it is a pretty elegant way to add spells. If you also like it, I might give a try at doing the same for missiles and, if possible, for the other entities.


Bertram25:
Looks cool. I'll test now. :)

Bertram25:
I like the pattern used to templatize the spells and thus make them easily added/removed.
I'll let @oyvindln tell whether he likes it or see improvement. :)

Bertram25:
IMHO, ok to merge it.

akien-mga:
Merging then, I'll keep #713 open for the last adjustment that was requested.

PR #728, opened 2015-05-04
Misc fixes related to particles and spells. :)

Bertram25:


hwoarangmy:
Looks good to me. I will test that asap.

akien-mga:
Works great on my Intel chipset :)
I agree with @hwoarangmy that 5 turns is maybe a bit low for a spell, but anyway balancing will come later IMO.

Bertram25:
Btw, the license of the heartflare file is including hearts from game-icons.net so it's license should be cc by sa 3.0. I'll fix this in my next PR.

PR #743, opened 2015-05-07
Mystery box

Danimal696:
Wtih animation called"Loop"


hwoarangmy:
Works well, no error in the logs (only the usual "old format"). Merging

PR #746, opened 2015-05-08
Misc improvements...

Bertram25:


hwoarangmy:
testing asap

hwoarangmy:
Works well. Good to merge IMO.
The spell icon do not seem as well as expected. It looks more like a palm tree. But icons are better than text IMO.

Bertram25:
> The spell icon do not seem as well as expected. It looks more like a palm tree. But icons are better than text IMO.

Let's fix this then. What about those icons?  (Pinging @Danimal696 @oyvindln @eugeneloza @akien-mga )
http://game-icons.net/lorc/originals/star-swirl.html
http://game-icons.net/lorc/originals/fairy-wand.html
http://game-icons.net/lorc/originals/magic-swirl.html
http://game-icons.net/lorc/originals/book-storm.html (flipped horizontally)
http://game-icons.net/lorc/originals/lightning-trio.html
http://game-icons.net/lorc/originals/burning-book.html
(Note that you can dynamically resize the icon in the given links.)

hwoarangmy:
I like the magic-swirl ^^

eugeneloza:
I like the icons.... actually exactly what I was looking for my project :D :D :D Thanks, Bertram!!!
Wow! they have all the 'skill icons' I might ever need!

Danimal696:
i like the book storm and the fairy wand

Bertram25:
> I like the icons.... actually exactly what I was looking for my project :D :D :D Thanks, Bertram!!!
> Wow! they have all the 'skill icons' I might ever need!

Cool :)

> i like the book storm and the fairy wand

I need a consensus. lol

hwoarangmy:
I don't like the book storm (when in 32x32, it looks like a mouth or a vegetable). But the fairy wand is ok too (even if I think the magic swirl makes me think to spells) ^^
But @Bertram25 you also have your word to say ;-)

Bertram25:
The fairy wand has a not-evil touch, I must say. I'll go for the magic swirl. Sorry @Danimal696 ;)

akien-mga:
+1 for magic swirl :)
The book storm has an evocative outline if I may say, we would get another trolling three on Debian MLs [1] if we used it :p

[1] https://lists.debian.org/debian-devel-games/2015/04/msg00029.html

akien-mga:
Looks very good to me, a nice set of improvements as always :)

I'll let @hwoarangmy and @oyvindln review 78c4e1e before merging though, as it's a big change (at least in terms of diff length :-P).

I think I spotted one CEGUI issue, but I haven't found how to reproduce it yet. I had the game playing in background while reviewing the commits here, and when I came back to the game there was a very small circle displayed on the screen (like 4px of diameter or something), and it would move a bit when I approached it with the cursor. My guess would be that it was an empty tooltip (with just the four corners of the tooltip box, making it look like a small circle). I haven't found yet how to reproduce it though, and it shouldn't prevent merging.

hwoarangmy:
I've tried and seen a bug in the sleep animation. It is looped. In the 2v2 skirmish tets map, build a dormitory a drop the BigKnight, you will see him loop ^^
But I've looked at the code and I think the bug was already there (I don't know why we didn't see it until now).

hwoarangmy:
> I think I spotted one CEGUI issue, but I haven't found how to reproduce it yet. I had the game playing in background while reviewing the commits here, and when I came back to the game there was a very small circle displayed on the screen

I confirm the bug but it is not related to this PR. I've seen it since a while but I've never understood why it happened (I've seen in something like 3 or 4 times)

hwoarangmy:
testing

hwoarangmy:
The problem with sleep looping is still present because the sleep animation is sent everyturn. However, there are plenty of changes and many improvements here so I think it is better to merge and fix that.

Bertram25:
About the sleep problem : let's open an issue about it. :)

PR #759, opened 2015-05-17
Implemented doors + fixes 2 crashes

hwoarangmy:
To use doors, you can slap to open/close (animation will change)


Bertram25:
The code looks great from what I could see. I think you fixed a crash I encountered earlier. :)

Just one question: Has the door a limited amount of "activations", closed/open times before being disabled? If so, I do think it shouldn't be limited.

I'll test now. :)

akien-mga:
Oh, shiny... :-D
I'll try that tomorrow.

Bertram25:
Looks fine in game. The mesh to be changed later of course. :)
I can finally lock-up my minions while I'm preparing an attack!
I caught a crash but I don't think it's related to the PR. (I'll make an issue.)

Bertram25:
IMHO, good to merge. :)
@Danimal696 Only the door mesh is missing. :)

hwoarangmy:
> Has the door a limited amount of "activations", closed/open times before being disabled?

No, you can open and close it as much as you want ^^

> I caught a crash but I don't think it's related to the PR

I've seen the issue and I confirm I have already had the same kind of crash without this PR

akien-mga:
It works great, and I love the door model :-P
I did not spot any regression so far, so let's merge and test it more along the road :)

Danimal696:
thats an interesting door... how are it animations called?

Nevermind, got them. i would be scared to approach such a funky door

akien-mga:
@Danimal696 The animations and mesh are hardcoded here:
https://github.com/hwoarangmy/OpenDungeons/blob/development/source/traps/TrapDoor.cpp#L33-L35

But feel free to use different names for the actual door animations, those ones are just for fun as I understand it ;)

hwoarangmy:
@Danimal696 Yep, don't bother about that. Just tell us what you made and we will adapt. That being said, for testing purpose, the current animations are "Squeeze" and "Dance"

PR #768, opened 2015-05-20
Some fixes + used trap mesh + kobolds AI improvements

hwoarangmy:
@oyvindln Will be happy about GameEntities using a boolean for knowing server/client gamemap reducing gamemap dependency.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -432,7 +432,7 @@ endif()
 #This has to cover the versions not already known by CMake
 set(Boost_ADDITIONAL_VERSIONS 1.47 1.47.0 1.47.1 1.55.0)
 
-set(OD_BOOST_COMPONENTS system filesystem locale)
+set(OD_BOOST_COMPONENTS system filesystem locale program_options)
 
 if(WIN32 AND ${OGRE_FOUND})
     # On windows we can try to use boost from the Ogre SDK instead
--- a/source/ODApplication.cpp
+++ b/source/ODApplication.cpp
@@ -23,7 +23,6 @@
 
 #include "render/ODFrameListener.h"
 #include "render/TextRenderer.h"
-#include "render/RenderManager.h"
 #include "sound/MusicPlayer.h"
 #include "sound/SoundEffectsManager.h"
 #include "render/Gui.h"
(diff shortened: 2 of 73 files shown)


Bertram25:
I just tested it live and overall no big behavioral changes are noticeable, which is a good thing when refactoring. :)

Ah btw, it seems the door isn't translucent enough when placed in game but tht's completely separate.
I'll come back about this once the model is completely functional.

hwoarangmy:
> Ah btw, it seems the door isn't translucent enough when placed in game but tht's completely separate.

Yes, I agree. But to be honest, I've not even tried to complete the door (only to place it unactivated). In any case, that's something that may be changed when @Danimal696 will have uploaded the new mesh

hwoarangmy:
Btw, note that I've tried to let a game play for many time with this version (on 2 different computers) on fallen keeper level.
I turteled a functional dungeon and closed the portal with a door.
Then, I let the game play, some of my creatures would become rogue and be killed by the others.
From time to time, I went to pickup the creatures kept in the portal (that spawned after the ones becoming rogue) every 15 to 30 minutes.
Like this, having nearly all creature types, fighting from time to time, I could let the game play for more than 3 hours. Actually, I've never seen a crash even after playing a real game.
@Bertram25 If you see a crash, even not explained or without bt, please just tell it so that we know if we have pending crashes. So far, I've not seen the driver bug with this version (and as I said, I've tested it).

Danimal696:
> Ah btw, it seems the door isn't translucent enough when placed in game but tht's completely separate.
> I'll come back about this once the model is completely functional.

the door has normals applied, and i remember normals and transparency dont work at all together, you are free to modify to the material file and erase the normal png for door if that fixes it

akien-mga:

hwoarangmy:
> the door has normals applied, and i remember normals and transparency dont work at all together, you are free to modify to the material file and erase the normal png for door if that fixes it

I confirm that with the normal, setting opacity to 0 do not change anything. We will have to deactivate them for the door.

Bertram25:
> I confirm that with the normal, setting opacity to 0 do not change anything. We will have to deactivate them for the door.

Ok, then that's another issue. We'll have either to make it work with normals or disable them for the trap models, I suppose.

Bertram25:
I'll merge :)

PR #771, opened 2015-05-21
Levels cleanup + put doors in model's fair test level

akien-mga:


Danimal696:
neither i do, its from someone from the original team

Bertram25:
> Let's attribute it to OD Team then. Personally I would be in favour of attributing credits for all our levels to OD Team directly, and maybe just specify a contributor name for external community contributions that we might want to upstream.

Sounds sane. Do you agree everyone?

Danimal696:
"OD Team", yes, best option for any core level we do, i dont think we are that hungry for credit on everything we do

Bertram25:
It's mergeable and working, so I'll merge it too. Let's fix the credits part afterwards.

PR #775, opened 2015-05-22
Allow launching OD in standalone server mode + handle seats on server side

hwoarangmy:


akien-mga:
This is working flawlessly :-D

akien-mga:
The dedicated server handles properly connections and disconnections of players, even if it's seat 1 (the one who set up seats) that disconnects in game.
There isn't much details in the server log when players connect or disconnect, but I guess this will be added later on.

When the last players leaves the game, the server process stops properly (in the future we might want it to get reinitialised and ready for another game maybe, but that should probably be discussed in its own ticket).

The only "issue" I've seen so far is that if seat 1 disconnects from the seat configuration lobby, and then reconnects (or another player connects), nobody has rights to setup and start the game :)

akien-mga:
> There isn't much details in the server log when players connect or disconnect, but I guess this will be added later on.

Actually there are details in the log, but it would be nice to show them in stdout/stderr too so that the server can be monitored from the terminal where it's running.

akien-mga:
BTW I also tested hosting a non-standalone server, and it still works fine as before the PR.

hwoarangmy:
> Actually there are details in the log, but it would be nice to show them in stdout/stderr too so that the server can be monitored from the terminal where it's running.

ok. I will also output logs on std::cout

hwoarangmy:
@akien-mga I've fixed the issues. I've also added logs to std::cout and added a few more logs

akien-mga:
I confirm that it works fine with 9ec35a2, #776 is fixed.

akien-mga:
Good to merge IMO, but I haven't really reviewed the code.

hwoarangmy:
Also note that I've fixed the issues when the "host" player disconnects (in this case, the next player, if any, becomes "host")

akien-mga:
Awesome :-D

Bertram25:
I'll test now. :) Looks good!

Bertram25:
Works like a charm, even with several local clients. I think this is mergeable. Ok for you @akien-mga ? :)

PR #782, opened 2015-06-01
Handle inputs in a generic way in gamemode/editormode + small fix on worker behaviour + used macro for logs

hwoarangmy:
I've splitted the commit between spell/room/trap to make it easier to read. The basic idea is to let each spell/room/trap handle the inputs (selected tiles/text) as it wants. For example, building a door will not allow to choose more than 1 tile and will display an explicit text if the door is not next to 2 full tiles.
Later, it will also allow to change easily icons if we want to (per spell/room/trap).

I've also changed a bit how text is displayed when building rooms/traps/spells
- If the player do not select tiles, the price will be for 1 target and red if not enough mana/gold for 1 target. White otherwise.
- if the player selects a tile (left button), the price will be displayed white and it will be what will be spent if he releases the button (0 if no target or not enough gold/mana or more if enough gold/mana).

Note that ATM, the tile selector is not displayed by default in gamemode (only when tiles are selected for digging). This is more to test this behaviour. If you prefer displaying it, I will make a commit.

### DIFF
--- a/source/rooms/RoomManager.h
+++ b/source/rooms/RoomManager.h
@@ -22,7 +22,11 @@
 #include <istream>
 #include <cstdint>
 
+class ClientNotification;
 class GameMap;
+class InputCommand;
+class InputManager;
+class ODPacket;
 class Player;
 class Room;
 class Seat;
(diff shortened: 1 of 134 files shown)


Bertram25:
I'll be on the review in something like half an hour. :)
It's good to be back at this.

Bertram25:
Hi again,

For once, I had a game before looking at the code. Here are my comments:
- The treasury tile displayed cost is always 0 while tiles seem to be properly paid.
- When you have creatures in your hand, they don't appear next to it anymore.
- I'd request to show the tile selector when you have a tile based option selected:
  - When you're in dig/undig mode. (Per default)
  - When placing a room or trap.

Why? Because sometimes my selection was off without it and it's annoying to destroy back a room or trap tile, or undig the selected tile in that case.

Bertram25:
Off-topic: What do you think about adding a 'Zzz' icon above creature's head when they are sleeping?

source/rooms/RoomManager.h
```diff
@@ -30,53 +34,65 @@ class Tile;
 
 enum class RoomType;
 
-//! Class to gather functions used for rooms
+//! Class to gather functions used for rooms. Each room should define static functions allowing to handle them:
```
Bertram25, line 37: Cool, thanks for the doc! :)

source/spells/SpellCreatureExplosion.cpp
```diff
@@ -18,44 +18,67 @@
...
 static SpellManagerRegister<SpellCreatureExplosion> reg(SpellType::creatureExplosion, "creatureExplosion");
 
-int SpellCreatureExplosion::getSpellCost(std::vector<EntityBase*>& targets, GameMap* gameMap, SpellType type,
-    int tileX1, int tileY1, int tileX2, int tileY2, Player* player)
+void SpellCreatureExplosion::checkSpellCast(GameMap* gameMap, const InputManager& inputManager, InputCommand& inputCommand)
 {
+    Player* player = gameMap->getLocalPlayer();
     int32_t priceTotal = 0;
-    int32_t pricePerTile = ConfigManager::getSingleton().getSpellConfigInt32("CreatureHealPrice");
+    int32_t pricePerTarget = ConfigManager::getSingleton().getSpellConfigInt32("CreatureExplosionPrice");
```
Bertram25, line 41: Lol, nice catch.

source/traps/TrapManager.cpp
```diff
@@ -155,52 +208,235 @@ void TrapManager::registerTrap(TrapType type, const std::string& name,
...
+        inputCommand.selectTiles(tiles);
+        return;
+    }
+
+    std::vector<Tile*> sellTiles;
+    std::vector<Tile*> tiles = gameMap->rectangularRegion(inputManager.mXPos,
+        inputManager.mYPos, inputManager.mLStartDragX, inputManager.mLStartDragY);
     for(Tile* tile : tiles)
     {
+        // FIXME: ATM, building is not refreshed when building a trap on client side. For this reason,
```
Bertram25: Want to open a new issue?

hwoarangmy: Done #783

hwoarangmy:
> The treasury tile displayed cost is always 0 while tiles seem to be properly paid.

Yes. I had forgotten that on client side, there is no room so the computation always show the first tile as free. I will have a look at how we can handle that.

> When you have creatures in your hand, they don't appear next to it anymore

I had not seen that. It might come from removing the default tile indicator. Probably not a big deal, I will fix that.

> When you're in dig/undig mode. (Per default)

For having played, I agree it is better to display it. However, as I said, it was more a demonstration that we don't have to. I will add it again.

> When placing a room or trap.

When placing room/traps, the tile selectors are displayed on the correct tiles. Do you want to always show the tile selector under the keeper hand (even if not buildable) ?

hwoarangmy:
> Off-topic: What do you think about adding a 'Zzz' icon above creature's head when they are sleeping?

Might be a good idea when a creature is sleeping or going to its bed...

Bertram25:
> For having played, I agree it is better to display it. However, as I said, it was more a demonstration that we don't have to. I will add it again.

Np. :)

> When placing room/traps, the tile selectors are displayed on the correct tiles. Do you want to always show the tile selector under the keeper hand (even if not buildable) ?

Now you tell it, it will be even better if it is shown only when buildable. I must have overlooked that part.

Bertram25:
> Might be a good idea when a creature is sleeping or going to its bed...

I was thinking only when they are sleeping as we have another one when sleepy. But we can reuse the same for both. :)

Bertram25:
Speaking about the code overall, this is a huge refactoring + a lot of cleanups. Great work. :)

hwoarangmy:
> I was thinking only when they are sleeping as we have another one when sleepy. But we can reuse the same for both. :)

That's not exactly the same concept. A creature can be sleepy but not going for its bed (for example if it has no bed, is fighting or doing something else). A creature can also be awake and go to bed (if wounded).
I agree the concept is similar but both information seem pertinent to me.

akien-mga:
I can confirm the small issues pointed out by @Bertram25, otherwise I don't see any regression. Nice work :)

hwoarangmy:
I've done requested changes and fixed some related stuff ^^

Bertram25:
I'll test the latest changes and review code now. :)
Congrats for all the work done!

Bertram25:
I tested it and as far as I tested, there is still something strange concerning rooms.
Try to open the model's fair level and try to place several different type or rooms. By playing with that, the rooms should start not being added or added on random tiles after some tries. (I mixed one tile room placement and several tiles room placements).

Good Week-End guys! :)

akien-mga:
> Try to open the model's fair level and try to place several different type or rooms. By playing with that, the rooms should start not being added or added on random tiles after some tries. (I mixed one tile room placement and several tiles room placements).

Actually the issue is not with tile placement, it's that the (currently) empty space in the model's fair is meant to have three rows of traps, each row being to tiles wide. For some reason those traps are not shown anymore, so you still can't place stuff on it, see:
od-mood1
I guess it might be a regression in 919d90611?

> Good Week-End guys! :)

Thanks, you too! :)

hwoarangmy:
@Bertram25 @akien-mga I've fixed the issue. I had tested enemy traps loaded from level and allied built ingame. But not loaded from level ^^
Now, it should be ok.

akien-mga:
Looks good :-) Everything works fine AFAICT so let's merge it, I know you already have stuff boiling locally :p

Bertram25:
Thanks @akien-mga :)

PR #793, opened 2015-06-06
Fixed the console auto-completion when using tab.

### LINKED ISSUE #757: Regression in command completion in the console
When I press tab in the console, the available commands are still displayed but the current command is cleared instead of being completed.

Bertram25:
The console editbox was executing the prompt command
when this event was happening: CEGUI::Editbox::EventTextAccepted
But this CEGUI event is also triggered when tab is pushed, breaking the auto-completion.

This is now fixed.

I also added support to use the keypad enter key as well as the standard enter key
for chat/console.

Fixes #757


akien-mga:
Works great, and it also seems to fix the issue of having the results of the first sent command displayed each time you input a wrong command and sent it with tab :)

PR #796, opened 2015-06-08
First part of #3

Bertram25:
- Created the SettingsWindow class that will take of creating/deleting the corresponding widget in any mother modes.
- Made it so the settings window is somewhat functional by loading/storing Ogre config.
- Removed the Ogre config window at startup and made OD capable to auto-create a viable config from scratch if necessary.

This PR is made to reach a first good step in implementing this feature and deal with a few questions at the same time.

What's left IMHO, is to move the default config making and the different stored values in the config manager and synd the config manager with Ogre.
Also, video settings changes should be applied wihout the need to restart the app.
WDYT?

I you agree, I'll work on that for the next PR.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -323,6 +323,7 @@ set(OD_SOURCEFILES
     ${SRC}/modes/MenuModeReplay.cpp
     ${SRC}/modes/MenuModeSkirmish.cpp
     ${SRC}/modes/ModeManager.cpp
+    ${SRC}/modes/SettingsWindow.cpp
 
     ${SRC}/network/ChatEventMessage.cpp
     ${SRC}/network/ClientNotification.cpp
--- a/gui/MenuMain.layout
+++ b/gui/MenuMain.layout
@@ -57,5 +57,13 @@
             <Property name="Text" value="Quit" />
             <Property name="AlwaysOnTop" value="True" />
         </Window>
+        <Window type="OD/ImageButton" name="SettingsButton" >
+            <Property name="Area" value="{{1,-27},{0,5},{1,-2},{0,30}}" />
+            <Property name="ButtonImage" value="OpenDungeonsIcons/OptionsIcon" />
+            <Property name="AlwaysOnTop" value="True" />
+            <Property name="TooltipText" value="Options" />
+            <Property name="ClippedByParent" value="False" />
+            <Property name="Visible" value="True" />
+        </Window>
     </Window>
 </GUILayout>
--- a/gui/MenuMain.layout
+++ b/gui/MenuMain.layout
@@ -25,37 +25,30 @@
         <Window type="OD/MainMenuButton" name="StartSkirmishButton" >
             <Property name="Area" value="{{0.5,-310},{0.5,-10},{0.5,-170},{0.5,70}}" />
             <Property name="Text" value="Skirmish" />
-            <Property name="AlwaysOnTop" value="True" />
         </Window>
         <Window type="OD/MainMenuButton" name="LoadGameButton" >
             <Property name="Area" value="{{0.5,-310},{0.5,100},{0.5,-170},{0.5,180}}" />
             <Property name="Text" value="Load" />
-            <Property name="AlwaysOnTop" value="True" />
         </Window>
         <Window type="OD/MainMenuButton" name="StartMultiplayerClientButton" >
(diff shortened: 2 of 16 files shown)


hwoarangmy:
Seems good to me apart from my few comments. I will try that tonight

akien-mga:
Yummy :-D I'll be testing that now.

akien-mga:
It works fine, including when starting with an empty ogre.cfg (I should try also to start with a corrupted ogre.cfg to see how it behaves, mwahahaha). On the code side, it looks good to me, I agree with @hwoarangmy's comments about config saving.

> What's left IMHO, is to move the default config making and the different stored values in the config manager and synd the config manager with Ogre.
> Also, video settings changes should be applied wihout the need to restart the app.

Sounds good to me. Having video settings applied without a restart needed would be really cool.

hwoarangmy:
> Also, video settings changes should be applied wihout the need to restart the app.

If it is possible, it would be better. But most games I know recquire to restart when changing video settings. I guess it won't hurt if we do the same if it is not possible/hard.

akien-mga:
> > Also, video settings changes should be applied wihout the need to restart the app.
> 
> If it is possible, it would be better. But most games I know recquire to restart when changing video settings. I guess it won't hurt if we do the same if it is not possible/hard.

Indeed. I guess the key factor will be whether we can recreate the Ogre/CEGUI with different settings without seeing regressions.

hwoarangmy:
I tested and it works. Video changes do not apply when we clic apply but I guess that's the expected behaviour ?

FYI, I've tried to integrate Ogre HLSL to check if win32 releases work with Direct3D9 and they do. Do you think it would be possible to let players choose the renderer in the settings window ?

Bertram25:
Disclaimer: I should have put that maybe a bit more clearly in the description but you may unfortunately know what it is to create PR late at night. ;)
This PR shouldn't forcefully be merged as is. Let's raise questions thanks to it and I'll fix them little by little, so we have an interesting ending change. :)

That being said,

@hwoarangmy 

> Video changes do not apply when we clic apply but I guess that's the expected behaviour ?

Actually, I'd like to make them apply right away if possible. After all, we can resize the render Window on the fly so we should be able to do that, IMHO. I'll work on that asap.

> FYI, I've tried to integrate Ogre HLSL to check if win32 releases work with Direct3D9 and they do. Do you think it would be possible to let players choose the renderer in the settings window ?

Sure! :) I was about to ask about showing the renderer option, ...
In fact, I'd like to add back at least the FSAA and Vsync options as well.

@akien-mga 

> (I should try also to start with a corrupted ogre.cfg to see how it behaves, mwahahaha)

Eh eh. Good idea. :)

Thanks for your code review, I'll fix what was discussed and try to improve all that tonight.

Bertram25:
@hwoarangmy @akien-mga Now, I should be done for the current PR. :)
(I can even hide back the renderer option for now if you want.)

The next step is to add user config support (I guess into the config manager) and save load settings only using that mean.

akien-mga:
The video settings can be applied fine in game, nice job.
I can't test changing the renderer as I only have OpenGL installed, but maybe @hwoarangmy will be able if he has patches to enable the Direct3D9 renderer.

> Note: That the app can't be resized when in window mode and maximized
> with certain composers, such as Kwin.

Is "can't" a typo? I've tested locally on KDE with Kwin, and I can resize the window at will, everything stretches just fine to the custom size of my choice (and that's awesome!). Models end up stretched too of course, but it's easy to go back to a "supported" resolution via the settings.

hwoarangmy:
Works well with MinGW. I'm testing with VS2013 with OpenGL and D3D. Note that we should adapt so that changing resolution should display more tiles instead of stretching. I confirm live resizing works for me, too.

hwoarangmy:
I've tested and I confirm there is a problem when changing the renderer. It is changed in the menu but when I restart, OpenGL is selected again. Since I guess I will have to debug it myself (since I'm the only one using it), I merge and open an issue

Bertram25:
Yep, see my comment in #799 about why we need custom user config.

Bertram25:
> Is "can't" a typo? I've tested locally on KDE with Kwin, and I can resize the window at will, everything stretches just fine to the custom size of my choice (and that's awesome!)

No, it wasn't ;) On kde in my system, when the window is maximized, OD won't resize for me, but the setting is applied as unmaximizing leads to the correct size. Fullscreen mode is not impacted.

PR #798, opened 2015-06-09
Several small fixes/improvements

hwoarangmy:
Note that I've only added the minimap camera for testing purpose (it is too slow). But if someone wants to mess with LOD, it will help ^^

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -294,6 +294,7 @@ set(OD_SOURCEFILES
     ${SRC}/gamemap/GameMap.cpp
     ${SRC}/gamemap/MapLoader.cpp
     ${SRC}/gamemap/MiniMap.cpp
+    ${SRC}/gamemap/MiniMapCamera.cpp
     ${SRC}/gamemap/TileContainer.cpp
     ${SRC}/gamemap/TileSet.cpp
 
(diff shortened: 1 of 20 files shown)


Bertram25:
Pretty excellent stuff done tonight. :D

Nothing to say about the code so far. I've got questions though:
Does the new minimap improve FPS? (I'll of course try it myself)
You didn't enable it because you wanted feed-back?

hwoarangmy:
because I wanted feedback and because it is laggy. I've barely tried but on a 400x400, I guess it will just be unplayable

Bertram25:
I tested everything but the needs-workshop message. To me, it's mergeable. Great job. :)

As for the new camera type, I wondered:
- Could we use the same view ratio than for the "old" minimap? (But this would mean adding movement logic to it) This could help with perf.
- Maybe we could also create low-poly LOD for models used by the minimap camera? http://www.ogre3d.org/tikiwiki/MeshLod

Just hints. I'm still searching about that myself. :)

regards,

hwoarangmy:
Yes, I've had a look. IMHO, it would be better to see the whole map in the minimap. However, for performance reasons, we can do something else. I guess LOD could be good and now, there is a good base to try ^^

akien-mga:
"Small fixes", you bet, this is huge :-P Pretty good stuff coming in over the last days :)

PR #800, opened 2015-06-11
Allow changing renderer in the settings window + integrate HLSL D3D renderer for windows

### LINKED ISSUE #799: Cannot change to D3D renderer
By default, OpenGL renderer is selected (which is fine). But when we go to the settings and choose D3D renderer, nothing happens and after restarting, OpenGL renderer is selected again.

hwoarangmy:
fixes #799

### DIFF
--- a/materials/RTShaderLib/HLSL/FFPLib_Common.hlsl
+++ b/materials/RTShaderLib/HLSL/FFPLib_Common.hlsl
@@ -1,0 +1,249 @@
+/*
+-----------------------------------------------------------------------------
+This source file is part of OGRE
+(Object-oriented Graphics Rendering Engine)
+For the latest info, see http://www.ogre3d.org
+
+Copyright (c) 2000-2014 Torus Knot Software Ltd
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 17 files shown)


Bertram25:
About the code, I think it's a way to handle that right away, but in term of user experience, I'd rather make it so OD ask the user whether he agrees to restart right away before restarting.
I can have a go at adding a pop up after your change if you're ok with it.

We'd still have to develop a custom config setting IMHO as sound settings and one day keys, etc will need that.

Bertram25:
I can't test on linux but it compiles fine and the D3D renderer isn't available on linux as expected, so I'll merge.

hwoarangmy:
> in term of user experience, I'd rather make it so OD ask the user whether he agrees to restart right away before restarting.

I agree. I just didn't feel like doing it ^^

> I can have a go at adding a pop up after your change if you're ok with it.

Yes, please.

> We'd still have to develop a custom config setting IMHO as sound settings and one day keys, etc will need that.

Indeed

Bertram25:
> I agree. I just didn't feel like doing it ^^

Don't worry, you don't have to do everything. I know I'm slow but I'll stick around killing my tasks little by little. :)
Ok, I'll add a popup.

Then, I'll be on user config support.

Issue #530, opened 2015-02-28
Minimap: It is slow and badly rendered


Bertram25:
Some code around the minimap is slow.
I don't know yet whether it's the pixelBox update or the swapping/rendering itself. @oyvindln will tell. ;)

Second thing, it is likely to use the GL_NEAREST method to magnify the texture and this is looking quite ugly in game.
We should have a look at making the minimap texture have smooth edges.

I'm setting this to 0.5.0 but if it becomes to hard to improve, we'll move it to later.

Bertram25:
After looking at the code, the pixels are drawn square per squares after applying the view rotation, explaining why it can be so ugly.
To fix this, the pixels should be rendered unrotated and the texture should be rendered rotated.

Bertram25:
Just had another idea, if we draw the texture unscaled (with a grain of 1) and let cegui rescale it, we may have the desired effect. I'll try it.

oyvindln:
It's also possible to rotate the CEGUI window, rather than drawing rotated. I don't how that would affect performance though.

Bertram25:
> It's also possible to rotate the CEGUI window, rather than drawing rotated. I don't how that would affect performance though.

I've tried, unfortunately, it looks rather poor. Btw, I managed to make the texture be rendered unscaled and zoom it but it also looks poor as it is too smoothed.
I'll thus try the reverse now: Draw the texture at a bigger size than required and make it smaller it when rendered.

oyvindln:
As for performance, it may make sense to draw directly to the texture instead of via a buffer, at least if we drop trying to render 4x4 blocks as we are now. Otherwise I think data locality could be an issue when it comes to performance. Changing the temporary buffer to be a single vector may help a bit, and should be quite straight forward, but on the tile container side that would be much bigger task. The tiles are currently allocated individually on the heap, which may not be ideal performance-wise.

Bertram25:
> It's also possible to rotate the CEGUI window, rather than drawing rotated. I don't how that would affect performance though.

After thinking about this, I wondered whether adding a sub-image to the map background image widget and rotating that one would be feasible. This way, the rotated map, as a child of the map baground image would be scissored correctly when rotated. I may look at that sooner or later when time permits.

oyvindln:
One option for later could also be to simply use renter-to-texture with an isometric camera instead of hand-drawing the minimap.

Bertram25:
[...]

PR #803, opened 2015-06-12
Display MiniMap Camera in the minimap

hwoarangmy:
part of #530 

With this PR, I've let the camera minimap by default. If requested, I can restore the old one before we merge this PR.

### DIFF
--- a/source/gamemap/MiniMap.cpp
+++ b/source/gamemap/MiniMap.cpp
@@ -111,8 +111,11 @@ Ogre::Vector2 MiniMap::camera_2dPositionFromClick(int xx, int yy)
 
 void MiniMap::update(Ogre::Real timeSinceLastFrame)
 {
-    updateCameraInfo(mCameraManager.getCameraViewTarget(),
-          mCameraManager.getActiveCameraNode()->getOrientation().getRoll().valueRadians());
+    Ogre::Vector3 vv = mCameraManager.getCameraViewTarget();
+    double rotation = mCameraManager.getActiveCameraNode()->getOrientation().getRoll().valueRadians();
+    mCamera_2dPosition = Ogre::Vector2(vv.x, vv.y);
+    mCosRotation = cos(rotation);
+    mSinRotation = sin(rotation);
 
     for (int ii = 0, mm = mCamera_2dPosition.x - mWidth / (2 * mGrainSize); ii < static_cast<int>(mWidth); ++mm, ii += mGrainSize)
     {
--- a/source/gamemap/MiniMap.h
+++ b/source/gamemap/MiniMap.h
@@ -119,13 +119,6 @@ private:
         *pDest++ = RR;  //R
     }
 
-    inline void updateCameraInfo(Ogre::Vector3 vv, double rotation)
-    {
-        mCamera_2dPosition = Ogre::Vector2(vv.x, vv.y);
-        mCosRotation = cos(rotation);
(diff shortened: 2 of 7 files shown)


akien-mga:
It's awesome :-D
The minimap is a bit laggy, but it does impact the main window at all from what I could see, so it's a minor disagreement compared to the nice looking camera minimap :)

What would be your ideas to improve the performance even more? Would it help to ignore lighting in the camera minimap?

Bertram25:
The code looks good. I'll test asap.

If the main game performance isn't hampered by the new camera (compared to the old one), I'd say let's drop the old one.
What we always wanted is this kind of minimap anyway.

Bertram25:
> What would be your ideas to improve the performance even more? Would it help to ignore lighting in the camera minimap?

Correct me if I'm wrong but AFAIK, the objects and lights are managed (rendered) per scene.
This means we may have to create a new scene to display things differently.
I've no idea of how it would impact performance or if there is something else to do about the scene, though.

Maybe there is another trick?

hwoarangmy:
Note that on my computer, the keeper hand blinks while on CEGUI windows while minimap is rendered (every 0.5 seconds). I guess that's the needed time to render the minimap.

> What would be your ideas to improve the performance even more?

To use custom LOD (Level Of Details - allows to display simpler meshes when the camera is far) and to display the whole map (as was the minimap camera before this PR)

> The minimap is a bit laggy

That's because it is refreshed every 0.5 seconds (that can be increased if performances allow - we might even add some config option about it).

> Maybe there is another trick?

I've seen something about implementing Ogre::RenderTargetListener. There is 1 function called before the minimap is rendered and another when it is done. The trick is to set unwanted stuff not visible before it is rendered and restore the state after.
I've seen that here:
http://www.ogre3d.org/tikiwiki/tiki-index.php?page=Intermediate+Tutorial+7

hwoarangmy:
I've tried disabling light and it works. However, there is no visible improvement (Keeper hand is still blinking on CEGUI windows)

Danimal696:
i demand a pic of it! or else no kobold!!!

akien-mga:
od-minimap
It's small but hey, it's a minimap ;)

hwoarangmy:
> i demand a pic of it! or else no kobold!!!

You will have a release during the week end so, give the kobold !!!!

hwoarangmy:
BTW, I've tried the D3D renderer and the old minimap is buggy (seems like drawing do not work with this renderer). But the minimap camera works well ^^

Danimal696:
i actually liked the old one, can they be toggled by user with an option (on config button, not while playing)? it would be nice to be able to choose between the two

hwoarangmy:
Yes. I've not removed the old one

Bertram25:
As far as I'm concerned, this is mergeable. Nice job. :)

> i actually liked the old one, can they be toggled by user with an option (on config button, not while playing)? it would be nice to be able to choose between the two

I think I'd rather have a key to switch between different minimaps and even hide it. Something like, using the M key to:
- 1. Show the minimap camera only a few tiles.
- 2. Show the minimap camera the whole map.
- 3. Show the old minimap.
- 4. Hide it
  ...

The old map will still have to die somewhen, IMHO, as we shouldn't support too much redundant functionalities.

WDYT?

akien-mga:
@Bertram25's proposal sounds good to me :)

Issue #810, opened 2015-06-16
Claimed wall textures are much too big


hwoarangmy:
The consequence is that when the first wall is claimed (and the mesh loaded), the game freezes.

Seems like I was too fast to integrate #808 

@Danimal696 Could you please reduce a bit its size ? 2048x2048 is too much...

I'm pretty sure that 512x512 would be enough. I also bet that's a reason why the game freezes when we build a room (because of loading textures for ground tiles as well as the building objects).

akien-mga:
Note that the 2048x2048 version could go in the -media-source repo, while we use a lower res version in the game.

Danimal696:
1024 will the least we can afford if we want them to have some quality, have in mind theres a lot of them packed into the same png

Danimal696:
please try now, on my pc both worked perfectly

hwoarangmy:
ok. On my way

hwoarangmy:
I've tried. It is faster but you also need to resize ClaimedWallMask.png and ClaimedWallNormal.png.
However, I still have a small lag just when the tile is claimed (and my computer is decent: i7 with GeForge GTX 560M).
I've tried with 512x512 pngs and got no lag but the quality is clearly lower.

akien-mga:
Could we split those textures into several files? I guess the issue is that the file itself is too big, and several 512x512 files might help circumvent this hardware issue.

Danimal696:
> Could we split those textures into several files?

When i did this one i wanted to avoid exactly that, it would help when adding diferent biomes to be into the same file, but im going to just cut them all.

> I still have a small lag just when the tile is claimed (and my computer is decent: i7 with GeForge GTX 560M).

I have a I5 and AMD Radeon HD 7700 and see no lag at all, nvidia drivers migth be the culprit in your case

hwoarangmy:
> I have a I5 and AMD Radeon HD 7700 and see no lag at all, nvidia drivers migth be the culprit in your case

Well, the main reason is that I use a debug version which is much slower. However, its pretty nice to see the bottlenecks and the problems that may happen on weaker configurations

Danimal696:
This is very important, are claimed walls harcoded to only use Claimedwall.material?
no matter what i change on the model or material it always uses whats in Claimedwall.material, and its making impossible to make the tiles piece by piece

Danimal696:
[...]

PR #811, opened 2015-06-16
resized walls textures to 1024*1024

Danimal696:
hopeful fix for #810


Danimal696:
er, crap, i messed it, can you cherry pick :
resized walls textures to 1024*1024
Fixed other texture sizes

ill need to erase my repo to fix this (i think?) so please tell me

PR #818, opened 2015-06-19
New tileset  + fixed textures

Danimal696:
i redid it so every tile has an unique 256*600 texture, this should
avoid the memory bottleneck and is also infenitely more artist friendly
than single multitextured material


hwoarangmy:
I've got no lag with this new version ^^

Danimal696:
i guess next is fixing gold and water lava/rotation; im almost tempted to go for a new tileset already

Bertram25:
I'm trying. :)

Bertram25:
@hwoarangmy 
Just in case, I don't know this is linked to a crash in OpenAL or not:

```
Server sends newturn=2065
Client (Player) received turnStarted=2065
CLIENT (Player) - Removing rendered object RenderedMovableEntity_Farm8_1988,MeshName=Chicken
CLIENT (Player) - Removing rendered object RenderedMovableEntity_Farm8_1991,MeshName=Chicken
CLIENT (Player) - Adding rendered object RenderedMovableEntity_Crypt14_1997,MeshName=SmallSpider
Computing turn 2066, timeSinceLastTurn=0.692534
Creature=Kobold6 RIP
> 20 loops in Creature::doUpkeep name:BigKnight173 seat id: 3. Breaking out..
> 20 loops in Creature::doUpkeep name:BigKnight183 seat id: 3. Breaking out..
Creature=Kobold199 RIP
signal 11 (Segmentation fault), address is 0x50 from 0x4d24f0

[bt]: (1) ./opendungeons : MissileObject::doUpkeep()+0x1f0 [0x4d24f0]
[bt]: (2) ./opendungeons : MissileObject::doUpkeep()+0x1f0 [0x4d24f0]
[bt]: (3) ./opendungeons : GameMap::doMiscUpkeep(double)+0x395 [0x52c7c5]
[bt]: (4) ./opendungeons : GameMap::doTurn(double)+0x473 [0x536a83]
[bt]: (5) ./opendungeons : ODServer::startNewTurn(double)+0x372 [0x5c12f2]
[bt]: (6) ./opendungeons : ODServer::serverThread()+0x1c1 [0x5d3cc1]
[bt]: (7) /usr/lib/x86_64-linux-gnu/libsfml-system.so.2 : +0x6828 [0x7f76cf0ab828]
[bt]: (8) /lib/x86_64-linux-gnu/libpthread.so.0 : +0x80a4 [0x7f76d11ff0a4]
[bt]: (9) /lib/x86_64-linux-gnu/libc.so.6 : clone+0x6d [0x7f76ce3a304d]
AL lib: (WW) FreeContext: (0x1bbe150) Deleting 48 Source(s)
An internal OpenAL call failed in SoundSource.cpp (181) : AL_INVALID_OPERATION, the specified operation is not allowed in the current state
x50
...
ON, the specified operation is not allowed in the cuate
opendungeons: /build/openal-soft-afkAHt/openal-soft-1.16.0/Alc/ALc.c:779: LockLists: Assertion `lockret == althrd_success' failed.
signal 11 (Segmentation fault), address is 0 from 0x7f76ce335b82
```

@Danimal696 I've tried the new tileset. I think technically you've found a good compromise (even if I'll attempt to make multitexturing material work just in case, who knows?)
You'll want to kill me, but I'll say it anyway: ;P
Why did you use bricks for the vertical walls? I looks anachronistic and not fitting the rest of the walls to me. Why not a sturdy stone wall instead?

Bertram25:
In any case, it's mergeable to me. :)

Danimal696:
> Why did you use bricks for the vertical walls? I looks anachronistic and not fitting the rest of the walls to me. Why not a sturdy stone wall instead?

I kinda get what you mean, but this look is the closest to original DK, no reason not to change it in the long run if a better looking texture is found. There also the factor that we are way too used to the last tileset and anything different feel weirds.

Bertram25:
> There also the factor that we are way too used to the last tileset and anything different feel weirds.

Fair enough, I promise you I'll give it time. (But I'll let my eyes open about other potential textures ;])

route1
Also, do you think it is possible to do something on the claimed walls models to more or less avoid this glitch?

Danimal696:
not an easy one at least; @hwoarangmy  is there a way to set up this kind of connector in tileset.config?

Bertram25:
> not an easy one at least; @hwoarangmy is there a way to set up this kind of connector in tileset.config?

There is more to that. Do you guys think we should auto-claim corners when claiming walls for instance?

hwoarangmy:
> Just in case, I don't know this is linked to a crash in OpenAL or not

Looking at the code, I don't really see how we can have a segfault there. It might be interesting to display line numbers in the callstack. I will have a closer look. Moreover, I will try to add some logs to see what's happening if we reproduce. Is it the first time you see this bug ? Do you have the replay ?

> not an easy one at least; @hwoarangmy is there a way to set up this kind of connector in tileset.config?

No, we cannot do that with the current tileset.

> There is more to that. Do you guys think we should auto-claim corners when claiming walls for instance?

That might be a way to fix the visual glitch

Bertram25:
Let's merge. I'll open an issue about the corner. 
:)

PR #820, opened 2015-06-22
Implemented listeners on GameEntity

hwoarangmy:
That allows to listen to what happens to an entity and be notified when something happens. It can, for example, be used on missiles to notify the sender that the missile hit (for counting killed enemies for example).


Bertram25:
Code looks good + some extra fixes if I saw well. ;)

I'll try asap.

hwoarangmy:
> some extra fixes if I saw well

I had forgotten but seems like you saw well ^^

Bertram25:
Works great! :) Mergeable as for me.

akien-mga:
I don't see any obvious regression, so let's merge :)

Issue #753, opened 2015-05-15
GUI: Improve the sliders visuals.


Bertram25:
Atm, the sliders visual are the poorest in the Skin.
I need to improve that.

Bertram25:
From @akien-mga 

> I find the slider a bit awkward to handle, sometimes it's hard to grab to actually make it slide, but this will be fine tuned later on.

akien-mga:
@Bertram25 Do you still want to handle this for 0.6.0, or would you prefer to leave it like that for now?

Bertram25:
I don't know whether I'll have the time to. Let's move it to 0.7 and if I do it before, we'll change again.

PR #825, opened 2015-06-22
Kept on settings.

Bertram25:
- Added support for spaces in parameter names.
- Improved the behaviour of the music volume widget.

Teaser: The next PR will concentrate on moving video renderer's options into settings... ;)

### DIFF
--- a/source/modes/SettingsWindow.cpp
+++ b/source/modes/SettingsWindow.cpp
@@ -83,6 +83,16 @@ SettingsWindow::SettingsWindow(CEGUI::Window* rootWindow):
         )
     );
 
+    // Music volume slider
+    CEGUI::Slider* volumeSlider = static_cast<CEGUI::Slider*>(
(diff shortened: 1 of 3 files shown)


hwoarangmy:
Code seems good to me. @akien-mga  I won't be around for some days so feel free to merge without waiting for me ^^

akien-mga:
It works fine :) I find the slider a bit awkward to handle, sometimes it's hard to grab to actually make it slide, but this will be fine tuned later on. When sliding, the actual value should probably rounded to an int or to a float with one decimal, currently it shows many decimals that are not really relevant.

Bertram25:
>  I find the slider a bit awkward to handle, sometimes it's hard to grab to actually make it slide, but this will be fine tuned later on.

Yeah, I don't like it too. Plus, I find it ugly. I'll add your note to #753 

> When sliding, the actual value should probably rounded to an int or to a float with one decimal, currently it shows many decimals that are not really relevant.

True. My bad. I'll fix this asap. :)

PR #826, opened 2015-06-28
Moved to full custom settings support.

### LINKED ISSUE #822: Video renderer's options should be auto-added to the videosettings menu.
As discussed a bit earlier with @hwoarangmy , when choosing/changing a renderer, the settings should list all the renderer's options and added them with default values in the video settings + auto-create widgets on need there.

This way, one will be able to fully configure the video renderer based on its capabilities.

### LINKED ISSUE #823: Move to full custom config support.
For now, we're using ogre.cfg to configure Ogre at startup. The goal of this issue is to use our user config support only.

Bertram25:
- Save and display custom renderer settings. Fixes #822 
- Dropped ogre.cfg need. Fixes #823 

The renderer should start with default settings at first run, and save them once you modify settings.
When changing the renderer settings, all the video settings should be reset.

Please test, especially on Windows. ;)

### DIFF
--- a/gui/WindowSettings.layout
+++ b/gui/WindowSettings.layout
@@ -2,7 +2,7 @@
 
 <GUILayout version="4" >
     <Window type="OD/FrameWindow" name="SettingsWindow" >
-        <Property name="Area" value="{{0,205},{0,128},{0,669},{0,481}}" />
+        <Property name="Area" value="{{0,205},{0,128},{0,669},{0,600}}" />
         <Property name="Text" value="Settings" />
         <Property name="SizingEnabled" value="False" />
         <Property name="AlwaysOnTop" value="True" />
@@ -75,6 +75,8 @@
                     <Property name="Text" value="" />
                     <Property name="VerticalSlider" value="False" />
                     <Property name="AutoRenderingSurface" value="True" />
+                    <Property name="MaximumValue" value="100" />
+                    <Property name="ClickStepSize" value="5" />
                 </Window>
             </Window>
         </Window>
--- a/gui/WindowSettings.layout
+++ b/gui/WindowSettings.layout
@@ -79,6 +79,24 @@
                     <Property name="ClickStepSize" value="5" />
                 </Window>
             </Window>
+            <Window type="DefaultWindow" name="Game" >
+                <Property name="Area" value="{{0,0},{0,0},{1,0},{1,0}}" />
+                <Property name="Text" value="Game" />
+                <Property name="MaxSize" value="{{1,0},{1,0}}" />
+                <Property name="Visible" value="False" />
+                <Property name="RiseOnClickEnabled" value="False" />
+                <Window type="OD/StaticText" name="NicknameText" >
+                    <Property name="Area" value="{{0,32},{0,40},{0,150},{0,60}}" />
+                    <Property name="MaxSize" value="{{1,0},{1,0}}" />
+                    <Property name="Text" value="Nickname:" />
+                    <Property name="BackgroundEnabled" value="False" />
+                    <Property name="FrameEnabled" value="False" />
+                </Window>
+                <Window type="OD/Editbox" name="NicknameEdit" >
+                    <Property name="Area" value="{{0,160},{0,40},{0,320},{0,60}}" />
+                    <Property name="Text" value="" />
+                </Window>
+            </Window>
         </Window>
     </Window>
 </GUILayout>
--- a/source/ODApplication.cpp
+++ b/source/ODApplication.cpp
@@ -100,15 +100,14 @@ void ODApplication::startClient()
     Random::initialize();
     //NOTE: The order of initialisation of the different "manager" classes is important,
     //as many of them depend on each other.
-    OD_LOG_INF("Creating OGRE::Root instance; Plugins path: " + resMgr.getPluginsPath()
-        + "; config file: " + resMgr.getOgreCfgFile());
+    OD_LOG_INF("Creating OGRE::Root instance; Plugins path: " + resMgr.getPluginsPath());
 
-    Ogre::Root ogreRoot(resMgr.getPluginsPath(),
-                        resMgr.getOgreCfgFile());
+    // N.B: We don't use any ogre.cfg file, hence setting the file path value to "".
+    Ogre::Root ogreRoot(resMgr.getPluginsPath(), "");
 
     ConfigManager configManager(resMgr.getConfigPath(), resMgr.getUserCfgFile());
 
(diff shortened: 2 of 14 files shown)


akien-mga:
It works great on Linux with the OpenGL renderer. I haven't checked if all options work perfectly in game, as I don't understand what half of them are supposed to do :p

I've only found one issue: while the game is in fullscreen, if you untick fullscreen, and then change the resolution to a lower one before applying, the resolution change is not taken into effect. The game just goes in windowed mode but with the initial resolution.
On the other hand, if you change the resolution first, and only then untick the fullscreen box, it works fine. So I guess the resolution change is lost somewhere if you start with the fullscreen box.
After a restart, both methods yield the windowed mode with the proper resolution, so the config file is modified properly.

Bertram25:
> It works great on Linux with the OpenGL renderer. I haven't checked if all options work perfectly in game, as I don't understand what half of them are supposed to do :p

Don't worry, me neither. XD

> I've only found one issue: while the game is in fullscreen, if you untick fullscreen, and then change the resolution to a lower one before applying, the resolution change is not taken into effect. The game just goes in windowed mode but with the initial resolution.
> On the other hand, if you change the resolution first, and only then untick the fullscreen box, it works fine. So I guess the resolution change is lost somewhere if you start with the fullscreen box.

Funny, it works on my end, but only if the window isn't maximized. Can you make sure of not letting the window "touch" any border before fullscreen and try again?

akien-mga:
> Funny, it works on my end, but only if the window isn't maximized. Can you make sure of not letting the window "touch" any border before fullscreen and try again?

Going fullscreen works fine. It's leaving fullscreen mode + setting a different resolution that does not work as expected, in the case where you untick fullscreen first, then change the resolution, then apply.

Bertram25:
> Going fullscreen works fine. It's leaving fullscreen mode + setting a different resolution that does not work as expected, in the case where you untick fullscreen first, then change the resolution, then apply.

k, I trust you upon that. (?) I'll have a closer look at the code asap.
(In the meanwhile, I've worked on nickname support. ;])

akien-mga:
> k, I trust you upon that. (?) I'll have a closer look at the code asap.

I think you should be able to reproduce it, it don't see why it would be hardware dependent. Step by step:
- Start OD in fullscreen mode (fullscreen ticked on) and with your monitor's highest native resolution (here I'm using 1920x1080).
- In the settings:
  - Untick the "Fullscreen" option first
  - Switch the resolution to a lower res one (e.g. 800x600)
  - Click "Apply"
- The game should then go to windowed mode, but the resolution should stay the one of your monitor instead of the lower one you specified.

Note that I don't seem to reproduce the issue if I start with fullscreen mode on but a lower resolution, like 800x600.

Bertram25:
Ok. Note that my own highest resolution is 1366x768. It's the same for me since the beginning in fact.
That's what I meant that setting the resolution won't work when the window is "maximized".

> On the other hand, if you change the resolution first, and only then untick the fullscreen box, it works fine. So I guess the resolution change is lost somewhere if you start with the fullscreen box.

Does this still work when starting from 1920x1080 for instance? To me, it can't come from my code, either from CEGUI or from some composer magic?

akien-mga:
> Does this still work when starting from 1920x1080 for instance? To me, it can't come from my code, either from CEGUI or from some composer magic?

Well the test case is:
- Start in fullscreen with (native) 1920x1080 resolution
- Untick "Fullscreen"
- Switch to a different resolution
- Apply and see that it goes in windowed mode (as expected, but the resolution is still 1920x1080). It will however be the chosen resolution on reboot.

But that's a really minor issue :) Let's merge, and I'll investigate more later on.

hwoarangmy:
Nice improvements. I will have  look and tell you if I can reproduce the bug @akien-mga saw

Bertram25:
> I will have look and tell you if I can reproduce the bug @akien-mga saw

Yeah, @akien-mga must wonder about my mental health. XD

Anyway, @akien-mga, could you tell me whether those two test cases are correct?
1. The one you just told above:
   - Start in fullscreen with (native) 1920x1080 resolution
   - Untick "Fullscreen"
   - Switch to a different resolution
   - Apply and see that it goes in windowed mode 
     **Result:** Windowed mode, but the resolution is still 1920x1080). It will however be the chosen resolution on reboot.
2. The other one:
   - Start in fullscreen with (native) 1920x1080 resolution
   - Switch to a different resolution
   - Untick "Fullscreen" (Note the swap between unticking fullscreen and switching the resolution)
   - Apply and see that it goes in windowed mode 
     **Result:** Windowed mode, with the correct resolution.

Trying to narrow down things on my end. :)

hwoarangmy:
I've made some tries and couldn't reproduce this issue. I remember having seen something pretty similar while playing with resolutions during our first tries with handling video settings but I cannot reproduce. I will make some tries with D3D renderer to check if it changes something ASAP.

@akien-mga If you confirm the bug on the current dev branch, could you please open an issue with a link here and the description you have already done ?

@Bertram25 Note that I have a cegui failing assert when I press cancel button in the settings windows. It should not be too hard to hunt but could you please try if you reproduce ?

Bertram25:
> @Bertram25 Note that I have a cegui failing assert when I press cancel button in the settings windows. It should not be too hard to hunt but could you please try if you reproduce ?

Sure. :) Thanks for the feedback.

Bertram25:
@hwoarangmy Is it something like this?

> ERROR: opendungeons/source/utils/ConfigManager.cpp line 1555 Unknown parameter param=Nickname

It is because I was a bit too confident with warnings. I'll remove them as it is possible to have non-existing parameters at least at the beginning.

Bertram25:
@hwoarangmy 

> I've made some tries and couldn't reproduce this issue.

Just in case, I believe it's a KDE issue (or feature?).

hwoarangmy:
> @hwoarangmy Is it something like this?

No, it's a real assert when destroying the window. If you don't reproduce, I will have a deeper look even if I don't see why my changes would have something to do with settings window.

> Just in case, I believe it's a KDE issue (or feature?).

ok. Anyway, that's not such a big issue, so...

hwoarangmy:
I've had a deeper look and it is because when we click cancel, there is a call to SettingsWindow::initConfig.

In this function, mCustomVideoComboBoxes is cleared. The problem is that we try to remove them from mSettingsWindow while their real parent should be mSettingsWindow->getChild("MainTabControl/Video").

Note that there is something similar with mCustomVideoTexts.

Bertram25:
Ok, I'll have a look at that + better handle missing parameters this Week-end. :) Thanks for all the feedback.

PR #829, opened 2015-07-02
Removed non-existing user parameter warnings as it is a possibility.

Bertram25:


Bertram25:
> Are you sure there is no way to avoid requesting non-existing values ?

I'm on limiting the cases, and I'll turn the error into an info.

hwoarangmy:
@Bertram25 If the values are not mandatory, why don't you add some:
bool ConfigManager::hasVideoValue(const std::string& param) const
for optional values ?

Bertram25:
Good idea. I'm still trying to limit other cases, though.

hwoarangmy:
> I'm still trying to limit other cases, though.

Yeah, IMHO, it's always better to not remove error logs too fast. It saved me lot of time debugging thanks to those errors when you are sure to have written correctly the parameter name but forgot an upper char ^^

Bertram25:
eh eh. So true. :)

PR #831, opened 2015-07-03
Implemented new library logic

hwoarangmy:
With this PR:
- Library now only releases books that represents a fixed amount of points
- When enough points are gathered, the currently pending research is unlocked
- Researches pending order can be changed. If you click a pending research, it will be canceled. If you click a non pending research, it will be queued
- Note that if you click a research that has required researches not pending, they will be automatically selected
- If a research is completed while the research tree is displayed, the research tree should be correctly adapted ( @akien-mga I'm looking at you right now ;-) )
- A research can still be dropped by a creature (can be tested in the skirmish small test map 2v2). But now, it is a gift box ( @Danimal696 note that I've let the mesh be easily changeable so we can later use the grimoire if preferred but right now, it was an occasion to introduce the gift box mesh and ATM, particle effects can only be used on creatures)
- Note that if you click a research with a dependency not allowed, it won't be selected (can be tested in the skirmish small test map 2v2 by selecting boulder trap). However, the corresponding research is not unavailable because we could later unallow a research to have the player kill some creature that will release the research and unlock the tree)

### DIFF
--- a/source/game/Seat.h
+++ b/source/game/Seat.h
@@ -212,6 +212,15 @@ public:
     inline void setPlayerType(const std::string& playerType)
     { mPlayerType = playerType; }
 
+    inline const std::vector<ResearchType>& getResearchDone() const
+    { return mResearchDone; }
+
+    inline const std::vector<ResearchType>& getResearchPending() const
+    { return mResearchPending; }
+
+    inline const std::vector<ResearchType>& getResearchNotAllowed() const
+    { return mResearchNotAllowed; }
+
     void setPlayer(Player* player);
 
     void addAlliedSeat(Seat* seat);
@@ -266,15 +275,9 @@ public:
     bool isResearching() const
     { return mCurrentResearch != nullptr; }
 
-    //! \brief Gets the current research being done
-    ResearchType getCurrentResearchType() const;
-
     //! \brief Tells whether the given research type is in the pending queue.
     //! \return The number of the pending research in the research queue or 0 if not there.
     uint32_t isResearchPending(ResearchType resType) const;
-
-    //! \brief Tells whether a research entity is existing for the given ResearchType.
-    bool hasResearchWaitingForType(ResearchType resType);
 
     //! \brief Checks if the given spell is available for the Player. This check
     //! should be done on server side to avoid cheating
@@ -292,10 +295,6 @@ public:
     //! otherwise
     bool isResearchDone(ResearchType type) const;
 
-    //! \brief Tells the server and client a new research entity is waiting to be brought
-    //! to the temple.
-    void addResearchWaiting(ResearchType type);
-
     //! Called when the research entity reaches its destination. From there, the researched
     //! thing is available
     //! Returns true if the type was inserted and false otherwise
@@ -304,7 +303,7 @@ public:
     //! Called from the library as creatures are researching. When enough points are gathered,
     //! the corresponding research will become available.
     //! Returns the research done if enough points have been gathered and nullptr otherwise
-    const Research* addResearchPoints(int32_t points);
+    void addResearchPoints(int32_t points);
 
     //! Used on both client and server side. On server side, the research tree's validity will be
     //! checked. If ok, it will be sent to the client. If not, the research tree will not be
@@ -470,9 +469,6 @@ private:
     //! \brief Researches already done. This is used on both client and server side and should be updated
     std::vector<ResearchType> mResearchDone;
 
-    //! \brief Research done but awaiting to be brought to the temple.
-    std::vector<ResearchType> mResearchWaiting;
-
     //! \brief Researches pending. Used on both client and server side and should be updated.
     std::vector<ResearchType> mResearchPending;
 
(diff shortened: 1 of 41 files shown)


Bertram25:
Another monster to tame? ;) I'll have a look asap. :)

hwoarangmy:
> Another monster to tame? ;)

No worries, you have been through worst ;-)

Bertram25:
@hwoarangmy Btw, thanks for doing my job :S I can't wait to test all that. :)

Bertram25:
Found something:
On the research window, let's say you have 5 researches. If you click on the 4th, both 4 and 5 will disappear, which is ok. But if you click on the previous 5 location, both 4 and 5 reappear.

Ah, or maybe it was this:

> Note that if you click a research that has required researches not pending, they will be automatically selected

Nevermind.

hwoarangmy:
> Ah, or maybe it was this

Yep. If 5 disapears when you click 4, that means that 5 depends on 4. And if you add 5 again, 4 will be automatically selected. That might allow us to easily build a tree for AI if we want (by choosing an advanced research and getting all its dependencies or picking basic ones) ^^

source/game/Seat.h
```diff
@@ -304,7 +303,7 @@ class Seat
     //! Called from the library as creatures are researching. When enough points are gathered,
     //! the corresponding research will become available.
     //! Returns the research done if enough points have been gathered and nullptr otherwise
```
Bertram25, line 305: For info, the return is now void.

source/game/SeatFunctions.cpp
```diff
@@ -1107,147 +1097,63 @@ uint32_t Seat::isResearchPending(ResearchType resType) const
...
-        }
+        if(research->getType() != researchType)
+            continue;
 
-        // Tell the Research GUI it needs refreshing
-        mGuiResearchNeedsRefresh = true;
+        mCurrentResearch = research;
+        break;
     }
+
```
Bertram25, line 1156: I like the logic simplification done here. :)

source/modes/GameMode.cpp
```diff
@@ -1313,11 +1406,13 @@ void GameMode::refreshResearchButtonState(ResearchType resType)
     }
     else
     {
+        // The skill is not available and research is not pending
         guiSheet->getChild(ceguiWidgetButtonName)->hide();
         researchButton->setText("");
         researchButton->setProperty("StateImage", "");
         researchButton->setEnabled(true);
     }
+    // TODO: handle case research is disabled
```
Bertram25, line 1415: Need help on that TODO? Is it more complicated than setEnabled(false) for instance?

source/modes/GameMode.cpp
```diff
@@ -1461,3 +1582,104 @@ void GameMode::handlePlayerActionSelectTile()
...
+
+void GameMode::resetResearchTree()
+{
+    mResearchPending = mGameMap->getLocalPlayer()->getSeat()->getResearchPending();
+    mIsResearchWindowOpen = true;
+}
+
+bool GameMode::researchButtonTreeClicked(ResearchType type)
+{
+    // If the research is already done or not allowed, nothing to do
```
Bertram25, line 1594: Ah, I was sure this was part of game finally. ;)

Bertram25:
A few comments, but overall, this is mergeable to me. :)
@akien-mga Your turn ;)

source/game/Seat.h
hwoarangmy, line 305: Ok, I will fix that

source/modes/GameMode.cpp
hwoarangmy, line 1415: I just forgot to remove it. The case when research is disabled is already handled. I will fix that in my next commit (next PR if nothing to change here)

Bertram25, line 1415: ok, perfect. :)

akien-mga:
On my way to test it :)

akien-mga:
It all works very well, congrats :-D
It's awesome to be able to select which researches one wants to do in priority, now we'll have the opportunity to add tons of interesting skills, spells and rooms to make the gameplay more varied :-)

The only comment that I have is that there is no feedback on the state of advancement of the current research. With the old logic, we used to have a different icon once the grimoire with the skill has been produced and is awaiting a kobold.

With the new logic, maybe we could display how many grimoires are still required to unlock the research (e.g. if you already brought 2 grimoires to the temple out of 3 required). Maybe a small text on the icon like "2 of 3 <grimoire icon>".
This brings the question of whether the "research points" stock is shared between all skills, or if each skill has its own pool (for example if you start researching the workshop, get one grimoire to the temple, and then switch to another research).

Danimal696:
Next release will be a nice  step forward

hwoarangmy:
> With the new logic, maybe we could display how many grimoires are still required to unlock the research (e.g. if you already brought 2 grimoires to the temple out of 3 required). Maybe a small text on the icon like "2 of 3 ".

For that, @Bertram25 asked on CEGUI forums how to do a progressbar button. I don't know if he got answers, however...

Bertram25:
> For that, @Bertram25 asked on CEGUI forums how to do a progressbar button. I don't know if he got answers, however...

I even "hacked" the open part of TorchLight II resources to know more about it. ;)
On torchlight II, they have skill buttons that do exactly that and it seems it is a mix between a progress bar and a button. Probably a button with a progress bar as child widget.
Ident told me it is possible if one set the (I don't remember the exact param name) "giveInputToParent" parameter to true.
One more thing on my todo list... :)

akien-mga:
Does this PR render the `[ResearchPending]` tag optional/obsolete in level definitions?

akien-mga:
> Does this PR render the [ResearchPending] tag optional/obsolete in level definitions?

The editor won't load the level files without those tags, so I guess the answer is "No". We could probably make it optional (i.e. the template is still valid without it, the research queue will just be empty) together with the `[ResearchNotAllowed]` tag, but I suppose for the time being it's better to keep them as mandatory, at least modders will know that they exist :)

PR #834, opened 2015-07-03
Hopefully fixed a few glitches with settings.

### LINKED ISSUE #830: Fix problems around config handling...
From @hwoarangmy 
- [x] Problems when destroying certain children windows in the window settings.

I've had a deeper look and it is because when we click cancel, there is a call to `SettingsWindow::initConfig`

In this function, `mCustomVideoComboBoxes` is cleared. The problem is that we try to remove them from `mSettingsWindow` while their real parent should be `mSettingsWindow->getChild("MainTabControl/Video").`

Note that there is something similar with `mCustomVideoTexts.`
- [x]  There are error on empty config param found with new config and when initing the config in the settings window.
- [x] Upgrade user config getters to support value existence check and potential error reporting. + Merge user config maps into one.

Bertram25:
Should fix #830

### DIFF
--- a/source/modes/MenuModeMultiplayerClient.cpp
+++ b/source/modes/MenuModeMultiplayerClient.cpp
@@ -80,7 +80,10 @@ void MenuModeMultiplayerClient::activate()
 
     CEGUI::Editbox* editNick = static_cast<CEGUI::Editbox*>(mainWin->getChild(Gui::MPM_EDIT_NICK));
     ConfigManager& config = ConfigManager::getSingleton();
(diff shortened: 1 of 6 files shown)


Bertram25:
@hwoarangmy Please tell me whether it improves things for you. :)

Bertram25:
@hwoarangmy Should be ok now. :)

@akien-mga The audio displayed value should be fixed, too. :)

hwoarangmy:
I will test that tonight ^^

akien-mga:
Works fine as far as I can tell :)

hwoarangmy:
Works well, no more warning in logs

PR #837, opened 2015-07-05
WarBanner - mesh for "Call to arms" spell

### LINKED ISSUE #836: Replacing mesh for "Call to arms spell"
Pretty soon
demo1

Danimal696:
Fixes #836.

with an animation called loop


akien-mga:
It looks really nice in-game! :)
I'm not 100% convinced by the current Loop animation where the banner swings to the sides. But we can improve it later on :)

akien-mga:
I'll do a pull request to make it used by the call to war spell.

akien-mga:
What about the credits by the way? I can handle it in my pull request, but I'll need the name(s), license and link if there is one :)

Danimal696:
ah, forgot those (strange, rigth?)

PR #842, opened 2015-07-08
Implemented haste spell

hwoarangmy:


Bertram25:
I'll test asap. :)

N.B: Now that the project is mature, I'd like to add a new soft rule:
When one of us write a todo, he should ask himself whether the todo should maybe be also added as an issue.
Yes, I'm thinking about the research manager at this very moment. ;)
I also say that because once again the project is mature but also because our level of quality is high. :)
IMHO, this will help us investigate design or feature issues sooner instead of burying them in the code.

hwoarangmy:
> When one of us write a todo, he should ask himself whether the todo should maybe be also added as an issue.

Ok, I will create the issue #843

Bertram25:
Testing. :)

Bertram25:
N.B.: I've got this:

```
/opendungeons/source/spells/SpellCreatureHaste.cpp:26:33: fatal error: gamemap/PathFinding.h: Aucun fichier ou dossier de ce type
 #include "gamemap/PathFinding.h"
```

F should be lower-cased. Testing anyway.

Bertram25:
Works like a charm. Another thing is that `materials/textures/hasteflare.png` lacks license information.
(Yes, I don't only do that with @Danimal696  :P)

Bertram25:
Once the compilation is fixed and the file has got license info, it's mergeable as for me. :)

hwoarangmy:
Concerning hasteflare.png, I've done that by merging our flare.png with the game-icons.net icon used for the icon. Who should the credits go for ?

Concerning Pathfinding.h, sorry, that's the second time I do that. Sometimes, code blocks autocompletion doesn't work (especially when there is no cpp file with the same name as the header I guess) and I end up doing typos. I will fix that tonight

akien-mga:
> Concerning hasteflare.png, I've done that by merging our flare.png with the game-icons.net icon used for the icon. Who should the credits go for ?

Looks like flare.png was done by @Bertram25. Maybe we should go the simple way and just give credits to "OD Team" for simpler textures like that if that's fine with you.

By the way, icons from game-icons.net are licensed under CC BY 3.0, so we are supposed to give credits to the icon designers too. Should we add simply "game-icons.net" as credits for OpenDungeonsIconSet.png?

akien-mga:
The spell works great btw :) Maybe it should also be added to the default research queue?

hwoarangmy:
> Maybe it should also be added to the default research queue?

It is not added on purpose. Now that the research queue can be changed, we need to choose how we want that to be handled by default. I will open an issue for that.
Because at the moment, adding a new spell requires to change every level if we want to add it to the queue...

Bertram25:
> Looks like flare.png was done by @Bertram25. Maybe we should go the simple way and just give credits to "OD Team" for simpler textures like that if that's fine with you.

Yes.

> By the way, icons from game-icons.net are licensed under CC BY 3.0, so we are supposed to give credits to the icon designers too. Should we add simply "game-icons.net" as credits for OpenDungeonsIconSet.png?

Yes (bis) :)

Bertram25:
> It is not added on purpose. Now that the research queue can be changed, we need to choose how we want that to be handled by default. I will open an issue for that.
> Because at the moment, adding a new spell requires to change every level if we want to add it to the queue...

If I get it, you want to know whether it's best to tell what's not allowed in levels, rather than what's allowed, right?

hwoarangmy:
> you want to know whether it's best to tell what's not allowed in levels, rather than what's allowed, right?

No, I'm talking about the pending queue. See #851 and give your opinion ;-)

hwoarangmy:
> By the way, icons from game-icons.net are licensed under CC BY 3.0, so we are supposed to give credits to the icon designers too

Note that from what I've seen, there are no credits to icons in ODIcons.png (and maybe in the others icon sets).

hwoarangmy:
@Bertram25 I've fixed requested stuff ^^

akien-mga:
Perfect, merging :-)

Bertram25:
cool :)

Issue #540, opened 2015-03-12
0.6.0 Release


Bertram25:
Here we go!! :D 

**General**
- [x] Allow to display callstack when a crash occurs with MSVS2013 versions #701 
- [x] Homogenize how the callstack is displayed independently from platform (to avoid using platform specific flags everywhere in the code) #701 
- [x] RectLayoutManager - is it used for anything? #690 
- [x] The training room max training level should be part of room config #698 
- [x] Move the game binary version to 0.6.0 #731 
- [x] Refactor a bit how modes are handled. #670 
- [x] Command-line options #688 
- [x] Dedicated server support. #665 
- [x] Add a cheat to unlock all skills at once #729 
- [x] Creature gold handling vs Payday #704 
- [x] Research manager to gather some glue code more properly. #843 
- [x] Build failure on MingW64 system due to bad CEGUI::getProperty() call. #889 

**Gameplay**
- [x] Add heal and explosion spells #706  
- [x] Add haste spell #842 
- [x] Improve spell logic - Done: #683 #727 #710 #741  
- [x] Reworked library logic #817
- [x] Add a warning when a player adds a trap but doesn't own a workshop (yet or anymore.) #671 
- [x] Tell about payday if at least one fighter and an active temple. #715 
- [x] Easy to sacrifice the creature when placing it #722 
- [x] Research queue handling improvement #851 

**Levels**
- [x] Remove levels redundancy #696 
- [x] Improve the Bridge level #673 
- [x] Drop the intermediate tile fullness values in the editor #590 
- [x] New Level! #847 

**Graphics**
- [x] Better-looking claimed walls models #808 #818
- [x] Add support for particle effects on entities. Done: #683 #709 #710 #711  
- [x] Add a position offset to creature position when there are more than one on the same tile. #697
- [x] Make goldbags have random orientation and offsets. #725
- [x] Scarab nose too thin #721 
- [x] Make most interesting models use normal maps. #511
- [x] Fix default tileset glitches Done: #273 #694 

**UI**
- [x] Key to hide keeper hand and tile selector #684 
- [x] Texts overlay handling improvement. #713 
- [x] Make a better distinction between names of game UI tabs and configuration menus #703 
- [x] Check whether there is a way to make the in-game buttons more responsive #343 
- [x] Skull flags - Improvement - Creature health and states. #733  #734
- [x] Add an in-game option + settings menu #3 
- [x] Research dependency + custom order handling #577 
[...]

PR #856, opened 2015-07-09
Add initial release notes for 0.6.0 + misc adjustments

akien-mga:
Note that the release notes are a bit more exhaustive than #540, as I went through the actual issues that were closed against the 0.6.0 milestone, and as usual, I found some gems that @hwoarangmy coded by surprise and that we forgot to list in the release issue ;-)

### DIFF
--- a/CREDITS
+++ b/CREDITS
@@ -144,7 +144,7 @@ textures/
 -> Grindstone*.png                     Daniel74, Nobiax        CC0                     http://forum.freegamedev.net/viewtopic.php?p=59775#p59775
 -> Hammock*.png                        Luke.RUSTLTD, Nobiax    CC0                     http://forum.freegamedev.net/viewtopic.php?p=59983#p59983
 -> Hasteflare.png                      OD Team                 CC-BY-SA 3.0            Used icon from game-icons.net
--> Heartflare.png                      Bertram                 CC-BY-SA 3.0            Own made flare + heart icon from game-icons.net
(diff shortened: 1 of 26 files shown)


akien-mga:
@Danimal696 Do we still use Claimedwall.material and ClaimedWall*.png, or have those been replaced by the DCW*.material and DungeonClaimedWall*.png files?

Danimal696:
only if we want to keep the legacy walls around, those wont be necesary if they are erased from everywhere

akien-mga:
> only if we want to keep the legacy walls around, those wont be necesary if they are erased from everywhere

But I think it's no longer the legacy walls:
https://github.com/OpenDungeons/OpenDungeons/blob/development/materials/textures/ClaimedWall.png

Bertram25:
@akien-mga 

> Fix misspellings found by codespell using the same procedure as described in #685

I'll be beaten for such arrogance I know ;), but could you add that small "procedure" to a new release management" page. I propose we add there what needs not to be forgotten when releasing, little by little.

akien-mga:
@Bertram25 I force-pushed a change to fix the commit log of e498c86 (and fix the two wrong # symbols used instead of *). Force-pushes are not a good practice, but it's cleaner with a correct log IMO :)

akien-mga:
> > Fix misspellings found by codespell using the same procedure as described in #685
> 
> I'll be beaten for such arrogance I know ;), but could you add that small "procedure" to a new release management" page. I propose we add there what needs not to be forgotten when releasing, little by little.

Good idea, I'll do that.

Bertram25:
All fine I merge.

Issue #840, opened 2015-07-07
Improve the Research button


Bertram25:
This issue is about adding visual cooldown support to research buttons, using a child vertical progress bar in it.

hwoarangmy:
cool. Do you think it is feasible for the 0.6.0 ? IMHO, it would be a nice addition (that will improve gameplay).

Bertram25:
Ok, I'll fix the last dicussed stuff for settings and I'll hop on that one. :)
Hopefully, it will be not too hard to implement. (Fingers crossed.)

Bertram25:
@hwoarangmy Still need to investigate on sending progress from server to client but with the above patch, you have progress bar in buttons which is looking even better that way.
I'll improve the progress texture used to be more readable afterwards.
Feel free to comment.

hwoarangmy:
As said in the commit, if the progressbar works, please create a PR and I will handle feeding the progressbars ^^

Bertram25:
Ok, I will. Just please keep the cog icon for the research being processed. :)
I add a function implementing what I need, even if I know it is valid only on server-side. Thanks a lot.

hwoarangmy:
> Just please keep the cog icon for the research being processed. :)

Ok, I will. However, research points are not attributed to a research so if you change the pending list, points will be on the new next research

Bertram25:
> However, research points are not attributed to a research so if you change the pending list, points will be on the new next research

Sure. But IMO, what should happen to the user, the cog icon will just move to the new active research button and the progress bar will then appear there.
Also, IMHO, the research progress bar should only appear if the progress is > 0 as you may have seen in the current logic. I'll also probably make it so it doesn't cover the cog icon afterwards by reducing it's width a bit.

For the record on the job done on that point, I first made a vertical research progressbar covering the whole button with a small alpha (0.4) and it felt quite like Torchlight. However it didn't look good or explicit to the user IMHO, so I change my initial thought on it, and added an horizontal progress bar on the button bottom part fully visible (not transparent), as it looked much better that way.
I'll make the progress image used configurable so that I can change it to something more visible for such a small progressbar.
Anyway, let's let @hwoarangmy do his magic first. ;)

hwoarangmy:
> Sure. But IMO, what should happen to the user, the cog icon will just move to the new active research button and the progress bar will then appear there.

Agreed.

For the record, it would be nice to use those buttons on the spells to display the cooldown

PR #857, opened 2015-07-10
Added support for a progress bar within the research buttons.

Bertram25:
- Also fixed Haste button location in the research tree.

For now, the progress isn"t sent by the server yet.

Part of #840 

Made for @hwoarangmy so that he can reuse it for the manager part.


hwoarangmy:
Works well. The progressbar is a bit small IMHO but it will be good until we have a proper button with a background that fills up according to progression

Bertram25:
> he progressbar is a bit small IMHO but it will be good until we have a proper button with a background that fills up according to progression

That was my first attempt, but I'll make a patch that remakes it that way so you can judge for yourself.

PR #861, opened 2015-07-10
Implemented a ResearchManager + use of progressbar for researches + small fixes

hwoarangmy:
Now, all data about research is handled in the ResearchManager (as well as the link between the research and the researched room/trap/spell).

The next step, if we want, would be to fill the rooms/traps/spells/researches directly from the manager by code (so that it is as simple as possible to add new data to the game).
Note that with that PR, I've added strong links on CEGUI widgets naming (research and room/trap/spell buttons should have the same name, the progressbar name should be buttonName + "ProgressBar", ...)


akien-mga:
The logic seems to still work just fine in this PR and the code looks like a nice simplification (I haven't proofread it thoroughly though).
Are the progress bars from @Bertram25 supposed to be implemented, or not yet? I did not notice them in the research tree.

hwoarangmy:
> Are the progress bars from @Bertram25 supposed to be implemented, or not yet?

No. It will be in a later commit.

hwoarangmy:
@akien-mga I'm currently trying to upload a new dev release (with my exact repo ATM ie until selling traps in the editor) on TuxFamily using a FTP client but it doesn't work with my user/passwd. Could you please check if I have enough rights to do so ? Or tell me how to do it if I have to use some special website ?

akien-mga:
@hwoarangmy I think you should have enough rights, just to be sure I've promoted the default ACL rights to "Can modify the ACL rights" (the highest level before "Can destroy this service"), but as far as I understand it only concerns the management of the download repository from the TuxFamily panel, and not upload with the FTP.

The wiki seems to imply that every contributor of the project can access the FTP: http://faq.tuxfamily.org/Downloads/En#Filling

Now for connecting, I'm having an issue right now with FileZilla on Linux, it doesn't seem to work properly (I'm having troubles with FileZilla lately on other FTPs, not sure it's related to TF). I tried with another client (gFTP), and it worked fine. With the command line "ftp" it works too on Linux:
`ftp ftp.tuxfamily.org`. It asks for my username, password, and gives me the FTP shell.

So if you're also using FileZilla, my best shot would be to try another client for the time being. I've just tried the "FireFTP" extension for Firefox, it works just fine for what we need (drag and drop upload of files).

hwoarangmy:
@akien-mga You are right, there seems to be a problem with filezilla. When I use command line ftp, it works fine. Strange. However, I'm uploading the new dev release so, if you want to make a linux version, please go on my repository.

akien-mga:
Done, thanks :)

akien-mga:
I'll try to investigate the FileZilla issue with TF guys, it might be that FileZilla has become more strict on some protocol stuff that TF does not fully comply with, and other FTP clients are more forgiving. Or FileZilla is just broken :-D

Bertram25:
Overall code looks good to me. Nice catch on capital letters. :)

I'll try that asap.

hwoarangmy:
> I'll try to investigate the FileZilla issue with TF guys

Note that I've tried 5-6 times to upload the snapshot yesterday with windows ftp command line but it failed every time (stopped before the end). Then, I gave a try again with filezilla and it worked well.
Note that first time I tried filezilla (when it didn't work), strangely, I could see the whole ftp content (not only opendungeons stuff). But it seems like I had no write permission (I tried on opendungeons path). When it worked, I could see opendungeons files only.

Bertram25:
I'm merging. :)

Issue #855, opened 2015-07-09
Portal wave AI configuration


akien-mga:
Thinking about #847, I got an idea on how to make the portal wave more interesting in various scenarios. IIUC, the current AI will locate the nearest player, and send waves towards him at regular intervals.

That means that to get a fair multiplayer map using portal waves, we'd need one portal wave per player. So I thought about making the portal wave choose for each wave a random target among the players within range. I don't know if the range of the portal wave is currently limited, but that could also be a nice addition to control its behaviour.

So we could have several behaviours:
- Attack the nearest player (current behaviour)
  - If a range is defined, we only attack the nearest player if he has claimed tiles in range. If not, we send no waves and just spawn some idle creatures.
  - If no range is defined (unlimited range), we attack the nearest player regardless of how far he is on the map.
- Attack any player in range
  - If a range is defined, we attack only the players in range; the target is chosen randomly among the players in range at each wave.
  - If no range is defined, we attack any player on the map, chosen randomly for each wave.

So in the config, we would need to new parameters:
- Attack behaviour: "Nearest" or "Random"
- Range: A number of tiles or unlimited (-1 maybe?)

WDYT?

Danimal696:
i love it, i want to see the undergroung become an ant farm crawling with coming and going ferals

Bertram25:
> If a range is defined, we only attack the nearest player if he has claimed tiles in range. If not, we send no waves and just spawn some idle creatures.

For the record, this is not part of the currently implemented behaviour, at least AFAIK.

Adding a range is a fun option and random target is cool, too! :)
I would even be for a third option: A fixed Team ID target. All allies in team X.

PR #866, opened 2015-07-11
New multiplayer map: Ruins of the Confluent

akien-mga:
A new map with an hydrographic map totally ripped off a real life city; my dungeon to whomever finds which one ;-)

It might be quite unbalanced for the time being due to #855: there is one portal wave in the middle, and I think it will constantly attack the player 1 at the moment even though all players are almost at the same distance of the portal (but player 1 might be a couple of tiles nearer).
Also the dwarf workers of the portal wave seat 5 dig their own walls when they have a target in mind, even though they just have to walk two tiles around them :-)

### DIFF
--- a/levels/multiplayer/RuinsOfTheConfluent.level
+++ b/levels/multiplayer/RuinsOfTheConfluent.level
@@ -10,7 +10,7 @@ FightMusic	TheDarkAmulet_MP.ogg
 [Seats]
 [Seat]
 seatId	1
-teamId	1
+teamId	1/2/3/4
 player	Choice
 faction	Keeper
 startingX	10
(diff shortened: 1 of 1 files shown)


Bertram25:
Need to try that :>

hwoarangmy:
I've tested the map and it works nice. We already know it is unbalanced so no need to test much until we have proper wave portal configuration

akien-mga:
> We already know it is unbalanced so no need to test much until we have proper wave portal configuration

Indeed. If we don't improve the portal wave AI for 0.6.0, I'll probably just comment out the portal wave part in the level file and replace it temporarily by a rogue portal with rogue creatures.

Issue #869, opened 2015-07-14
Notify player when the research queue is empty


akien-mga:
As discussed previously, a notification should be sent to the player when the research queue is empty, in a similar way as done for the lack of treasury tiles.

It sounds relatively easy to do, so I'll take it.

akien-mga:
For the reference, the fix should probably be similar to #435.

Bertram25:
@akien-mga Thanks for getting more in the dirt for the project. :)

akien-mga:
> Thanks for getting more in the dirt for the project. :)

Thanks for letting me use our project as a learning material, and mentoring me on the way ;)

Bertram25:
I guess this is done.! Congrats. :)

What's left is to check whether the player has got a library before complaining.
Should we also disabled the library panel until the player has got a library (in the menus and on the UI)?

hwoarangmy:
> Should we also disabled the library panel until the player has got a library

I would say no as it allows to choose researches at the beginning of the game, when there might be not much to do.

Bertram25:
> I would say no as it allows to choose researches at the beginning of the game, when there might be not much to do.

Good point.

akien-mga:
> What's left is to check whether the player has got a library before complaining.

The problem with this is that the current logic triggers the warning only when defining the next research; so if we don't trigger the warning when the player defines an empty research queue without a library, the warning won't get triggered after the library is built.
We could of course also add a check in the function called when creating the library, or find a better logic altogether.

Bertram25:
> The problem with this is that the current logic triggers the warning only when defining the next research; so if we don't trigger the warning when the player defines an empty research queue without a library, the warning won't get triggered after the library is built.
> We could of course also add a check in the function called when creating the library, or find a better logic altogether.

IMHO, as the warning is periodically triggered whatever the library state is, I guess adding a check to silence the warning whenever there is no library before triggering it may be the way to go.
(I mean maybe you can disable the visual warning in that case without disabling the whole periodical check?)

akien-mga:
> IMHO, as the warning is periodically triggered whatever the library state is, I guess adding a check to silence the warning whenever there is no library before triggering it may be the way to go.
> (I mean maybe you can disable the visual warning in that case without disabling the whole periodical check?)

That's a good point yes, I'll have a look.

hwoarangmy:
[...]

PR #870, opened 2015-07-15
Notify player when the research queue is empty

akien-mga:
Part of #869. What is missing is to add a check whether there are
actual skills left to be researched or not, and also make sure the
warning is displayed if there are no researches in queue from the start
(i.e. not only after a valid research is added).

### DIFF
--- a/source/game/PlayerFunctions.cpp
+++ b/source/game/PlayerFunctions.cpp
@@ -40,8 +40,9 @@ const float BATTLE_TIME_COUNT = 10.0f;
 //! Note that the distance is squared (100 would mean 10 tiles)
 const int MIN_DIST_EVENT_SQUARED = 100;
 
-//! \brief The number of seconds the local player will not be notified again if no treasury is available
-const float NO_TREASURY_TIME_COUNT = 30.0f;
+//! \brief The number of seconds the local player will not be notified again for looping notifications
+//! (no treasury tile available, no research in queue, etc.)
+const float NOTIFICATION_TIME_COUNT = 30.0f;
 
 void PlayerEvent::exportToPacket(GameMap* gameMap, ODPacket& os)
 {
@@ -250,13 +251,27 @@ void Player::notifyTeamFighting(Player* player, Tile* tile)
     fireEvents();
 }
 
+void Player::notifyNoResearchInQueue()
+{
+    if(mNoResearchInQueueTime == 0.0f)
+    {
+        mNoResearchInQueueTime = NOTIFICATION_TIME_COUNT;
+
+        std::string chatMsg = "Your research queue is empty, while there are still skills that could be unlocked.";
+        ServerNotification *serverNotification = new ServerNotification(
+            ServerNotificationType::chatServer, this);
+        serverNotification->mPacket << chatMsg << EventShortNoticeType::genericGameInfo;
+        ODServer::getSingleton().queueServerNotification(serverNotification);
+    }
+}
+
 void Player::notifyNoTreasuryAvailable()
 {
     if(mNoTreasuryAvailableTime == 0.0f)
     {
-        mNoTreasuryAvailableTime = NO_TREASURY_TIME_COUNT;
+        mNoTreasuryAvailableTime = NOTIFICATION_TIME_COUNT;
 
-        std::string chatMsg = "No treasury available. You should build a bigger one";
+        std::string chatMsg = "No treasury available. You should build a bigger one.";
         ServerNotification *serverNotification = new ServerNotification(
             ServerNotificationType::chatServer, this);
         serverNotification->mPacket << chatMsg << EventShortNoticeType::genericGameInfo;
@@ -411,6 +426,14 @@ void Player::upkeepPlayer(double timeSinceLastUpkeep)
         ServerNotification *serverNotification = new ServerNotification(
             ServerNotificationType::playerNoMoreFighting, this);
         ODServer::getSingleton().queueServerNotification(serverNotification);
+    }
+
+    if(mNoResearchInQueueTime > 0.0f)
+    {
+        if(mNoResearchInQueueTime > timeSinceLastUpkeep)
+            mNoResearchInQueueTime -= timeSinceLastUpkeep;
+        else
+            mNoResearchInQueueTime = 0.0f;
     }
 
     if(mNoTreasuryAvailableTime > 0.0f)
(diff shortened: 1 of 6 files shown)


akien-mga:
Actually rereading my changes, I see that mimicking the code for the treasury tile is not enough, as the treasury notification's "looping" come from being triggered repeatedly by workers, while the "no research in queue" message will only be triggered once after adding a completed research.

I'll see if I can make it appear again every 30s until a new research has been selected.

akien-mga:
> also make sure the warning is displayed if there are no researches in queue from the start
> (i.e. not only after a valid research is added).

Looks like the current code handles this case fine already thanks to the "null research entity" that gets added when no researches are selected.

akien-mga:
The warning should now loop properly. The last thing needed to fix #869 properly would be to check if there are still skills that could be unlocked before printing the warning (i.e. don't print a warning if a seat has researches 100% of the tree). I wouldn't be against a small hint from @hwoarangmy for this one (not that I couldn't find it myself, but I'm on holidays with my other half and have little time to dig into code :p).

hwoarangmy:
> i.e. don't print a warning if a seat has researches 100% of the tree

You have to check if all researches are done, not allowed or depend on a research not allowed. To do that, you have to create a new function in the ResearchManager that takes a Seat as parameter and checks Research::canBeResearched to know if a research is possible.

akien-mga:
> You have to check if all researches are done, not allowed or depend on a research not allowed. To do that, you have to create a new function in the ResearchManager that takes a Seat as parameter and checks Research::canBeResearched to know if a research is possible.

Great, I'll have a look at that. That was more or less what I had hinted while looking the ResearchManager and Seat files, but I just wanted to check if you had something bigger in mind with regard to a potential "autofill" button that should generate a sensible queue for the player.

akien-mga:
> To do that, you have to create a new function in the ResearchManager that takes a Seat as parameter and checks Research::canBeResearched to know if a research is possible.

Looking at the current code, it feels like such a function would fit more next to the other research-related functions of game/SeatFunctions.cpp, don't you think? Especially since it could then access mResearchDone directly, and this removes the need to pass the seat as a parameter.

It might make sense to move most of the research-related functions in SeatFunctions.cpp to the ResearchManager in a second time though.

akien-mga:
Actually I'll try to work it out in the ResearchManager class and just expose the needed mResearchDone variable in the Seat.

hwoarangmy:
I don't have any strong opinion about what should be where. What I would request is to let all everything related to the GUI links (button name, path in research, path in GUI, points for research, ...) in the same place so that if someone wants to create a new spell, he doesn't have to check everywhere in the code where he should add something. And since researches are shared by rooms/traps/spells, it made sense to put the code there.
But if you see a better suiting place, please feel free to change

Bertram25:
> related to the GUI links (button name, path in research, path in GUI, points for research, ...) in the same place so that if someone wants to create a new spell, he doesn't have to check everywhere in the code where he should add something. And since researches are shared by rooms/traps/spells, it made sense to put the code there.

+1 I guess that's the main architectural point here.

> But if you see a better suiting place, please feel free to change

In fact, if seomeone sees something, please discuss it first (as usual). [Yeah, I know, I'm an egomaniac dictator. ;)]

> Actually I'll try to work it out in the ResearchManager class and just expose the needed mResearchDone variable in the Seat.

All that said, putting the logic in the Research Manager and exposing the result in the Seat sounds pretty relevant. :)

source/game/PlayerFunctions.cpp
```diff
@@ -250,13 +251,27 @@ void Player::notifyTeamFighting(Player* player, Tile* tile)
     fireEvents();
 }
 
+void Player::notifyNoResearchInQueue()
+{
+    if(mNoResearchInQueueTime == 0.0f)
+    {
+        mNoResearchInQueueTime = NOTIFICATION_TIME_COUNT;
+
+        std::string chatMsg = "Your research queue is empty, while there are still skills that could be unlocked.";
```
Bertram25, line 263: This reminds me, I'll rename all that to skill one day :P

Bertram25:
Btw, it's cool to see you're opening the bowels of the beast. ;) @akien-mga 
So far, the code looks good, IMHO.

source/game/PlayerFunctions.cpp
akien-mga, line 263: Agreed :)

akien-mga:
I've done a first attempt at counting the number of remaining researches per seat, but I don't manage to call the function with the seat as argument, when in the seat source file.

Here is my diff: http://www.hastebin.com/xurokujire.coffee

And I get this error:

```
/home/akien/Projects/opendungeons/OpenDungeons-akien/source/game/SeatFunctions.cpp: In member function ‘void Seat::setNextResearch(ResearchType)’:
/home/akien/Projects/opendungeons/OpenDungeons-akien/source/game/SeatFunctions.cpp:1077:63: error: cannot call member function ‘std::vector<ResearchType> ResearchManager::getRemainingResearchesForSeat(const Seat*) const’ without object
         if(ResearchManager::getRemainingResearchesForSeat(this).size() == 0)
                                                               ^
```

Bertram25:
This:
`if(ResearchManager::getRemainingResearchesForSeat(this).size() == 0)`
Such call (`MyNameSpace::MyFunction()`) is to be done when your function is static. Or if you prefer not linked to an object.
(Yes, classes can be seen as namespaces for static functions)

You'll have to add the `static` word in your function declaration, just like it is done with `listAllResearches()`

Hope this will help. :)

akien-mga:
Thanks, that seems to do the trick :-)

akien-mga:
Done, now my pull request should be ready :)
Please note that I haven't tested my last commit extensively yet. It does seem to work though from a first test with unlocking all researches using the console command.

akien-mga:
I've changed the function to return a boolean instead (the name is a bit wonky, but I couldn't come up with something better). I've also added a small change to make sure that the warning does not loop when the players unlock their last research during the timer.

hwoarangmy:
Am I testing @akien-mga code ???

hwoarangmy:
First thing, the message triggers even if you don't own a library. That may be acceptable, though

hwoarangmy:
Works well. Mergeable for me

Bertram25:
> First thing, the message triggers even if you don't own a library. That may be acceptable, though

I suggest we open an issue for that. But let's merge one of the most c++ implied PR from @akien-mga first. ;)

PR #880, opened 2015-07-21
Fix unused VeteranScarab in ruins level

### LINKED ISSUE #878: Veteran Scarab is not used in the new map "Ruins of the confluent"
A level specific class "VeteranScarab" is defined but the scarabs on the map are simple Scarab, not VeteranScarab. Is it wanted ?

@akien-mga BTW, when you create a level, when it is done, please do not hesitate to open it in the editor and to save it (even without changing anything) to test if map saving is correct. Moreover, it would ensure that everything is sorted as the editor does which is better because if someone changes the map in the editor, he will see his changes only.

akien-mga:
Fixes #878


hwoarangmy:
Works well

PR #892, opened 2015-07-25
Add mouse-controlled camera rotation

tomluchowski:
This code attempts to do the autoscroll via mentioned function in >>ISSUES<<
It alows to rotate via RMB the camera about some Axis ...

### DIFF
--- a/source/camera/CameraManager.cpp
+++ b/source/camera/CameraManager.cpp
@@ -24,7 +24,6 @@
 #include "gamemap/TileContainer.h"
 #include "sound/SoundEffectsManager.h"
 #include "utils/LogManager.h"
-
 #include <OgreCamera.h>
 #include <OgreSceneNode.h>
 #include <OgreSceneManager.h>
@@ -76,10 +75,7 @@ CameraManager::CameraManager(Ogre::SceneManager* sceneManager, TileContainer* gm
 {
     createViewport(renderWindow);
     createCamera("RTS", 0.02, 300.0);
-    createCameraNode("RTS", Ogre::Vector3(static_cast<Ogre::Real>(10.0),
-                                              static_cast<Ogre::Real>(10.0),
-                                              MAX_CAMERA_Z / 2.0),
-                                              Ogre::Degree(0.0), Ogre::Degree(DEFAULT_X_AXIS_VIEW));
+    createCameraNode("RTS");
 
     setActiveCamera("RTS");
     setActiveCameraNode("RTS");
@@ -102,19 +98,16 @@ void CameraManager::createCamera(const Ogre::String& ss, double nearClip, double
     OD_LOG_INF("Creating " + ss + " camera...");
 }
 
-void CameraManager::createCameraNode(const Ogre::String& ss, Ogre::Vector3 xyz,
-                                     Ogre::Degree yaw, Ogre::Degree pitch, Ogre::Degree roll)
+void CameraManager::createCameraNode(const std::string& name)
 {
-    Ogre::SceneNode* node = mSceneManager->getRootSceneNode()->createChildSceneNode(ss + "_node", xyz);
-    Ogre::Camera* tmpCamera = getCamera(ss);
-    node->yaw(yaw, Ogre::Node::TS_LOCAL);
-    node->pitch(pitch, Ogre::Node::TS_LOCAL);
-    node->roll(roll, Ogre::Node::TS_LOCAL);
-    node->attachObject(tmpCamera);
-    node->setPosition(Ogre::Vector3(0,0,2) +  node->getPosition());
-    mRegisteredCameraNodeNames.insert(ss);
+    Ogre::SceneNode* node = mSceneManager->getRootSceneNode()->createChildSceneNode(name + "_node");
+    Ogre::Camera* tmpCamera = getCamera(name);
+    Ogre::SceneNode* node2 = node->createChildSceneNode(name + "_node2");
+    node2->attachObject(tmpCamera);
 
-    OD_LOG_INF("Creating " + ss + "_node camera node...");
+    mRegisteredCameraNodeNames.insert(name);
+
+    OD_LOG_INF("Creating " + name + "_node camera node...");
 }
 
 void CameraManager::setNextDefaultView()
@@ -173,7 +166,7 @@ void CameraManager::createViewport(Ogre::RenderWindow* renderWindow)
 
 Ogre::SceneNode* CameraManager::getActiveCameraNode()
 {
-    return getActiveCamera()->getParentSceneNode();
+    return mActiveCameraNode;
 }
 
 Ogre::SceneNode* CameraManager::setActiveCameraNode(const Ogre::String& ss)
@@ -255,15 +248,24 @@ void CameraManager::updateCameraFrameTime(const Ogre::Real frameTime)
     // Prevent the tilting to show a reversed world or looking too high.
     if (mRotateLocalVector.x != 0)
     {
-        Ogre::Real currentPitch = getActiveCameraNode()->getOrientation().getPitch().valueDegrees();
+        Ogre::Real currentPitch = getActiveCameraNode()->getChild(0)->getOrientation().getPitch().valueDegrees();
         if ((currentPitch >= 0.0 && mRotateLocalVector.x < 0)
             || (currentPitch <= 50.0 && mRotateLocalVector.x > 0))
         {
             // Tilt the camera up or down.
-            getActiveCameraNode()->rotate(Ogre::Vector3::UNIT_X,
-                                          Ogre::Degree(mRotateLocalVector.x * frameTime),
-                                          Ogre::Node::TS_LOCAL);
+            getActiveCameraNode()->getChild(0)->rotate(Ogre::Vector3::UNIT_X,
+                                        Ogre::Degree(mRotateLocalVector.x * frameTime),
+                                        Ogre::Node::TS_LOCAL);
         }
+
+    }
+    if (mRotateLocalVector.y != 0)
+    {
+        // Tilt the camera up or down.
+        getActiveCameraNode()->rotate(Ogre::Vector3::UNIT_Z,
+                                      Ogre::Degree(mRotateLocalVector.y * frameTime),
+                                      Ogre::Node::TS_WORLD);
+
     }
 
     // Swivel the camera to the left or right, while maintaining the same
@@ -311,9 +313,9 @@ void CameraManager::updateCameraFrameTime(const Ogre::Real frameTime)
     if (mCameraIsRotating && frameTime > 0.0)
     {
         Ogre::Quaternion orientationQuat = getActiveCameraNode()->getOrientation();
-
+        Ogre::Quaternion orientationQuat2 = getActiveCameraNode()->getChild(0)->getOrientation();
         // Tilting - X-Axis - Looking up or down
-        Ogre::Real pitch = orientationQuat.getPitch().valueDegrees();
+        Ogre::Real pitch = orientationQuat2.getPitch().valueDegrees();
         Ogre::Real pitchUpdate = ROTATION_SPEED.valueDegrees();
         Ogre::Real pitchDiff = std::abs(pitch - mCameraPitchDestination);
         Ogre::Real pitchChange = ((pitchUpdate * frameTime) > pitchDiff) ? pitchDiff / frameTime : pitchUpdate;
@@ -347,11 +349,14 @@ void CameraManager::updateCameraFrameTime(const Ogre::Real frameTime)
         if (pitchDone && rollDone)
             mCameraIsRotating = false;
 
-        /* Useful to debug... or test.
-        std::cout << "Current pitch: " << pitch << ", desired pitch: " << mCameraPitchDestination << std::endl;
-        std::cout << "Pitch update: " << pitchChange * frameTime << std::endl;
-        std::cout << "Current roll: " << roll << ", desired roll: " << mCameraRollDestination << std::endl;
-        std::cout << "Roll update: " << rollChange * frameTime << std::endl;
+        /*
+        // Useful to debug... or test.
+        std::stringstream ss;
+        ss << "Current pitch: " << pitch << ", desired pitch: " << mCameraPitchDestination << std::endl;
+        ss << "Pitch update: " << pitchChange * frameTime << std::endl;
+        ss << "Current roll: " << roll << ", desired roll: " << mCameraRollDestination << std::endl;
+        ss << "Roll update: " << rollChange * frameTime << std::endl;
+        OD_LOG_INF(ss.str());
         */
     }
 
@@ -417,9 +422,18 @@ Ogre::Vector3 CameraManager::getCameraViewTarget() const
     return target;
 }
 
-void CameraManager::setCameraPosition(const Ogre::Vector3& position)
+void CameraManager::resetCamera(const Ogre::Vector3& position)
 {
-    getActiveCameraNode()->setPosition(position);
+    Ogre::Quaternion quat(1, 0, 0, 0);
+    Ogre::Node* nodeRotation = getActiveCameraNode()->getChild(0);
+    nodeRotation->setOrientation(quat);
+
+    Ogre::Node* nodeCamera = getActiveCameraNode();
+    nodeCamera->setOrientation(quat);
+
+    nodeCamera->setPosition(position);
+
+    nodeRotation->rotate(Ogre::Vector3::UNIT_X, Ogre::Degree(DEFAULT_X_AXIS_VIEW), Ogre::Node::TS_LOCAL);
 }
 
 
@@ -444,7 +458,7 @@ void CameraManager::move(const Direction direction, double aux)
     switch (direction)
     {
     case moveRight:
-        if (currentPitch > 0.0)
+        if (currentPitch <= 0.0)
             mTranslateVectorAccel.x += MOVE_SPEED_ACCELERATION;
         else
             mTranslateVectorAccel.x -= MOVE_SPEED_ACCELERATION;
@@ -455,7 +469,7 @@ void CameraManager::move(const Direction direction, double aux)
         break;
 
     case moveLeft:
-        if (currentPitch > 0.0)
+        if (currentPitch <= 0.0)
             mTranslateVectorAccel.x -= MOVE_SPEED_ACCELERATION;
         else
             mTranslateVectorAccel.x += MOVE_SPEED_ACCELERATION;
@@ -466,7 +480,7 @@ void CameraManager::move(const Direction direction, double aux)
         break;
 
     case moveBackward:
-        if (currentPitch > 0.0)
+        if (currentPitch <= 0.0)
             mTranslateVectorAccel.y -= MOVE_SPEED_ACCELERATION;
         else
             mTranslateVectorAccel.y += MOVE_SPEED_ACCELERATION;
@@ -477,7 +491,7 @@ void CameraManager::move(const Direction direction, double aux)
         break;
 
     case moveForward:
-        if (currentPitch > 0.0)
+        if (currentPitch <= 0.0)
             mTranslateVectorAccel.y += MOVE_SPEED_ACCELERATION;
         else
             mTranslateVectorAccel.y -= MOVE_SPEED_ACCELERATION;
@@ -534,19 +548,19 @@ void CameraManager::move(const Direction direction, double aux)
         break;
 
     case randomRotateX:
-        mSwivelDegrees = Ogre::Degree(static_cast<Ogre::Real>(64 * aux));
+        mRotateLocalVector.y = static_cast<Ogre::Real>(8.0*aux);
         break;
 
     case zeroRandomRotateX:
-        mSwivelDegrees = Ogre::Degree(0.0);
+        mRotateLocalVector.y = 0.0;
         break;
 
     case randomRotateY:
-        mRotateLocalVector.x = static_cast<Ogre::Real>(64 * aux);
+        mRotateLocalVector.x = static_cast<Ogre::Real>(8.0*aux);
         break;
 
     case zeroRandomRotateY:
-        mRotateLocalVector.x =  0.0;
+        mRotateLocalVector.x = 0.0;
         break;
 
     default:
--- a/source/modes/GameMode.cpp
+++ b/source/modes/GameMode.cpp
@@ -59,6 +59,7 @@
 #include <algorithm>
 #include <vector>
 #include <string>
+#include "utils/LogManager.h"
 
 GameMode::GameMode(ModeManager *modeManager):
     GameEditorModeBase(modeManager, ModeManager::GAME, modeManager->getGui().getGuiSheet(Gui::guiSheet::inGameMenu)),
@@ -288,6 +289,13 @@ bool GameMode::mouseMoved(const OIS::MouseEvent &arg)
     InputManager& inputManager = mModeManager->getInputManager();
     inputManager.mCommandState = (inputManager.mLMouseDown ? InputCommandState::building : InputCommandState::infoOnly);
 
+    // TODO: Here we should check whether the terminal is active...
+    if(inputManager.mMMouseDown)
+    {
+        ODFrameListener::getSingleton().moveCamera(CameraManager::randomRotateX,arg.state.X.rel);
+        ODFrameListener::getSingleton().moveCamera(CameraManager::randomRotateY,arg.state.Y.rel);
+    }
+
     // If we have a room/trap/spell selected, show it
     // TODO: This should be changed, or combined with an icon or something later.
     TextRenderer& textRenderer = TextRenderer::getSingleton();
@@ -349,7 +357,6 @@ bool GameMode::mouseMoved(const OIS::MouseEvent &arg)
         inputManager.mLStartDragX = inputManager.mXPos;
         inputManager.mLStartDragY = inputManager.mYPos;
     }
-
     return true;
 }
 
@@ -399,13 +406,14 @@ bool GameMode::isMouseDownOnCEGUIWindow()
 
 bool GameMode::mousePressed(const OIS::MouseEvent& arg, OIS::MouseButtonID id)
 {
+    InputManager& inputManager = mModeManager->getInputManager();
+
     CEGUI::System::getSingleton().getDefaultGUIContext().injectMouseButtonDown(
         Gui::convertButton(id));
 
     if (!isConnected())
         return true;
 
-    InputManager& inputManager = mModeManager->getInputManager();
     inputManager.mMouseDownOnCEGUIWindow = isMouseDownOnCEGUIWindow();
     if (inputManager.mMouseDownOnCEGUIWindow)
         return true;
@@ -441,6 +449,11 @@ bool GameMode::mousePressed(const OIS::MouseEvent& arg, OIS::MouseButtonID id)
         return true;
 
     RenderManager::getSingleton().moveWorldCoords(inputManager.mKeeperHandPos.x, inputManager.mKeeperHandPos.y);
+
+    // The player should be able to move the mouse even if not clicking on a tile. Because of that, we set
+    // mMMouseDown before checking which tile is clicked
+    if (id == OIS::MB_Middle)
+        inputManager.mMMouseDown = true;
 
     int tileX = Helper::round(inputManager.mKeeperHandPos.x);
     int tileY = Helper::round(inputManager.mKeeperHandPos.y);
@@ -488,7 +501,6 @@ bool GameMode::mousePressed(const OIS::MouseEvent& arg, OIS::MouseButtonID id)
     if (id == OIS::MB_Right)
     {
         inputManager.mRMouseDown = true;
-
         // Stop creating rooms, traps, etc.
         inputManager.mLStartDragX = inputManager.mXPos;
         inputManager.mLStartDragY = inputManager.mYPos;
@@ -628,6 +640,16 @@ bool GameMode::mouseReleased(const OIS::MouseEvent &arg, OIS::MouseButtonID id)
 
     InputManager& inputManager = mModeManager->getInputManager();
     inputManager.mCommandState = InputCommandState::validated;
+
+    // First check for the axis rotation release, as this
+    // seems to be the most expected action the user wants to put at end
+
+    if(id == OIS::MB_Middle)
+    {
+        inputManager.mMMouseDown = false;
+        ODFrameListener::getSingleton().moveCamera(CameraManager::zeroRandomRotateX, 0.0);
+        ODFrameListener::getSingleton().moveCamera(CameraManager::zeroRandomRotateY, 0.0);
+    }
 
     // If the mouse press was on a CEGUI window ignore it
     if (inputManager.mMouseDownOnCEGUIWindow)
--- a/source/modes/InputManager.h
+++ b/source/modes/InputManager.h
@@ -53,7 +53,7 @@ public:
 
     //! \brief mouse handling related member
     OIS::Mouse*         mMouse;
-    bool                mLMouseDown, mRMouseDown;
+    bool                mLMouseDown, mRMouseDown, mMMouseDown;
     bool                mMouseDownOnCEGUIWindow;
 
     Ogre::Vector3       mKeeperHandPos;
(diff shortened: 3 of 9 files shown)


hwoarangmy:
I will try that tonight to see how it feels

source/modes/GameMode.cpp
```diff
@@ -350,6 +363,27 @@ bool GameMode::mouseMoved(const OIS::MouseEvent &arg)
         inputManager.mLStartDragX = inputManager.mXPos;
         inputManager.mLStartDragY = inputManager.mYPos;
     }
+    
+    if (!inputManager.mDirectionKeyPressed)
+    {
+      if (arg.state.X.abs < 30)
+	    ODFrameListener::getSingleton().moveCamera(CameraManager::moveLeft);;
```
hwoarangmy: Here, there are 2 ';' at the end

hwoarangmy:
I have seen that your editor uses tabs. Could you please configure it so it uses 4 spaces instead ?

hwoarangmy:
I've tested and seen some problems:
- When you move the camera rotation, if you move the mouse too fast, the move do not stop when you release the button
- With this PR, the keyboard cannot be used to move anymore. That's not acceptable IMHO since, as many other players, I guess, I use the keyboard to move much more than the mouse
- Using the mouse to navigate works to go down and right but not left and up (I would guess that it is because we don't reach mouse coordinate 0,0 for some reason)

After playing a bit with your PR, my impression is that the rotation is not very usefull as it is currently. If we had some blender like controls allowing to move everywhere around, I would say it is nice. But changing on 1 axe only is not very usefull IMHO.
Concerning navigation with the mouse, I think it is pretty interesting. To make it work, a solution might be to check if the cursor is on the 5% left, right, top or bottom rather than 0 or max. But that's a usefull feature IMHO.

Let's see what the others will say about it ^^

akien-mga:
Nice to see a PR from you @tomluchowski :) I'll test it asap.

akien-mga:
The PR does not compile for me with gcc 4.9:

```
[  1%] Building CXX object CMakeFiles/opendungeons.dir/source/modes/GameMode.cpp.o
/home/akien/Projects/opendungeons/OpenDungeons/source/modes/GameMode.cpp: In member function 'virtual bool GameMode::mouseMoved(const OIS::MouseEvent&)':
/home/akien/Projects/opendungeons/OpenDungeons/source/modes/GameMode.cpp:371:8: error: expected '}' before 'else'
        else
        ^
/home/akien/Projects/opendungeons/OpenDungeons/source/modes/GameMode.cpp: At global scope:
/home/akien/Projects/opendungeons/OpenDungeons/source/modes/GameMode.cpp:388:5: error: expected unqualified-id before 'return'
     return true;
     ^
/home/akien/Projects/opendungeons/OpenDungeons/source/modes/GameMode.cpp:389:1: error: expected declaration before '}' token
 }
 ^
/home/akien/Projects/opendungeons/OpenDungeons/source/modes/GameMode.cpp: In member function 'virtual bool GameMode::mouseMoved(const OIS::MouseEvent&)':
/home/akien/Projects/opendungeons/OpenDungeons/source/modes/GameMode.cpp:385:5: error: control reaches end of non-void function [-Werror=return-type]
     } 
     ^
```

tomluchowski:
Using the mouse to navigate works to go down and right but not left and up (I would guess that it is because we don't reach mouse coordinate 0,0 for some reason)
That's very strange as I don't get autoscrolling in here at all .....

tomluchowski:
How do you build your binary ? Do you allow on grabing mouse and keyboard ? What might be other options affecting that ?

akien-mga:
> How do you build your binary ? Do you allow on grabing mouse and keyboard ? What might be other options affecting that ?

Well I build with the default options using CMake, but the issue is pretty easy to find. You put two semi-colons at the end of line 370 (and the indentation is all broken, but that's not what's causing the issue).

```
      if (arg.state.X.abs < 30)
        ODFrameListener::getSingleton().moveCamera(CameraManager::moveLeft);;
        else
        ODFrameListener::getSingleton().moveCamera(CameraManager::stopLeft);
```

tomluchowski:
@hwoarangmy  : >>Using the mouse to navigate works to go down and right but not left and up (I would guess that it is because we don't reach mouse coordinate 0,0 for some reason)<<
So it somehow works, what settings do you have regarding mouse and keyboard grab ? 
  " But changing on 1 axe only is not very usefull IMHO. " well actually there are two axes ( try horizontal and vertical margins ) :)

akien-mga:
> Ok, now I removed the sliding bug , could you now merge into my code ?

Please click on the commits, there are many comments that @hwoarangmy and I made on the coding style and some dirty hacks that you need to take into account before we can merge it.

hwoarangmy:
> Ok, now I removed the sliding bug , could you now merge into my code ?

If you mean into your PR, note that now that as long as you have a PR on your current repository/branch, if you add commits to your github repository, they will automatically be added to the PR.
If you mean into the main repository, as @akien-mga said, you need to take into account our comments because for now, many of your code do not follow the code guidelines. Moreover, navigation do not work and I cannot use the keyboard anymore. If I am the only one to have this problem and you cannot reproduce, please tell me and I will have a deeper look.

tomluchowski:
I removed features that were not working, also I added patches to the code points you were mentioning ? Could you know merge in ?

Bertram25:
> I removed features that were not working, also I added patches to the code points you were mentioning ? Could you know merge in ?

Still a few fix requests left. ;)

hwoarangmy:
@tomluchowski Note that you have removed the feature I think was the most usefull: navigation.
However, I've tested playing with mouse rotation and it seems to work pretty well. It is a bit too sensitive to me, especially on the top of the screen. If you move only a little, you rotate from 360°. IMHO, half the rotation speed at least would be better.
Moreover, concerning rotation on the right, it might be good to have a min and max because I've reached a position where I was not looking at all at the map and I couldn't come back using the mouse only

tomluchowski:
@hwoarangmy 

> I've reached a position where I was not looking at all at the map and I couldn't come back using the mouse

The same here, camera seems to strangely stop under certain degree of view. I cannot find what blocks the camera. There is no code which would depend on current camera rotation.

```
        // Tilting - X-Axis - Looking up or down
        Ogre::Real pitch = orientationQuat.getPitch().valueDegrees();
        Ogre::Real pitchUpdate = ROTATION_SPEED.valueDegrees();
        Ogre::Real pitchDiff = std::abs(pitch - mCameraPitchDestination);
        Ogre::Real pitchChange = ((pitchUpdate * frameTime) > pitchDiff) ? pitchDiff / frameTime : pitchUpdate;
        bool pitchDone = false;
        if (pitchDiff < 0.5)
        {
            mRotateLocalVector.x = 0.0;
            pitchDone = true;
        }
        else if (pitch > mCameraPitchDestination)
            mRotateLocalVector.x = -pitchChange;
        else if (pitch < mCameraPitchDestination)
            mRotateLocalVector.x = pitchChange;

        // Roll - Z-Axis - Left or right
        Ogre::Real roll = orientationQuat.getRoll().valueDegrees();
        Ogre::Real rollUpdate = 1.3 * ROTATION_SPEED.valueDegrees();
        Ogre::Real rollDiff = std::abs(roll - mCameraRollDestination);
        Ogre::Real rollChange = ((rollUpdate * frameTime) > rollDiff) ? rollDiff / frameTime : rollUpdate;
        bool rollDone = false;
        if (std::abs(roll - mCameraRollDestination) < 0.5)
        {
            mSwivelDegrees = 0.0;
            rollDone = true;
        }
        else if (roll < mCameraRollDestination)
            mSwivelDegrees = rollChange;
        else if (roll > mCameraRollDestination)
            mSwivelDegrees = -rollChange;

        if (pitchDone && rollDone)
            mCameraIsRotating = false;
```

Could someone explain to me what that code does ?

akien-mga:
@tomluchowski In GitHub issues you can use " > " to indicate a quote, and encapsulate a code snippet with " ``` " (three back quotes) to get the code formatting (as you can see in your post that I edited).

tomluchowski:
Now I fixed the up down issue, could you finnally merge in ?

Bertram25:
@tomluchowski Hey, there is still commented code and stub functions to remove (as I pointed out in the corresponding commits.)
Also, I'm a bit lost now. In the end, what is feature(s) you are adding in this PR?

hwoarangmy:
I'm also a bit lost so I will comment again from file differences instead of commits.
@Bertram25 AFAIU, the PR is about rotating the camera when dragging RMB on the border of the screen

source/camera/CameraManager.cpp
```diff
@@ -253,18 +253,31 @@ void CameraManager::updateCameraFrameTime(const Ogre::Real frameTime)
...
-        if ((currentPitch >= 0.0 && mRotateLocalVector.x < 0)
-            || (currentPitch <= 50.0 && mRotateLocalVector.x > 0))
-        {
-            // Tilt the camera up or down.
-            getActiveCameraNode()->rotate(Ogre::Vector3::UNIT_X,
-                                          Ogre::Degree(mRotateLocalVector.x * frameTime),
-                                          Ogre::Node::TS_LOCAL);
-        }
-    }
+    // if (mRotateLocalVector.x != 0)
```
hwoarangmy: Please remove commented code

source/modes/GameMode.cpp
```diff
@@ -289,7 +290,20 @@ bool GameMode::mouseMoved(const OIS::MouseEvent &arg)
...
     inputManager.mCommandState = (inputManager.mLMouseDown ? InputCommandState::building : InputCommandState::infoOnly);
 
-    // If we have a room/trap/spell selected, show it
+    // What is the equivalent of the below ? How do I check whether terminal is active ? 
+    // if (ODFrameListener::getSingleton().isTerminalActive())
+    //     return true;
+
+
+    if(inputManager.mDragType == InputManager::rotateAxisX)
+      {
```
hwoarangmy: Wrong indentation

source/modes/GameMode.cpp
```diff
@@ -289,7 +290,20 @@ bool GameMode::mouseMoved(const OIS::MouseEvent &arg)
...
+	ODFrameListener::getSingleton().moveCamera(CameraManager::randomRotateX,arg.state.X.rel);
```
hwoarangmy: Use of tab (it should be 4 spaces)

source/modes/GameMode.cpp
```diff
@@ -289,7 +290,20 @@ bool GameMode::mouseMoved(const OIS::MouseEvent &arg)
...
+      }
+    else if(inputManager.mDragType == InputManager::rotateAxisY){
+	ODFrameListener::getSingleton().moveCamera(CameraManager::randomRotateY,arg.state.Y.rel);
+      }
+     
+    // If we have a room or trap (or later spell) selected, show what we have selected
```
hwoarangmy: You have removed the more accurate comment and left an older one

source/modes/GameMode.cpp
```diff
@@ -289,7 +290,20 @@ bool GameMode::mouseMoved(const OIS::MouseEvent &arg)
     InputManager& inputManager = mModeManager->getInputManager();
     inputManager.mCommandState = (inputManager.mLMouseDown ? InputCommandState::building : InputCommandState::infoOnly);
 
-    // If we have a room/trap/spell selected, show it
+    // What is the equivalent of the below ? How do I check whether terminal is active ?
```
hwoarangmy: You can test GameMode::mCurrentInputMode == InputModeConsole

source/modes/GameMode.cpp
```diff
@@ -400,13 +413,30 @@ bool GameMode::isMouseDownOnCEGUIWindow()
 
 bool GameMode::mousePressed(const OIS::MouseEvent& arg, OIS::MouseButtonID id)
 {
+    InputManager& inputManager = mModeManager->getInputManager();
```
hwoarangmy, line 409: Wrong indentation

source/modes/GameMode.cpp
```diff
@@ -400,13 +413,30 @@ bool GameMode::isMouseDownOnCEGUIWindow()
...
+    if (id == OIS::MB_Right)
+    {
+        inputManager.mRMouseDown = true;
```
hwoarangmy: Why are you doing that here and a few lines later ?

source/modes/GameMode.cpp
tomluchowski: Strange in my emacs editor it shows like a proper 4 spaces ....

source/modes/GameMode.cpp
```diff
@@ -489,7 +519,18 @@ bool GameMode::mousePressed(const OIS::MouseEvent& arg, OIS::MouseButtonID id)
...
         inputManager.mRMouseDown = true;
-
+        inputManager.mRStartDragX = inputManager.mXPos;
+        inputManager.mRStartDragY = inputManager.mYPos;
+        if(arg.state.Y.abs < 0.1*arg.state.height || arg.state.Y.abs > 0.9*arg.state.height)
+        {
+            inputManager.mDragType=InputManager::rotateAxisX;
+            return true;
+        }
+        else if(arg.state.X.abs > 0.9*arg.state.width || arg.state.X.abs < 0.1*arg.state.width)
```
hwoarangmy: Note that its pretty hard to get here because from the conditions you have set, you have to clic on the GUI to move the camera. And if you clic on the GUI, you won't get here

source/modes/InputManager.h
```diff
@@ -63,6 +71,7 @@ class InputManager
     //! \brief In editor mode, it contains the selected seat Id. In gamemode, it is not used
     int                 mSeatIdSelected;
     InputCommandState   mCommandState;
+    bool mDirectionKeyPressed;
```
hwoarangmy: You have defined and initialized this but it is not used

source/render/ODFrameListener.cpp
```diff
@@ -278,9 +278,26 @@ void ODFrameListener::setCameraPosition(const Ogre::Vector3& position)
     mCameraManager.setCameraPosition(position);
 }
 
-void ODFrameListener::moveCamera(CameraManager::Direction direction)
+void ODFrameListener::moveCamera(CameraManager::Direction direction, double aux)
 {
-    mCameraManager.move(direction);
+    mCameraManager.move(direction, aux);
+}
+// Code below stopped working after removing methods from CameraManager; I'm leaving
```
hwoarangmy: And why do you want this to work ? What is it for ?

tomluchowski:
Well it has to be shifted earlier, because the following code is returning from that function if it does get rayQuery empty. That's why you couldn't rotate when looking up , with nothing on screen  .

source/render/ODFrameListener.cpp
tomluchowski: aux - auxilery value for move method. For randomRotateX it is the distance of the mouse moved,
0 as default.

hwoarangmy:
@tomluchowski Would it be much harder to rotate view using the mouse middle button and moving the mouse like most applications do ? That would feel much more user friendly to do as google maps, blender and most others do rather than using the border of the screen. For my part, I would never have tried it on purpose. And if by mistake I would have rotated the camera this way, I think I would have guessed it was a bug.

Moreover, why did you remove the mouse navigation ? IMHO, it was more interesting than the camera rotation.

source/modes/GameMode.cpp
tomluchowski: "you have to clic on the GUI to move the camera. And if you clic on the GUI, you won't get here"
Ahh alright the bottom case ....

tomluchowski:
Well probably it was you who removed mouse navigation much earlier ;)
Now I try to add one feature at a time.

source/render/ODFrameListener.cpp
hwoarangmy: I was talking about the following functions that your comment says that need to be fixed but are not used

hwoarangmy:
> Well probably it was you who removed mouse navigation much earlier ;)

Maybe. But AFAIK, we only removed broken or not working functionalities ;-)

> Now I try to add one feature at a time.

Good idea.

source/camera/CameraManager.cpp
```diff
@@ -255,16 +255,29 @@ void CameraManager::updateCameraFrameTime(const Ogre::Real frameTime)
...
-            || (currentPitch <= 50.0 && mRotateLocalVector.x > 0))
-        {
-            // Tilt the camera up or down.
-            getActiveCameraNode()->rotate(Ogre::Vector3::UNIT_X,
-                                          Ogre::Degree(mRotateLocalVector.x * frameTime),
-                                          Ogre::Node::TS_LOCAL);
-        }
+        // NOTE : The camera loses the desired sense of left, right, top, down
+        // when the camera pitch is more than 90 degrees.
+            Ogre::Real pitch = getActiveCameraNode()->getOrientation().getPitch().valueRadians();
```
Bertram25: Wrong indent here as well.

Bertram25:
@tomluchowski Once the PR will be clean, I'll ask you to squash every commits into one, before a second review pass.
I'm asking this because I wouldn't like to be the guy later bisecting in this area. ;)
If you have problems with squashing, don't hesitate to ask for help.

Bertram25:
@tomluchowski

> Well probably it was you who removed mouse navigation much earlier ;)

Nope, It was I who removed the functionality. I did this because the feature wasn't finished and not reliable unfortunately. (I couldn't play with it). But we already had this discussion several times.

> Now I try to add one feature at a time.

Thanks. But don't forget features are better discussed before being implemented for it would be sad to reject some changes because you coded something not inline with current plans.

akien-mga:
> But don't forget features are better discussed before being implemented for it would be sad to reject some changes because you coded something not inline with current plans.

I agree, we would all have gained some time if we had discussed the two features before hand, instead of just getting old commits from half a codebase ago (the repo grows fast nowadays!) merged into the newer codebase.

I agree with hwo that middle clic + mouse rotation would be more user friendly than having to drag the edges of the screen to rotate the camera. I would maybe expect the edges feature on a smartphone, but not on a desktop :)

tomluchowski:
You keep setting high requierments on mine code, while you changed many things which does not look nice -- like Terminal and new floor .....The middle mouse is problematic, because user with one drag would both pitch and roll, loosing precision. Once I set proper roll, then I don't want  it to be broken while setting pitch ..... We could that an option in the future, but for now I wish you could merge in what I have done so far ....

akien-mga:
We keep high requirements on everybody's code. Just check the recent pull requests of @hwoarangmy or @Bertram25 to know the code quality standard that we follow. And currently, I'm sorry, but your code is quite messy in some parts, and you haven't replied to our comments about it, so I don't see this PR merged anytime soon.

akien-mga:
> you changed many things which does not look nice -- like Terminal and new floor

Care to be more specific? The terminal works way better than it used to, and the floor looks great too IMO.

tomluchowski:
@akien-mga  : So I should reply to every issue raised in here ? Like : I removed this wrong intendation etc. ???

akien-mga:
>  So I should reply to every issue raised in here ? Like : I removed this wrong intendation etc. ???

Yes, and fix it with a new commit as you just did (you can use one commit to fix all issues if you want).

source/camera/CameraManager.cpp
tomluchowski: Fixed that .

source/modes/GameMode.cpp
tomluchowski, line 409: What's wrong in here ?

source/modes/InputManager.h
tomluchowski: Fixed that.

source/modes/GameMode.cpp
akien-mga, line 409: I don't see anything wrong either on this line.

hwoarangmy, line 409: sorry, seems like I didn' see well

source/modes/InputManager.h
hwoarangmy: Well, the variable seems to still be there and still be unused

hwoarangmy:
> You keep setting high requierments on mine code

That's true for everybody. I remember a PR from @oyvindln that ended with more than 100 comments. Keeping a good quality is a requirement if we want to have a chance for coders to want to join and for the ones already on the project to not leave. Honestly, if the code was to become as it was, with random indentation, commented code everywhere and no coding rule becoming a mess to read and understand what it does, I wouldn't continue to work on it.

Bertram25:
> You keep setting high requierments on mine code, while you changed many things which does not look nice -- like Terminal and new floor .....

@tomluchowski Are we really having this discussion again? Really?
Yes, if you want some of your code merged in, it will have to be fixed for everything seen by the QA phase. And more than that, yes, it will have to fit our common vision of the project beforehand (we're always open to discussion.). You already opened issues in the past to ask questions, right?

Second thing very important, you keep adding '....' and sad emoji in each and every of your comments. I don't get it sincerely. If participating in this project makes you sad, or get you bored, then please don't. We're a bunch of coders just trying get a fun game out and having a fun experience learning and coding in the meanwhile. Yes, we are. :)

In conclusion, don't even think you can play the victim with me (or us) anymore. You know the deal perfectly well. I'm not going to let @akien-mga or @hwoarangmy spend more time on your flawed changes than necessary. And I especially don't want them to potentially lose their cool for you anymore.
I'm just sorry for you, sincerely, but sometimes you have to show the facts. Either you can fix the indentation flaws + the rest you seem unable to fix at once, merge every commits into something (finally) presentable, and then we can work on that, (You can ask for help on anything, and in that case, you'll do politely, without the usual bad temper, please.)
Ask yourself: Are you able to commit without adding broken indent, useless comments, or half-done code? (But of course, you are.)

Otherwise, then, it's simple, I won't want to lose time with your work anymore. If you want to have fun with opendungeons and make your own experiments, please keep them for your fork. I've got nothing against that after all.
For everyone to know, this is my last warning.

tomluchowski:
I will try on that, sorry for getting on your nerves.
Now I have my branch fully intended and tabs removed ( checked with git diff and grep)
"merge every commits into something (finally) presentable " 
Do you want me to squeze all the commits into one ? Afaik that would require history rewriting , and it would be troublesome to people who already cloned that repository . 
best regards paul424, 

The biggest problem I see is if we are in no - grab mode ( for K&M) the keyboard stops working after some time, is that an issue in master branch  ?

akien-mga:
Note: we've worked on IRC with @tomluchowski to squash all 23 commits of this PR in a single one (if you check the total diff here https://github.com/OpenDungeons/OpenDungeons/pull/892/files it seems quite ok to have this in one single commit). Now he just needs to clean his fork and apply the patch we made.

tomluchowski:
@OpenDungeons/developers Now this PR is ready to a new review after we have it in one single commit.

akien-mga:
I've made some comments on the code. Regarding the feature itself, it's way better when it's governed by the middle mouse button and not by right clicking on the edges of the screen IMO. It's more intuitive, and similar to what one can find in 3D modelling software.

As a further enhancement possibility, I'd like to see how it would behave if we set lower deltaX and deltaY limits to prevent having both rotation + tilting when you're just trying to get one of them (currently it rotates if there is a deltaX, and tilts if there is a deltaY, even if it's really small, like if you drag to an angle of 5° instead of a perfect 0°).

tomluchowski:
@akien-mga  already set false value to x11_mouse_grab . 
Also let's see how current setup works for people, and if they really need the "mouse beautifier " ...

akien-mga:
>  Also let's see how current setup works for people, and if they really need the "mouse beautifier " ...

Yeah, but that's just that such a change has to be discussed where it's appropriate, and not hidden in a commit that takes care of a totally different feature. But I see you've reverted it so it's fine.

akien-mga:
For me the PR is mergeable, but I'll let @hwoarangmy or @Bertram25 have a look first.

hwoarangmy:
The PR is much better like this. I've tested and the feature feels much more user friendly. However, I've seen a bug which seems pretty easy to reproduce. I've tried to play with camera rotation and got lost (couldn't see the map anymore). Then, I pressed 'V' to get back to default camera (which didn't seem to work BTW) and 'T' to go back to my dungeon temple (because I thought I might be out of the map). Then, the camera kept rotating.

Bertram25:
@tomluchowski Wow; now that's quite some nice change in quality. Congrats. :D 
Two commits really clean to read.

@hwoarangmy 's comments about the fact to use the `mActiveCameraNode` if possible should be done someday. Also, is it possible to use dedicated members for the middle button, instead of setting right button state when pushing the middle button?
Same for dragging members, and the fact that you're not checking whether the GUI has input precedence. But we're definitely getting there!

tomluchowski:
@hwoarangmy 
Yeap pressing V causes trouble, although I don't know how to fix that for now.
It's caused by CameraManager::RotateTo() method , it seems it should be rewritten for an simpler one.

tomluchowski:
@OpenDungeons/developers I already fixed the V-key bug.

source/modes/GameMode.cpp
```diff
@@ -487,8 +496,6 @@ bool GameMode::mousePressed(const OIS::MouseEvent& arg, OIS::MouseButtonID id)
     // Right mouse button down
     if (id == OIS::MB_Right)
     {
-        inputManager.mRMouseDown = true;
```
Bertram25, line 490: IMHO, this removed line should be readded. (We're almost there!)

Bertram25:
Still one small thing. The overall change is looking good, except for a few trivial things.
I'll test asap.

Bertram25:
It took me time to find time to fix my system and compile OD again but that's done!

The first thing I've noticed is that now. The view is moving when starting a map and it feels a bit weird to me. I'd rather have a view set up from the start and moving only when I ask it to.

The second thing is that the middle click movement is way too sensitive to me. (I guess this may become a setting?)

Apart from those two points, it's working fine. :)

@hwoarangmy @akien-mga WDYT of the feature?

akien-mga:
> The first thing I've noticed is that now. The view is moving when starting a map and it feels a bit weird to me. I'd rather have a view set up from the start and moving only when I ask it to.

Indeed, I had missed that. It's quite annoying, especially as it moves the camera in an angle that triggers a huge FPS drop.

> The second thing is that the middle click movement is way too sensitive to me. (I guess this may become a setting?)

I also find it too sensitive, but there could indeed be a sensitivity slider in the settings.

How do we want to go further? Merge it and work on the issues in development, or wait for @tomluchowski to fix some of the issues first in this PR?

I wouldn't cherry-pick the changes for 0.6.0 though, it's still too rough IMO.

Bertram25:
> How do we want to go further?

Good question. If the starting view movement can be fixed, it's mergeable to me (for 0.7).

akien-mga:
> If the starting view movement can be fixed, it's mergeable to me (for 0.7).

Sounds good to me too. @tomluchowski, do you see how you could fix it? (the fact that the camera tilts when the level is loaded, instead of staying with the default view)

source/modes/GameMode.cpp
```diff
@@ -59,6 +59,7 @@
 #include <algorithm>
 #include <vector>
 #include <string>
+#include "utils/LogManager.h"
```
hwoarangmy, line 62: OD includes should be before C/C++ standard ones (like all the other ones). Moreover, in this case, I don't think you use LogManager

source/modes/GameMode.cpp
```diff
@@ -450,6 +458,7 @@ bool GameMode::mousePressed(const OIS::MouseEvent& arg, OIS::MouseButtonID id)
 
     if (id == OIS::MB_Middle)
     {
+        inputManager.mDragType=InputManager::rotateAxisXY;
```
hwoarangmy: Since you used the middle clic for rotation, it should be a good thing to display the status window in the released event. Note that it is not blocking and if you don't feel like doing it, I can handle this part. Just tell me.

source/modes/InputManager.h
```diff
@@ -42,6 +42,13 @@ enum class InputCommandState
 class InputManager
 {
 public:
+    enum DragType
+    {
+        rotateAxisXY,
+        nullDragType
+    };
+
+    DragType mDragType;
```
hwoarangmy: mDragType is not initialized

hwoarangmy:
I confirm the camera issue at game start. I guess it's because of the initial camera position.
As everybody, I think the mouse move is too sensible but it should be configurable so it can be done later.
However, I think it is too easy to tilt the camera and get lost (especially because the mouse is too sensible). Could you please add a min and max to avoid having the map on the roof ?
When this small changes will be fixed, I would just ask to merge all commits into one (no need to do it now, only when everything is ready). The feature is becoming pretty cool ^^

source/modes/GameMode.cpp
tomluchowski: It's very nice that someone has offered me a help. I would gladly accept all the help for finishing this PR. Also if you could help with removing the starting camera tilt bug ...

Bertram25:
> It's very nice that someone has offered me a help. I would gladly accept all the help for finishing this PR. Also if you could help with removing the starting camera tilt bug ...

Sure, I guess it's a matter of spare time, though. Please let your code accessible. If someone is okay to take it and finish the feature, I'll ask him to create a personal branch, check-pick/pull the current commits in his own branch and do whatver is necessary on top of it before making a new PR.
The current commits could for instance be squashed into one, keeping @tomluchowski authoring.

hwoarangmy:
@tomluchowski I've fixed the camera tilt and the upper and lower rotation limits. You can find it here:
http://pastebin.com/Zv9F12jh

What is left:
- initialize mDragType
- remove the LogManager include (which was already included BTW)
- use mActiveCameraNode in getActiveCameraNode
- merge commits into 1

Once it is done, it will be mergeable for me. And once mouse sensitiveness will be configurable, it will be pretty cool to have a look in the dungeon. It might also be nice to allow to zoom a little bit more to allow to see better creatures ^^

tomluchowski:
@hwoarangmy Ok , many many thanks for that. 
"remove the LogManager include (which was already included BTW)"
well it;s undoable science there are calls to OD_LOG_INF

akien-mga:
@Bertram25 @hwoarangmy The commits were all merged into one again.

hwoarangmy:
> well it;s undoable science there are calls to OD_LOG_INF

I was talking about the second include (after the standard includes - LogManager was include twice).

hwoarangmy:
I've made some tries and it seems the fix for camera tilt is not enough. ATM, if you move the camera, go back to main menu and start a new one, the camera will move to the starting position as it used to. Moreover, I think the camera is too high per default (more than it used to be).
@tomluchowski I will have a look at that and post you hints ASAP

tomluchowski:
"ATM, if you move the camera, go back to main menu and start a new one, the camera will move to the starting position as it used to. Moreover, I think the camera is too high per default (more than it used to be)."
I hardly understand what you mean. I tried to restart new game several times from main menu , and the camera always had the same starting position.

hwoarangmy:
Have you changed the view before restrating ? My guess is that the camera keeps the initial position from the last game and switches (as it used to) to the default position.

tomluchowski:
keeps the initial position from the last game and switches (as it used to) to the default position.
You mean it transforms the first camera angle from previous game into first camera angle of current game , right ?

hwoarangmy:
No, I mean that the camera manager is created by the framelistener which is not deleted/created when you quit a game and start another one. Thus, I guess it keeps the last position from the previous game when you start a new one.

hwoarangmy:
@tomluchowski I've created a patch for your branch you can find it:
http://pastebin.com/PPpWpcmD

To apply it, create a file in your OD local path and use
git apply myPatch.patch

To summarize:
- I've removed camera initial position at camera creation (because position and rotation are set at game launch)
- I've reset the camera rotation at game launch
- I've created in the input manager a boolean for middle button instead of mDragType to have the same behaviour as the other buttons
- I've allowed to rotate camera even when clicking outside a tile

Note that I forgot to remove the second LogManager include. Please remove it.

If you are OK with this changes and see no bug, please create a commit with all this

akien-mga:
> To apply it, create a file in your OD local path and use
> git apply myPatch.patch

Even better would be:

```
git am myPatch.patch
```

It then commits the patch directly as if it had been committed by @hwoarangmy.

hwoarangmy:
For the record, the patch can be downloaded here:
http://dl.free.fr/mEnLUGNah

tomluchowski:
Oki done , who could apply my PR then ? 
ediT : hwoarangmy I already applied your patch.

hwoarangmy:
It works well. Excepted the unneeded include and the direction being inverted depending on pitch. I'm merging and I will fix the remaining in my PR

PR #896, opened 2015-07-29
MinGW64 compilation issue + display readable room/trap/spell name ingame + small fixes

hwoarangmy:


Bertram25:
Code looks good. +1000 for all what you did to permit to this release to exist. :D I'll test asap.

Bertram25:
Works like a charm!! :D I'll merge now.

I had a full game of Stonekeep siege and it was really fun and feeling already quite different in term of gameplay than 0.5.0! Everything looks much more polished now. :D Woohoo.

Noticed while playing while it isn't linked to current PR at all:
Sounds, nothing urgent but it's hearable now that part has improved:
- No sound when casting a spell (Be it Heal or Explosion, while sounds are there.) I didn't check the config, though.
- No sound when placing a room, or is that sound to shy?
  New: (For 0.7.0)
- No sound played when a creature places a trap.
- Every creature have sword attack sounds per default, maybe a few punch and claw ones could add some variety.

Balancing:
- @akien-mga is right, library stuff grimoire cost should be doubled, so that it's not that easy to get everything, and this will add more strategy.
- Why are spikes depending on the cannon trap while they are at same level in the research tree?
- Spell heal and explosion cooldown values should be tripled, IMHO.
  I can handle the balancing config if you guys agree with the proposed changes.

akien-mga:
>  Sounds, nothing urgent but it's hearable now that part has improved:

I guess we should open an issue to keep track of the sound improvements that we want to do? The cannon should also have a firing sound I guess, the spike trap a spike sound (à la Prince of Persia), and the boulder trap a rumbling sound, etc.

> I can handle the balancing config if you guys agree with the proposed changes.

Sounds good to me :)

hwoarangmy:
I've created an issue about sounds #897
I will try to fix it soon as it should not be big bugs (I have probably forgotten to trigger sounds when the event occurs).

> Why are spikes depending on the cannon trap while they are at same level in the research tree?

The whole tree has been built for test purpose. Feel free to change it as you want

> I can handle the balancing config if you guys agree with the proposed changes.

Thanks ^^

Bertram25:
Thanks for the issue, I'll create the same kind of one for balancing and will deal with all that tonight.

PR #899, opened 2015-07-30
Fire sound on forgotten events + fixed wrong research dependency check

hwoarangmy:


akien-mga:
Works fine. The sound effects have a too low volume compared to the music IMO, it makes it hard to hear the cannon firing for example, but I can confirm that it does fire :)

Bertram25:
Let's merge. :D

PR #902, opened 2015-07-30
Fix #898

### LINKED ISSUE #898: 0.6.0 Balancing tweaks
- [ ] @akien-mga is right, library stuff grimoire cost should be doubled, so that it's not that easy to get everything, and this will add more strategy.
- [ ] Why are spikes depending on the cannon trap while they are at same level in the research tree?
  Overall, researches shouldn't depend on same level research tree items but on the one level higher items.
- [ ] Spell heal and explosion cooldown values should be tripled, IMHO. Have a look at haste, too.

Bertram25:
Enjoy ;)

### DIFF
--- a/config/researches.cfg
+++ b/config/researches.cfg
@@ -15,15 +15,15 @@
 # spellCreatureExplosion            (int)    Points needed for researching spell CreatureExplosion
 
 [Researches]
-    roomCrypt	75
-    roomWorkshop	75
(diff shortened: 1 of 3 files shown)


hwoarangmy:
It works well. Now we need to test that ingame ^^

Bertram25:
> It works well. Now we need to test that ingame ^^

Yes!!

PR #911, opened 2015-08-02
Improve quick help and update release notes

akien-mga:
In the end there wasn't much to do for the quick help, as the version for 0.5.0 already mentioned the need for a library to research spells, etc.. We just made the concept evolve a lot and become way cooler, but the basic instructions remain similar :)

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -15,6 +15,7 @@ our repository at https://github.com/OpenDungeons/OpenDungeons
 * Particle effects!
 * Reworked library logic and made research order configurable
 * New creature overlays that show the creatures' mood
+* Customisable creature sound effects!
 * Doors to better block enemies and macromanage allies
 * In-game settings menu support!
 * New claimed walls graphics and various other graphical improvements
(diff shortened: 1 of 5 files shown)


akien-mga:
I've also changed the funny "Lan Solo" to something more boring, but that should be more international. After discussing it with @alketii, it seems that the use of the name "Lan" to refer to any kind of multiplayer game with friends is particularly French ;)

"LAN" is only "Local Area Network", so it doesn't accurate refer to "playing online with acquaintances". Actually "WAN" would already be more accurate ("Wide Area Network"), but also less understandable IMO. So I'm using "Multiplayer Levels" instead, though I liked the Star Wars reference better :p

akien-mga:
I've also reduced the inactive doors opacity, as I found it hard to see the difference between active and inactive doors during our LAN... er, WAN game I mean :p

hwoarangmy:
Works well. I will let @Bertram25 review and merge. Congrats for the text, it's nice ^^

Bertram25:
Let's merge :)

PR #920, opened 2015-08-06
Updated SFML cmake file + do not show mood overlay to enemy players

hwoarangmy:


Bertram25:
Testing...

Bertram25:
Works fine to me. :)

hwoarangmy:
Note that if you want to test easily, you can do a savegame where you have vision on an enemy unit (not rogue seat) and edit the level to set its hunger/awakness to 100/0. If everything goes well, you should not see the mood overlay ^^

akien-mga:
BTW when testing the CMake, make sure to start from a clean repo (without CMakeCache.txt), else it might say that it found SFML just because it's already referenced in the cache.

akien-mga:
I think @Bertram25 and I are both somewhat busy these days, but it doesn't make sense to let this PR idle for too long IMO, I don't see changes that would really threaten stability :)

Issue #415, opened 2015-01-03
OD_LINUX_NO_KEYBOARD_GRAB default value


oyvindln:
This is currently on by default, but having it on makes the keyboard not work at all in the game on my system. Unless the issue of having it off by default on other systems is larger than this, I think it should be off by default, as someone compiling the game for the first time on the wrong system without checking it out will be unable to play. I think that is more serious than the annoyance of having OD capture the keyboard on other systems.

akien-mga:
Since there are already two of us with this issue (@oyvindln and @tomluchowski), I guess we should indeed set the default to OFF.

I know that @tomluchowski is on openSUSE, but which distro are you using @oyvindln? It's a bit puzzling that it works great on Debian and Mageia but not on openSUSE. I'll compare Mageia's and openSUSE's OIS package to see if I can find something.

akien-mga:
There aren't many differences between openSUSE's and Mageia's OIS package it seems:
- https://svnweb.mageia.org/packages/cauldron/ois/current/SPECS/ois.spec?view=markup
- https://build.opensuse.org/package/view_file/home:openSUSE_user1/ois/ois.spec

oyvindln:
I am on Ubuntu 14.10

Bertram25:
@oyvindln This is a funny issue, isn't it?
I also wondered whether the desktop environement/login manager may be linked to that problem.

I'm using KDE/KDM on Debian. What are you using on Ubuntu?

akien-mga:
@tomluchowski said his input and focus issues were linked to his window manager, I think that's likely to be the issue.

I'm also using KDE/KDM and KWin as WM, with the default configuration for window focus (i.e. a window as focus until I click on another window).

oyvindln:
I am on XFCE, using standard click to focus.

Bertram25:
After thinking about this, I think we have no choice but grabbing both mouse and keyboard per default. @tomluchowski and @oyvindln are right about this after all.
Why? Because otherwise there is always a case where you can't either use the mouse or the keyboard with the game on a certain mix of OS + WM.

After some research about this, I also found it is possible to change dynamically this behaviour by deleting and reinitializing the input manager:
http://www.ogre3d.org/forums/viewtopic.php?f=2&t=62051
http://www.ogre3d.org/tikiwiki/Using+OIS
[...]

PR #934, opened 2015-09-01
Add support for changing input grab in the settings window

akien-mga:
Currently settings are only applied after a restart. To prevent the need for
a restart, the InputManager should be destroyed and recreated with the new
parameters.

The feature is implemented for both X11 and Win32, though untested on Win32.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -59,20 +59,6 @@ if(OD_BUILD_TESTING)
     include(CTest)
 endif()
 
-if (UNIX AND NOT APPLE)
-    # Linux option - Do not grab the keyboard when using OIS
-    # This is breaking the game's input on certain linux distributions and thus, must stay an option for now...
-    option(OD_LINUX_NO_KEYBOARD_GRAB "Do not grab keyboard input when using OIS under Linux. Permits input in other applications." ON)
-
-    # Defines the value for the preprocessor.
-    if(OD_LINUX_NO_KEYBOARD_GRAB)
-        message(STATUS "Linux (OIS): x11_keyboard_grab = false")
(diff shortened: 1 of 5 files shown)


akien-mga:
So here is the feature discussed in #415. It's only a first implementation, as currently changing the input grabbing requires to restart OD so that the changes are taken into account. But this is really a workaround for the annoying bug some users are having on Linux, so I think it's worth backporting into 0.6 even though it's not very well polished.

akien-mga:
@Bertram25 @hwoarangmy The feature works fine on X11, but I never tested it on Windows. Please check if it works OK (or at least does not do something completely unrelated to input grabbing). If it does not, I can revert the Win32 part and simply add "(Linux)" next to the options in the GUI.

Bertram25:
Code looks fine as for me. I'll test now. @hwoarangmy Could you handle the windows test part?

Bertram25:
Works fine. Mergeable as for me. :)

hwoarangmy:
I will try that tonight

akien-mga:
Once this one is merged, I'll cherry-pick all the relevant commits into the 0.6 branch (including the fix for ScrollPanel issues) and prepare RC3 binaries.

hwoarangmy:
I confirm it works
Note that on windows, when enabled, mouse grabbing do not the mouse to leave the window (I always play in windowed mode). So, it works as expected.

Concerning the keyboard, I don't really see a difference but it always worked for me. At least, I've seen no regression.

I merge

hwoarangmy:
@akien-mga It is merged. Once you have updated the 0.6 branch, I will also go for the 0.6.0 RC3 windows binaries

akien-mga:
@hwoarangmy I've cherry-picked our last two commits in the 0.6 branch, so it should be ready.

PR #936, opened 2015-09-03
Implemented KO state + prison

hwoarangmy:


akien-mga:
Nice! :-D

It builds fine. When cheating, I already found a first issue: "unlockresearches" mentions an unknown enum value in place of the Prison.

akien-mga:
Here be skulls! :P

hwoarangmy:
> "unlockresearches" mentions an unknown enum value in place of the Prison.

I do not reproduce. Could you please post the log ?

akien-mga:
> > "unlockresearches" mentions an unknown enum value in place of the Prison.
> 
> I do not reproduce. COuld you please post the log ?

I don't see anything related in the logs, but here's how it looks like in the info box when you scroll up:
missing

There are actually two issues shown in this image:
- "No Research type is now available." (I guess the unlockresearches function is a bit overly enthusiastic :p)
- "Unknown enum value:8 is now available." (I guess it's the Prison)

hwoarangmy:
ok. Seems like I missed one enum to string convertion ^^
That's why I like when they are all at the same place. I will have a look ASAP

hwoarangmy:
@akien-mga I've fixed the research text

Bertram25:
> I'll check whether they are triggering a warning. at compile time.

No warnings. All fine. :)
Btw, latest ogre dev package don't trigger warnings anymore with gcc5 :D 

I tested it, it's a great start and I really like what's coming with the player settings window :D 
There are a few things to improve as listed by @Danimal696 of course, and I would definitely merge the player settings and the objective window but it's mergeable as for me.

hwoarangmy:
> I would definitely merge the player settings and the objective window

If there is enough space, I've nothing against

hwoarangmy:
I've rebased and fixed requested changes + camera small fixes

Bertram25:
The latest commit creates no regression AFAICS. Great to stuff merged. :)

hwoarangmy:
@OpenDungeons/developers I've added a commit for using the registered name for rooms/traps when generating the room/trap unique name. That will allow to have empty meshname for rooms/traps that do not use a mesh over the ground

Bertram25:
It's working fine as far as I tested. We should merge, @hwoarangmy already got some bridges to add! ;)

PR #943, opened 2015-09-12
Implemented bridges

hwoarangmy:
multiplayer games will be faster with that I guess ^^


Bertram25:
Code looks good AFAICS. Needs a few games to make sure everything is fine.
Btw, the prison icon shadow needs to be redone (I can do it.) :)

akien-mga:
Very nice, I'll test this asap :)

akien-mga:
I'm not convinced by this GUI change. Shouldn't we put everything on the same line, maybe with some empty space before the two bridge icons?
gui

Or is it required to fit on very small resolutions?

akien-mga:
Awesome work on the mouse drag logic to build bridges, it works really fine.

About the colouring, is it expected that the whole tile gets coloured, or is it a driver-specific issue?
colour

Also, the stone bridge does not seem coloured, but I guess it's because there is no reference to its mask in the material definition. @Danimal696, could you have a look at it using the snapshot @hwoarangmy recently uploaded? I guess StoneBridge.material should be more similar to WoodBridge.material.

hwoarangmy:
> Shouldn't we put everything on the same line

It didn't fit in editor mode (with portals + dungeon temple). Even with prison icon only. So, adding 2 more icons needed to change that somehow. However, if you see some better GUI, do not hesitate.

> is it expected that the whole tile gets coloured, or is it a driver-specific issue

I have the same. I guess it comes from the model. I've already said there was this problem in the forum and @Danimal696 is having a look.

akien-mga:
I think we can already merge what's here. I'll open issues for the small things noted here:
- WoodBridges gets completely coloured, and StoneBridge does not get coloured at all
- Rooms UI panel is starting to be crowded; how to improve it?
- There was also a question about move even more code from RoomBridge{Wooden,Stone} to RoomBridge, but I'll wait for input from @Bertram25 before opening a ticket.

hwoarangmy:
@akien-mga IMHO, it is better to open an issue and close it if we decide to do nothing. For my part, I usually do not check new messages in closed PR or issues...

Issue #944, opened 2015-09-14
Wooden bridges are completely coloured and stone bridges have no colour


akien-mga:
As shown here:
![colour](https://cloud.githubusercontent.com/assets/4701338/9836209/f14ff4c8-5a10-11e5-961f-bd7dcb832f01.png)

Also, the stone bridge does not seem coloured, but I guess it's because there is no reference to its mask in the material definition. @Danimal696, could you have a look at it using the snapshot @hwoarangmy recently uploaded? I guess StoneBridge.material should be more similar to WoodBridge.material.

Danimal696:
![bridges](https://cloud.githubusercontent.com/assets/8299756/9848664/7df92b1a-5ae3-11e5-9468-bd620743a6be.jpg)

akien-mga:
Fixed by efc2b01.

PR #946, opened 2015-09-14
Fixed bridges materials

Danimal696:
Now they properly show the star colour and dont leak colour to texture


akien-mga:
Nice, I'll test it tonight. It should fix #944, so I've assigned it to you.

hwoarangmy:
works well. Can be merged IMO. I will let @akien-mga test, though

akien-mga:
Works well indeed.

PR #949, opened 2015-09-14
Improvements on prison + bridges

hwoarangmy:
As I said, I've not used the icon for stunned creatures because I couldn't get something good. @Danimal696 if you can get something nice on 32x32, please don't hesitate to change


Danimal696:
> Creatures in prison should display an icon to any player with vision … 

I donrt understand this one, what do you mean by one icon?

hwoarangmy:
> I donrt understand this one, what do you mean by one icon?

I mean that if a creature is in prison, every player (allied or enemy) should see a mood icon to know that it is in prison. Otherwise, if one of your creatures is in my prison during a fight, you would not understand why it is not attacking (and not attacked) or you could try to pick it up but it wouldn't work.

akien-mga:
There seems to be a merge conflict, probably due to #948.

_Edit:_ Or actually no, it's because of the new icons from Danimal that I merged earlier in #950.

akien-mga:
Code-wise the PR is fine to me. I'll try to test the features this evening.

hwoarangmy:
ok. I will rebase ASAP

hwoarangmy:
@OpenDungeons/developers I've rebased ^^

Bertram25:
Will test asap. :)

Bertram25:
Works damn well! There are details that can be sorted out afterwards:
- the lizardman bed is above the lizardman when it sleeps.
- The sleepy icon are missing?
- The Spiders still look misplaced when sleeping but it could be config only.
- It's pretty hard to make a worker take a research book when dropping it on one. Overall, they are handled but late compared to other activities, IMHO. Maybe this should be handled before anything else when the kobolds sees it? WDYT?

Great job @hwoarangmy :D

Danimal696:
Yes, lots of candy in here :)

akien-mga:
I could test the features only briefly, but code-wise I'm fine with merging. @Bertram25 Feel free to do it and to create one or two bug reports for the issues you noticed.

hwoarangmy:
> the lizardman bed is above the lizardman when it sleeps.

Yes. I don't know if we should change the animation or somehow define a Z value fo sleeping creatures

> The sleepy icon are missing?

I've tested and the sleepy icon works on my side. Note that it is displayed when the creature is sleepy (<20 awakness), not when it is sleeping

> The Spiders still look misplaced when sleeping but it could be config only.

Same as for lizardman

> It's pretty hard to make a worker take a research book when dropping it on one

Yes, that's not new. I guess carrying entities should be a higher priority too (AFAIK, it even comes after claiming walls when you drop a kobold ^^)

akien-mga:
About the lizardman and spider sleep (=death) animation, the true solution would be to actually create a sleep animation instead of making them die on their bed :p

akien-mga:
Let's merge, I'll open an issue for the carrying entities priority.

Danimal696:
> About the lizardman and spider sleep (=death) animation, the true solution would be to actually create a sleep animation instead of making them die on their bed :p

Quite true :D

PR #954, opened 2015-09-18
[Missing fixes] Tile Culling mechanism

tomluchowski:
The game gets tile culling mechanism with this upgrade. Please open console and type 'startTileCulling'

### DIFF
--- a/source/camera/CameraManager.h
+++ b/source/camera/CameraManager.h
@@ -24,8 +24,7 @@
 #define CAMERAMANAGER_H_
 
 #include "camera/HermiteCatmullSpline.h"
-#include "gamemap/GameMap.h"
-#include "entities/Tile.h"
+
 
 #include <OgrePrerequisites.h>
 #include <OgreVector3.h>
@@ -36,6 +35,9 @@
 
 class TileContainer;
 class CullingManager;
+class GameMap;
+class TileContainer;
+
 
 // The min/max camera height in tile size
 const Ogre::Real MIN_CAMERA_Z = 3.0;
--- a/source/camera/CameraManager.h
+++ b/source/camera/CameraManager.h
@@ -25,7 +25,6 @@
 
 #include "camera/HermiteCatmullSpline.h"
 
-
 #include <OgrePrerequisites.h>
 #include <OgreVector3.h>
 #include <OgreString.h>
@@ -37,7 +36,6 @@ class TileContainer;
 class CullingManager;
 class GameMap;
 class TileContainer;
-
 
 // The min/max camera height in tile size
 const Ogre::Real MIN_CAMERA_Z = 3.0;
--- a/source/camera/CameraManager.h
+++ b/source/camera/CameraManager.h
@@ -24,6 +24,8 @@
 #define CAMERAMANAGER_H_
 
 #include "camera/HermiteCatmullSpline.h"
+#include "gamemap/GameMap.h"
+#include "entities/Tile.h"
 
 #include <OgrePrerequisites.h>
 #include <OgreVector3.h>
@@ -33,6 +35,7 @@
 #include <set>
 
 class TileContainer;
+class CullingManager;
 
 // The min/max camera height in tile size
 const Ogre::Real MIN_CAMERA_Z = 3.0;
@@ -48,6 +51,10 @@ enum class ViewModes : uint16_t
 
 class CameraManager
 {
+    friend class Console;
+    friend class CullingManager;
+    friend class ODFrameListener;
+    
 public:
     enum Direction
     {
@@ -62,7 +69,7 @@ public:
         fullStop
     };
 
-    CameraManager(Ogre::SceneManager* sceneManager, TileContainer* gameMap, Ogre::RenderWindow* renderWindow);
+    CameraManager(Ogre::SceneManager* sceneManager, GameMap* gameMap, Ogre::RenderWindow* renderWindow);
     ~CameraManager()
     {}
 
@@ -86,10 +93,10 @@ public:
         return mTranslateVectorAccel;
     }
 
-    bool onFrameStarted()
-    { return true; }
-    bool onFrameEnded()
-    { return true; }
+    bool onFrameStarted();
+    bool onFrameEnded();
+
+    CullingManager* mCullingManager;
 
     /*! \brief Sets the camera to a new location while still satisfying the
     * constraints placed on its movement
@@ -185,7 +192,8 @@ private:
     Ogre::Camera* mActiveCamera;
     Ogre::SceneNode* mActiveCameraNode;
 
-    TileContainer* mGameMap;
+
+    GameMap* mGameMap;
 
     //! \brief Is true when a camera is flying to a given position.
     bool            mCameraIsFlying;
--- a/source/camera/CameraManager.h
+++ b/source/camera/CameraManager.h
@@ -51,10 +51,7 @@ enum class ViewModes : uint16_t
 
 class CameraManager
 {
-    friend class Console;
     friend class CullingManager;
-    friend class ODFrameListener;
-    
 public:
     enum Direction
     {
@@ -102,6 +99,9 @@ public:
     * constraints placed on its movement
     */
     void updateCameraFrameTime(const Ogre::Real frameTime);
+
+    void startCreatureCulling();
+    void startTileCulling();
 
     /*! \brief Computes a vector whose z-component is 0 and whose x-y coordinates
     * are the position on the floor that the camera is pointed at.
--- a/source/camera/CullingManager.cpp
+++ b/source/camera/CullingManager.cpp
@@ -28,9 +28,6 @@
 #include <sstream>
 #include <algorithm>
 
-using  std::set; using std::swap; using std::max; using std::min;
-using  std::cerr; using std::endl;
-
 //! Values used to know whether to show and/or hide a mesh
 static const int HIDE =  1;
 static const int SHOW =  2;
@@ -179,7 +176,7 @@ void CullingManager::newBashAndSplashTiles(int64_t mode){
    
         if(mDebug)
         {
-            ss << endl;
+            ss << std::endl;
             ss << "DxLeft, DxRight: " << double(DxLeft)/Unit << " " << double(DxRight)/Unit << " " ;
             LogManager::getSingleton().logMessage( ss.str(),LogMessageLevel::NORMAL);
             ss.str("");
@@ -189,16 +186,16 @@ void CullingManager::newBashAndSplashTiles(int64_t mode){
         }
     
         int64_t mm = ((std::min(xxLeft, xxLeftOld) >> mPrecisionDigits) << mPrecisionDigits) ;
-        if(std::min(xxLeft, xxLeftOld) < max(xxRight,xxRightOld) )
+        if(std::min(xxLeft, xxLeftOld) < std::max(xxRight,xxRightOld) )
         {
-            for (int64_t xx = mm ; xx <= max(xxRight,xxRightOld) ; xx+= Unit)
+            for (int64_t xx = mm ; xx <= std::max(xxRight,xxRightOld) ; xx+= Unit)
             {
                 bool bash = (xx >= xxLeftOld && xx <= xxRightOld && (yy >= oldWalk.getBottomLeftVertex().y) && yy <= oldWalk.getTopLeftVertex().y);
                 bool splash = (xx >= xxLeft && xx <= xxRight && (yy >= mWalk.getBottomLeftVertex().y) && yy <= mWalk.getTopLeftVertex().y);
 
                 if (mDebug)
                 {
-                    ss<< " x " <<  (xx >> mPrecisionDigits )<< " y " << (yy >> mPrecisionDigits ) << " " << (splash && (mode & SHOW)) << (bash && (mode & HIDE)) << endl;
+                    ss<< " x " <<  (xx >> mPrecisionDigits )<< " y " << (yy >> mPrecisionDigits ) << " " << (splash && (mode & SHOW)) << (bash && (mode & HIDE)) << std::endl;
             
                 }
                 xxp = ((xx >>mPrecisionDigits) - 128.0);
@@ -210,9 +207,13 @@ void CullingManager::newBashAndSplashTiles(int64_t mode){
                 }
 
                 else if (gm && bash && (mode & HIDE) && xxp>=0 && yyp>= 0 && xxp<mCm->mGameMap->getMapSizeX() && yyp < mCm->mGameMap->getMapSizeY())
+                {
                     gm->getTile(xxp, yyp)->hide();
+                }
                 else if (gm && splash && (mode & SHOW) && xxp>=0 && yyp>= 0 && xxp<mCm->mGameMap->getMapSizeX() && yyp < mCm->mGameMap->getMapSizeY())
+                {
                     gm->getTile(xxp, yyp)->show();
+                }
             }
         }
     }
@@ -236,8 +237,8 @@ bool CullingManager::getIntersectionPoints()
             mOgreVectorsArray[ii]= (mMyRay[ii].getPoint(intersectionResult.second));
         else
         {
-            cerr<< "I didn't find the intersection point for " << ii <<"th ray"<<endl;
-            exit(1);
+            std::cerr<< "I didn't find the intersection point for " << ii <<"th ray"<<std::endl;
+            // exit(1);
         }
     }
     return true;
--- a/source/camera/CullingManager.cpp
+++ b/source/camera/CullingManager.cpp
@@ -144,8 +144,6 @@ void CullingManager::hideAllTiles(void)
     }
 }
 
-
-
 void CullingManager::showAllTiles(void)
 {
     GameMap* gm = mCm->mGameMap;
@@ -160,10 +158,6 @@ void CullingManager::showAllTiles(void)
         }
     }
 }
-
-
-
-
 
 void CullingManager::newBashAndSplashTiles(int64_t mode){
     int64_t xxLeftOld = oldWalk.getTopLeftVertex().x;
@@ -199,6 +193,7 @@ void CullingManager::newBashAndSplashTiles(int64_t mode){
     
         int64_t mm = ((std::min(xxLeft, xxLeftOld) >> mPrecisionDigits) << mPrecisionDigits) ;
         if(std::min(xxLeft, xxLeftOld) < max(xxRight,xxRightOld) )
+        {
             for (int64_t xx = mm ; xx <= max(xxRight,xxRightOld) ; xx+= Unit)
             {
                 bool bash = (xx >= xxLeftOld && xx <= xxRightOld && (yy >= oldWalk.getBottomLeftVertex().y) && yy <= oldWalk.getTopLeftVertex().y);
@@ -222,7 +217,7 @@ void CullingManager::newBashAndSplashTiles(int64_t mode){
                 else if (gm && splash && (mode & SHOW) && xxp>=0 && yyp>= 0 && xxp<mCm->mGameMap->getMapSizeX() && yyp < mCm->mGameMap->getMapSizeY())
                     gm->getTile(xxp, yyp)->show();
             }
-
+        }
     }
 }
 
@@ -257,7 +252,6 @@ bool CullingManager::onFrameStarted()
         getIntersectionPoints();
     if(mCullTilesFlag)
         cullTiles();
-
     return true;
 }
 
--- a/source/camera/CullingManager.cpp
+++ b/source/camera/CullingManager.cpp
@@ -53,10 +53,10 @@ CullingManager::CullingManager(CameraManager* cameraManager):
 void CullingManager::cullTiles()
 {
     mOldWalk = mWalk;
-    mWalk.myArray[0] =  Vector3i(mOgreVectorsArray[0]);
-    mWalk.myArray[1] =  Vector3i(mOgreVectorsArray[1]);
-    mWalk.myArray[2] =  Vector3i(mOgreVectorsArray[2]);
-    mWalk.myArray[3] =  Vector3i(mOgreVectorsArray[3]);
+    mWalk.mVertices[0] =  Vector3i(mOgreVectorsArray[0]);
+    mWalk.mVertices[1] =  Vector3i(mOgreVectorsArray[1]);
+    mWalk.mVertices[2] =  Vector3i(mOgreVectorsArray[2]);
+    mWalk.mVertices[3] =  Vector3i(mOgreVectorsArray[3]);
 
     
     // create a slope -- a set of left and rigth path
@@ -86,10 +86,10 @@ void CullingManager::startTileCulling()
 {
     getIntersectionPoints();
 
-    mWalk.myArray[0] =  Vector3i(mOgreVectorsArray[0]);
-    mWalk.myArray[1] =  Vector3i(mOgreVectorsArray[1]);
-    mWalk.myArray[2] =  Vector3i(mOgreVectorsArray[2]);
-    mWalk.myArray[3] =  Vector3i(mOgreVectorsArray[3]);
+    mWalk.mVertices[0] =  Vector3i(mOgreVectorsArray[0]);
+    mWalk.mVertices[1] =  Vector3i(mOgreVectorsArray[1]);
+    mWalk.mVertices[2] =  Vector3i(mOgreVectorsArray[2]);
+    mWalk.mVertices[3] =  Vector3i(mOgreVectorsArray[3]);
     mWalk.buildSlopes();
     mOldWalk = mWalk;
     mOldWalk.prepareWalk();
--- a/source/camera/CullingManager.cpp
+++ b/source/camera/CullingManager.cpp
@@ -52,7 +52,7 @@ CullingManager::CullingManager(CameraManager* cameraManager):
 
 void CullingManager::cullTiles()
 {
-    oldWalk = mWalk;
+    mOldWalk = mWalk;
     mWalk.myArray[0] =  Vector3i(mOgreVectorsArray[0]);
     mWalk.myArray[1] =  Vector3i(mOgreVectorsArray[1]);
     mWalk.myArray[2] =  Vector3i(mOgreVectorsArray[2]);
@@ -64,13 +64,13 @@ void CullingManager::cullTiles()
 
     if(mDebug)
     {
-        LogManager::getSingleton().logMessage(oldWalk.debug(),LogMessageLevel::NORMAL);
+        LogManager::getSingleton().logMessage(mOldWalk.debug(),LogMessageLevel::NORMAL);
         LogManager::getSingleton().logMessage(mWalk.debug(),LogMessageLevel::NORMAL);
         
     }
 
     // reset index pointers to the begging of collections
-    oldWalk.prepareWalk();
+    mOldWalk.prepareWalk();
     mWalk.prepareWalk();
 
     newBashAndSplashTiles(SHOW | HIDE);
@@ -91,8 +91,8 @@ void CullingManager::startTileCulling()
     mWalk.myArray[2] =  Vector3i(mOgreVectorsArray[2]);
     mWalk.myArray[3] =  Vector3i(mOgreVectorsArray[3]);
     mWalk.buildSlopes();
-    oldWalk = mWalk;
-    oldWalk.prepareWalk();
+    mOldWalk = mWalk;
+    mOldWalk.prepareWalk();
     mWalk.prepareWalk();
     hideAllTiles();
     newBashAndSplashTiles(SHOW);
@@ -117,7 +117,7 @@ void CullingManager::stopTileCulling()
     // mOldMiddleLeft = mMiddleLeft;
     // mOldMiddleRight = mMiddleRight;
 
-    oldWalk = mWalk;
+    mOldWalk = mWalk;
     newBashAndSplashTiles(HIDE);
     showAllTiles();
     mCullTilesFlag = false;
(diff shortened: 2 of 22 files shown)


tomluchowski:
The culling mechanism sometime crashes when going outside/ on the border of  the game plane . Non the less, I please you to test it under other cases and report the bugs if you find it.

akien-mga:
I confirm that there's a big FPS gain when starting the tile culling, especially in big maps. It does crash when reaching the borders of the level, though.

Another issue seems to be that the minimap does not really work properly, probably because the tiles it should render are culled.

Bertram25:
Cool to see progress there! I do hope you'll be able to fix the bugs noticed by @akien-mga :)

tomluchowski:
Bertram25 : OK , may I set the limits for camera position ? Actually see my Issue >> Eliminating black space << please ... :)
What do you think about bringing the old mini-map ? It looked quite old - school and it was better than rendered mini-map IMHO . ( Which now suffers from custom tile culling ) .... 

@All ; Have you noticed this crash happens only for edges at X = 0 or Y =0 leaving the other pair intact ... seems this bug would be quickly found ....

hwoarangmy:
I've just tested and confirm the game exists when map borders are reached. Moreover, during game, many tiles are not displayed.

odscreenshot_2015-09-21_205018_0

tomluchowski:
Ok , I fixed all that , the only problem lurking is the minimap presentation layer . Would you like the old one back ?

Danimal696:
im a fan of the old one...

akien-mga:
I confirm that the crashes at the border are gone. I didn't see the issue in @hwoarangmy's screenshot so far, on the other hand when you move very quickly with the arrows / WASD, you can see some culled tiles in the direction where you're moving.

tomluchowski:
I tried to even enlarge the area with  myArray.zoom(1.2532); but it helps only a little. 
Could it be fixed  ? Maybe we could arrange lower maximal speed for the camera ? 
Problem is inside Ogre3d , that it does not link / unlink objects in SceneNodes' tree instantly .
For me it's less annoying than the whole board hoovering in black space ... 

I still expect some help with making it automaticly start startTileCulling and stopTIleCulling at the beginning and at the end from the other part of the team .

akien-mga:
@OpenDungeons/developers Tom merged all commits into one as the single commits were not really meaningful taken alone. It's ready for review I believe :)

Bertram25:
Cool! I'll have a look asap. (Busy as hell, but still around ;])

source/camera/CullingManager.cpp
```diff
@@ -0,0 +1,289 @@
...
+#include "camera/CameraManager.h"
+#include "gamemap/GameMap.h"
+
+#include "entities/Creature.h"
+#include "utils/Vector3i.h"
+#include "utils/LogManager.h"
+#include <sstream>
+#include <algorithm>
+
+using  std::set; using std::swap; using std::max; using std::min;
```
Bertram25: Please no using. Those leads to bad practices IMHO.

source/camera/CullingManager.cpp
```diff
@@ -0,0 +1,289 @@
...
+
+    
+    // create a slope -- a set of left and rigth path
+    mWalk.buildSlopes();
+
+    if(mDebug)
+    {
+        LogManager::getSingleton().logMessage(oldWalk.debug(),LogMessageLevel::NORMAL);
+        LogManager::getSingleton().logMessage(mWalk.debug(),LogMessageLevel::NORMAL);
+
```
Bertram25, line 69: Useless line here.

source/camera/CullingManager.cpp
```diff
@@ -0,0 +1,289 @@
...
+#include "utils/Vector3i.h"
+#include "utils/LogManager.h"
+#include <sstream>
+#include <algorithm>
+
+using  std::set; using std::swap; using std::max; using std::min;
+using  std::cerr; using std::endl;
+
+//! Values used to know whether to show and/or hide a mesh
+static const int HIDE =  1;
```
Bertram25, line 32: There are useless spaces on the next 4 lines. after the '=', and before ';'

source/camera/CullingManager.cpp
```diff
@@ -0,0 +1,289 @@
...
+            ss << "DxLeft, DxRight: " << double(DxLeft)/Unit << " " << double(DxRight)/Unit << " " ;
+            LogManager::getSingleton().logMessage( ss.str(),LogMessageLevel::NORMAL);
+            ss.str("");
+            ss << "xxLeft " << (xxLeft  >> mPrecisionDigits) << "xxRight " << (xxRight  >> mPrecisionDigits );
+            LogManager::getSingleton().logMessage( ss.str(),LogMessageLevel::NORMAL);
+            ss.str("");
+        }
+    
+        int64_t mm = ((std::min(xxLeft, xxLeftOld) >> mPrecisionDigits) << mPrecisionDigits) ;
+        if(std::min(xxLeft, xxLeftOld) < max(xxRight,xxRightOld) )
```
Bertram25: Please never ever do multiline ifs without brackets. The if clearly needs brackets in similar cases, and in fact, in this case, it should be inverted with a return to avoid useless cyclomatic complexity:

``` c++
if(std::min(xxLeft, xxLeftOld) >= std::max(xxRight,xxRightOld))
    return;
```

source/camera/CullingManager.cpp
```diff
@@ -0,0 +1,289 @@
...
+                {
+                    ss<< " x " <<  (xx >> mPrecisionDigits )<< " y " << (yy >> mPrecisionDigits ) << " " << (splash && (mode & SHOW)) << (bash && (mode & HIDE)) << endl;
+            
+                }
+                xxp = ((xx >>mPrecisionDigits) - 128.0);
+                yyp = ((yy >>mPrecisionDigits) - 128.0);
+                GameMap* gm = mCm->mGameMap;
+                if(bash && splash && (mode & HIDE) && (mode & SHOW))
+                {
+                    // Nothing
```
Bertram25, line 206: ??

source/camera/CullingManager.cpp
```diff
@@ -0,0 +1,289 @@
...
+    mMyRay[3]= Ogre::Ray (myvector[3], myvector[7] - myvector[3]);
+
+    std::pair<bool, Ogre::Real> intersectionResult;
+
+    for(int ii = 0; ii < 4; ++ii)
+    {
+        intersectionResult =  mMyRay[ii].intersects(mMyplanes);
+        if(intersectionResult.first)
+            mOgreVectorsArray[ii]= (mMyRay[ii].getPoint(intersectionResult.second));
+        if(intersectionResult.first)
```
Bertram25: Please invert that one, this is looking naive.

source/camera/CullingManager.cpp
```diff
@@ -0,0 +1,289 @@
...
+
+    return true;
+}
+
+bool CullingManager::onFrameEnded()
+{
+    return true;
+}
+
+/*! \brief Sort two Vector3i p1 and p2  to satisfy p1 <= p2 according to
```
Bertram25, line 261: Please document the header, not the implementation.

source/camera/CullingManager.h
```diff
@@ -0,0 +1,87 @@
...
+
+    bool mFirstIter;
+    CameraManager* mCm;
+
+    Ogre::Plane mMyplanes;
+    Ogre::Ray mMyRay[4];
+
+    bool mCullCreaturesFlag;
+    bool mCullTilesFlag;
+
```
Bertram25, line 100: Useless spaces.

source/camera/DummyArrayClass.h
```diff
@@ -0,0 +1,41 @@
...
+
+#ifndef DUMMYARRAYCLASS_H_
+#define DUMMYARRAYCLASS_H_
+
+#include "utils/Vector3i.h"
+
+#include <array>
+#include <vector>
+
+class DummyArrayClass
```
Bertram25, line 32: Could you describe a bit what this class is used for here?

source/camera/FastPODAllocator.h
```diff
@@ -0,0 +1,72 @@
...
+/// class A { };
+/// =>
+/// class A : public gms::FastPODAllocator<A> { };
+///
+/// Note that FastPODAllocator<> is not thread safe.
+/// Use as you wish.
+/// @see http://www.boost.org/
+/// @author Hannu Kankaanp�� (hkankaan@cc.hut.fi)
+template <class T>
+class FastPODAllocator
```
Bertram25, line 40: Out of curiosity, why are you using that piece of software?

source/camera/SlopeWalk.cpp
```diff
@@ -0,0 +1,283 @@
...
+
+#include "camera/SlopeWalk.h"
+
+#include <algorithm>
+#include <iostream>
+#include <sstream>
+
+extern const int mPrecisionDigits = 20;
+extern const int Unit = (1 << mPrecisionDigits);
+using std::cerr;
```
Bertram25: No using please.

source/camera/SlopeWalk.cpp
```diff
@@ -0,0 +1,283 @@
...
+    for(ii = top_left_index; ii != down_left_index ; ii+=3, ii%=4  )
+    {
+        leftSlopes.push_back((myArray[ii].x - myArray[ii-1].x) * Unit / (myArray[ii].y - myArray[ii-1].y ));
+        leftVertices.push_back(ii);
+    }
+    leftSlopes.push_back(0);
+    leftVertices.push_back(ii);
+}
+
+// reset indexes to the begginging of containers
```
Bertram25, line 74: Same here, please document the header instead.

source/camera/DummyArrayClass.h
tomluchowski, line 32: Yeah , you are right, I miss a few descriptions in the header file .

source/camera/CullingManager.h
tomluchowski, line 100: Pardon which spaces are useless? 
Do we agree to have some new lines when splitting a logic groups of correlated member items  , BTW ? 
The newlines in the end are useless .

Bertram25:
The code still has a lot of useless spaces and newlines all over the place, could you fix that?
Also, the headers copyright are stating 2014.

Btw, why are you using extern values of Units and precision digits? Couldn't this be avoided by setting a correct value an set it in a proper common header?

source/camera/FastPODAllocator.h
tomluchowski, line 40: This should be excluded, it is used with other , non - present mechanism here of creature culling when using QuadTrees.

source/camera/SlopeWalk.cpp
tomluchowski: Pardon then, should I write std:: before every standard function and class ?

source/camera/CullingManager.cpp
tomluchowski, line 261: So, ehem comments in implementation should go off ?

source/camera/CullingManager.cpp
tomluchowski, line 206: Could this one stay, I just wanted to have all cases in the if / else if / else if triad, this // nothing is to show that in the first case we do nothing.

source/camera/CullingManager.h
Bertram25, line 100: Err, wow, I'm tired, I meant useless newlines, sorry.

source/camera/CullingManager.cpp
Bertram25, line 206: Ah, if it's about readability, I'd propose this instead:

``` c++
//if(bash && splash && (mode & HIDE) && (mode & SHOW)) // We do nothing in that case
if (gm && bash && (mode & HIDE) && xxp>=0 && yyp>= 0 && xxp<mCm->mGameMap->getMapSizeX() && yyp < mCm->mGameMap->getMapSizeY())
    gm->getTile(xxp, yyp)->hide();
else if (gm && splash && (mode & SHOW) && xxp>=0 && yyp>= 0 && xxp<mCm->mGameMap->getMapSizeX() && yyp < mCm->mGameMap->getMapSizeY())
    gm->getTile(xxp, yyp)->show();
```

This way, you show you thought about that case and you avoid computing the if each time. WDYT about it?

source/camera/CullingManager.cpp
Bertram25, line 261: Eh eh, of course not. But this is not a comment, this is a function description.

source/camera/DummyArrayClass.h
Bertram25, line 32: Cool, thanks.

source/camera/FastPODAllocator.h
Bertram25, line 40: K, could you remove it then? Let's add it once it's useful only.

source/camera/SlopeWalk.cpp
Bertram25: Yes.

source/camera/CullingManager.cpp
tomluchowski, line 206: Actually , that way it brakes my code . The first case someone want to both show and hide some tile so we ignore that request. Without that I would have to give additional constraints to compensate for the lack of that first case .

Bertram25, line 206: True. Nevermind about this then. But if you add brackets to the first if, then please brackets to the others for better readability.

Bertram25:
I'll have a look asap. :)

tomluchowski:
Maybe over a  weekend you could hopefully find some time ....

Bertram25:
Yeah you're right sorry for keeping you waiting. :S
I'm quite busy but you're now on top of my spare-time TODO list. :)

tomluchowski:
When you are done , we could take care of creature's culling .

source/camera/CameraManager.h
```diff
@@ -24,6 +24,8 @@
 #define CAMERAMANAGER_H_
 
 #include "camera/HermiteCatmullSpline.h"
+#include "gamemap/GameMap.h"
+#include "entities/Tile.h"
```
hwoarangmy: You should avoid to include headers as much as possible in other headers. Especially gamemap and tile which are pretty much used (and likely to be changed). That increases the compilation time.

source/camera/CameraManager.h
```diff
@@ -48,6 +51,10 @@ enum class ViewModes : uint16_t
 
 class CameraManager
 {
+    friend class Console;
+    friend class CullingManager;
+    friend class ODFrameListener;
```
hwoarangmy: IMHO, we should not use friend classes except for strongly linked classes. It is better to add setters/getters when you need. The problem with friend classes is that is makes hard to follow where a private variable is used (when you try to understand what some code does)

source/camera/CameraManager.h
```diff
@@ -185,7 +192,8 @@ class CameraManager
     Ogre::Camera* mActiveCamera;
     Ogre::SceneNode* mActiveCameraNode;
 
-    TileContainer* mGameMap;
+
```
hwoarangmy, line 195: useless empty line

source/camera/CullingManager.h
```diff
@@ -0,0 +1,103 @@
...
+    bool onFrameEnded();
+
+    // set the new tiles
+    int64_t bashAndSplashTiles(int64_t);
+    void newBashAndSplashTiles(int64_t);
+
+    void sort(Vector3i& p1, Vector3i& p2, bool sortByX);
+
+    //! \brief Activate/deactivate debug output.
+    bool mDebug;
```
hwoarangmy, line 80: is this really needed ? In theory, you should use OD_LOG_DBG for that. However, if you think this will be needed for more than your current dev, you can let it.

source/camera/CullingManager.h
```diff
@@ -0,0 +1,103 @@
...
+
+private:
+    SlopeWalk mWalk, oldWalk;
```
hwoarangmy: oldWalk should be mOldWalk (or mWalkOld). Moreover, please use 2 lines for 2 such classes (it is pretty easy to miss there are 2 classes when declared like that)

source/camera/CullingManager.h
```diff
@@ -0,0 +1,103 @@
...
+
+    
+    Ogre::Vector3 mOgreVectorsArray[4];
```
hwoarangmy: A few comments for theses variables would be welcome as you are working on something technical most readers (including me) won't be used to

source/camera/DummyArrayClass.cpp
```diff
@@ -0,0 +1,97 @@
...
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "camera/DummyArrayClass.h"
+
+#include <algorithm>
+
+void DummyArrayClass::sort()
+{
+    std::vector<Vector3i> ll[4];
```
hwoarangmy, line 24: please give names with meaning. Remember that people reading this code are not in your head and every merged code might be modified by anybody.

source/camera/DummyArrayClass.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+void DummyArrayClass::sortByPhi(std::vector<Vector3i> &ll, Vector3i& cc)
+{
+    if (ll.size() < 2)
+        return;
+
+    std::sort(ll.begin(), ll.end(),
+              [&cc](Vector3i vv, Vector3i ww)
+              {
+                  return static_cast<double>(vv.y - cc.y) / static_cast<double>(vv.x - cc.x) > static_cast<double>(ww.y - cc.y) / static_cast<double>(ww.x - cc.x);
```
hwoarangmy, line 85: lambdas are nice ^^

Note that here, you doesn't test that vv.x - cc.x != 0 and ww.x - cc.x != 0. Are you sure it cannot be the case ?

source/camera/SlopeWalk.cpp
```diff
@@ -0,0 +1,270 @@
...
+    leftVertexPassed(false),
+    rightVertexPassed(false)
+{
+}
+
+void SlopeWalk::buildSlopes()
+{
+    int ii;
+    myArray.sort();
+    myArray.zoom(1.2532);
```
hwoarangmy: hardcoded value like that looks strange

source/camera/SlopeWalk.h
```diff
@@ -0,0 +1,103 @@
...
+
+#include "utils/Vector3i.h"
+
+#include <deque>
+#include <array>
+#include <string>
+
+using std::deque;
+using std::array;
+using std::string;
```
hwoarangmy: Please do not use using

source/entities/EntityBase.h
```diff
@@ -19,6 +19,7 @@
 #define ENTITYBASE_H
 
 #include "entities/GameEntityType.h"
+#include "render/RenderManager.h"
```
hwoarangmy: Adding a dependency to rendermanager here is a real bad idea. That will make a recompilation of the whole project if you change something in the render manager

source/entities/EntityBase.h
```diff
@@ -127,6 +128,15 @@ class EntityBase
     { return false; }
     virtual void drop(const Ogre::Vector3& v)
     {}
+    inline void show()
+    {
+        RenderManager::getSingleton().rrAttachEntity(this);
```
hwoarangmy: this doesn't need to be inlined to avoid the render manager dependency

source/entities/GameEntity.h
```diff
@@ -24,6 +24,7 @@
 #define GAMEENTITY_H
 
 #include "entities/EntityBase.h"
+#include "render/RenderManager.h"
```
hwoarangmy: same here

source/modes/ConsoleCommands.cpp
```diff
@@ -567,6 +568,12 @@ Command::Result cTriggerCompositor(const Command::ArgumentList_t& args, ConsoleI
     return Command::Result::SUCCESS;
 }
 
+Command::Result cStartTileCulling(const Command::ArgumentList_t& args, ConsoleInterface& c, AbstractModeManager&)
+{
+    ODFrameListener::getSingleton().getCameraManager()->mCullingManager->startTileCulling();
```
hwoarangmy: It would be better to use something like ODFrameListener::getSingleton().getCameraManager().startTileCulling();

Using a CullingManager or not should be hidden from the API

source/network/ODClient.cpp
```diff
@@ -382,6 +384,7 @@ bool ODClient::processOneClientSocketMessage()
                 startY = 0.0;
 
             frameListener->resetCamera(Ogre::Vector3(startX, startY, MAX_CAMERA_Z));
+            frameListener->mCameraManager.mCullingManager->startTileCulling();
```
hwoarangmy: same here

source/utils/Vector3i.h
```diff
@@ -0,0 +1,58 @@
...
+class Vector3i
+{
+public:
+    Vector3i():
+        x(0),
+        y(0),
+        z(0)
+    {}
+
+    Vector3i(int64_t nx, int64_t ny , int64_t nz):
```
hwoarangmy, line 39: int64 ? Is it really needed ?

source/camera/CullingManager.h
```diff
@@ -0,0 +1,103 @@
...
+    Vector3i mTop, mBottom, mMiddleLeft, mMiddleRight;
```
hwoarangmy: Those do not seem to be used. Please do not let useless variables

source/camera/CullingManager.h
```diff
@@ -0,0 +1,103 @@
...
+    Vector3i mOldTop, mOldBottom, mOldMiddleLeft, mOldMiddleRight;
```
hwoarangmy: same here

Bertram25:
@hwoarangmy Thanks for doing the job! :)
I also think we should make the custom culling a setting option, btw. (This would make it perfect.)

source/camera/DummyArrayClass.cpp
```diff
@@ -0,0 +1,97 @@
...
+        );
+}
+
+
+Vector3i& DummyArrayClass::operator[](int ii)
+{
+    while(ii < 0)
+        ii += 4;
+
+    return mMyArray[ii % 4];
```
hwoarangmy: It would be nice to add a check here that you are not over the array size

source/camera/SlopeWalk.cpp
tomluchowski: Should I put there const double value then ?

source/camera/CullingManager.h
tomluchowski, line 80: "you can let it" You mean having a debugging code in CullingManager.cpp or using the mDebug flag ?

source/camera/DummyArrayClass.cpp
tomluchowski, line 24: It's just the array with a bunch of auxilary functions , how should I call it ? hwoarangmy arrays :P ?

source/entities/EntityBase.h
tomluchowski: So we sacriface game speed for the sake of compilation speed ? Ok I can un - inline that function.

tomluchowski:
@Bertram25 No please, don't do it . I spend  a lot of time polishing that feature and don't want it to get it wasted .

source/utils/Vector3i.h
tomluchowski, line 39: Yeap this is size of int allowing enough fixed precision digits for my needs.

source/entities/EntityBase.h
hwoarangmy: No worries, you won't see the game speed difference. And if this call is called so much that inlining it would change something, it will be time to inline it.
Trying to have readable code if the first goal we should have when coding. Performances will come after (where needed). The real time wasters are almost never where you expect them.

source/camera/SlopeWalk.cpp
hwoarangmy: Yes, it would be better. And also adding a comment where you define it to try to explain why you used this magic number would be good (so that we know for which reason it may be useful to change it)

source/camera/CullingManager.h
hwoarangmy, line 80: both. But only if it is useful later, once your PR is merged. In most cases, there should be no debug code.

source/camera/DummyArrayClass.cpp
hwoarangmy, line 24: That would indeed be a nice name ;-)
You know, usually, if you don't know how to name a variable, that's because there is a flaw in your code. And usually, this kind of variable ends called "myArray" and such.

tomluchowski:
@hwoarangmy Ok , how about now ?

source/camera/SlopeWalk.h
```diff
@@ -0,0 +1,101 @@
...
+    int top_left_index;
+    bool leftVertexPassed;
+    bool rightVertexPassed;
+
+
+public:
+    SlopeWalk();
+    void printState();
+    // A place for raw points
+    DummyArrayClass myArray;
```
hwoarangmy: Doesn't follow code guidelines

hwoarangmy:
@tomluchowski That's better, indeed. However, you have not removed the now useless headers from the cameramanager and not answered some of my questions in code.

Concerning SlopWalk, could you please explain what the values in the arrays are supposed to be ? I couldn't guess from reading the code.

Note that when you will have to cull entities on tile, you can easily do it by using Tile::mEntitiesInTile (on client side, creatures and most of the entities are added to mEntitiesInTile even the moving ones).

source/camera/SlopeWalk.h
tomluchowski: DummyArrayClass myArray should be private , right ?

hwoarangmy: if possible. And name should begin with m then capital letter for the name

source/camera/DummyArrayClass.cpp
tomluchowski, line 85: I cannot find an argument for or against .... I have never seen sig zero exception to be honest.

tomluchowski:
How about now ?

hwoarangmy:
I will review the whole PR tomorrow but I guess it should be good.
I will also try it.

hwoarangmy:
I've tried but there is still the bug with the minimap. However, once it is fixed, it is mergeable for me.

@tomluchowski For my information, I've seen you use 4 Ogre rays to know the tiles the player can see (I guess by checking the 4 corners of the screen). Once there, I would have expected a simple linear extrapolation to know all the tiles within vision. Is the algorithm you used better in some way ? I must admit I didn't take enough time to really understand how it works.

tomluchowski:
It is much better in many ways , first of it does not check for visibility of every Tile there is on the board, second is it attaches/ detaches Tile only when it has to (when it appeared or disappeard due to previous iteration ) ; naive algorithm would hide all the Tiles from the previous iteration and compute the new set of all visible TIles -- and attach them .

tomluchowski:
"I would have expected a simple linear extrapolation to know all the tiles within vision"
Yeah I also extrapolate lineary the borders of visible area. Implicitly.

tomluchowski:
I've tried but there is still the bug with the minimap. However, once it is fixed, it is mergeable for me.
Well I propose to return to the old pixel based version of minimap.
Danimal's for it ... who else  ? 
Could you just shift that to separate issue ?

hwoarangmy:
> naive algorithm would hide all the Tiles from the previous iteration and compute the new set of all visible TIles -- and attach them

Yes, that's what I was expecting. Simple. To do better, your algorithm has really to be fast. I should have a deeper look as it is interesting ^^
The good thing with the current minimap is that is allows to see which tiles are culled. And from what I can tell, it is pretty good as we are not far from the tiles in vision.

> Well I propose to return to the old pixel based version of minimap.

IMHO, removing a feature because it is broken by another one is not the way to go. If I implement gem tiles and break tile culling in my way, would you think disabling tile culling would be a good solution ?
Moreover, the old minimap do not work with D3D renderer. I know we only want to support OpenGL officially but as so far, D3D works out of the box, I would think it is a waste to disable it. Moreover, because it doesn't work with D3D, I would guess there is a bug / wrong use of the texture for the old minimap.

Bertram25:
> > Well I propose to return to the old pixel based version of minimap.
> 
> IMHO, removing a feature because it is broken by another one is not the way to go. If I implement gem tiles and break tile culling in my way, would you think disabling tile culling would be a good solution ?

@tomluchowski  We're a spare-time project, so finishing what we start is a key point if we want to stay relevant as a project. Otherwise, we'll be leaving a trail of unfinished features behind us, which is a known common cause of spare time projects death. The former OD team is plainly concerned by that sadly. But I am, too, somehow.

Bertram25:
Btw, I consider both minimaps broken or sub-optimized, but again, we're a spare-time project. ;)

Bertram25:
Btw, what is left to do here, apart the minimap fix?

hwoarangmy:
> Btw, what is left to do here, apart the minimap fix?

I would say only the minimap fix. Apart from that, the features seems to work well. I think @tomluchowski wants to go for creature culling but IMHO, it would be better to do that in a new PR

Bertram25:
Yes, he should do that in a new PR, the poor one already bleed a lot to get this one, let's not be too cruel. ;)

tomluchowski:
"Btw, I consider both minimaps broken or sub-optimized, but again, we're a spare-time project. ;)"

sub-optimized ? 
What do you mean ? 
Yes I bleeded quite much , could somone give me a helping hand , or give advice how to finish this off ?

hwoarangmy:
> Yes I bleeded quite much , could somone give me a helping hand , or give advice how to finish this off ?

I've not looked deep enough your code to know every subtility but you said that only changed tiles are culled. That would mean that you somehow keep a list of culled tiles. If it is the case, the easiest way would be to unhide tiles before rendering the minimap (I don't know how hard it will hit performances, though). You might also want to somehow cull tiles not rendered in the minimap.

tomluchowski:
" That would mean that you somehow keep a list of culled tiles." 
The secret is I never keep such a list. I could unhide the tiles, but doing that to 400x400 or 100x100 tiles at once would stall down everything.  Introducing second camera is painfully difficult and beyond my head .

akien-mga:
@OpenDungeons/developers So what's the status quo? It looks like we have an interesting new feature that also happens to break the minimap, and @tomluchowski doesn't know how to fix it. Should we merge and have someone else try to fix it, or keep it around until we find a solution?

hwoarangmy:
For my part, as I've said in the comments, I couldn't easily follow the code by reading it (complex  code and not enough comments). In a professional context, that would be enough to reject the PR. But this is a free time project so we don't have to be too strict. However, I'm not convinced by the use of int64_t. By experience, it is likely to hide a bug. Moreover, if @tomluchowski, the author, cannot find the list of impacted tiles from his own code, it doesn't sound very good.
From what I've seen, finding the 4 tiles based on rays might be a good idea (since rays do not seem to be too cpu consuming). But the algorithm seems too complex for such a simple need and not readable (which means hard to maintain).
And last (but not least), it breaks the minimap.
If someone wants to take the time to really understand what the code does and fix the minimap, I won't be against merging. But if I had to go for a culling algorithm, I would go for something more "naive" but I'm pretty sure at least as efficient and much more readable.
As for now, I will vote against merging.

Bertram25:
Overall, I must say I agree with @hwoarangmy 's points which have to be fixed before going further.

> But this is a free time project so we don't have to be too strict.

Here, I disagree. Please note that this very project (before the fork) was dead because of such PR were allowed, and even without any review. As a spare-time project, I'd rather say it's even more important to agree about merging PRs that are readable only. More than in a professional environment.
The area where we shouldn't be too strict, and we aren't IMHO, is about the fact to communicate within code reviews with much less rigor, and use fancy comments and jokes wherever we see fit.

But please, never undermine quality in spare-time project, you'll pay it even more than when there is someone paid to fix it.

akien-mga:
Ok, I tend to agree with both of you. I haven't reviewed the code myself, but if you say that it hasn't the readability we strive for usually, then it's probably best kept out of the main branch.

Since the result is still relatively interesting, what do you think about merging the PR in a specific branch (i.e. not development, but a tile-culling branch that we would create)?

hwoarangmy:
> it's even more important to agree about merging PRs that are readable only

I would say that the most important thing is that we need to have a code repository as easy to get in as possible to allow newcomers to join.

> The area where we shouldn't be too strict

I just wanted to say that it is a free time project and everybody can propose what he wants (and we cannot ask more than what one wants to do). However, it is our responsability to keep the code as clean and accessible as possible and to reject code we don't find good enough to be merged.

> Since the result is still relatively interesting, what do you think about merging the PR in a specific branch

Honestly, I don't see the point. If the code is good enough, it should be merged in the main repository. If not, it should be dropped. If we keep it in a separate branch, it will rot until not mergeable and we will forget about it.

tomluchowski:
"If someone wants to take the time to really understand what the code does and fix the minimap, I won't be against merging."
I am not any algorithm God, the solution I come up with is really simple, I really tried to explain in comments what it does. It's simple interpolation based on the idea that the line formula is y = ax + b .

 The problem with minimap is when I started working on culling feature it was 2D pixel based version, so my hiding or unhidding some tiles didn't broke anything. Both minimaps would cause a problem for me to work properly, the other issue being that you seem to be undecided which version should stay for good. The easiest way for me would be to go for 2d pixel verison and ask on ogre3d forum why our code does not work under DirectX, and what to do to make it working, despite me not having Windows installed...

I really put much effort and don't want to get it discarded, I wanted to have position in CV about an open- source game somehow improved by me, however clumsy I am at it tough ....

Bertram25:
I'm closing this one as it cannot be finished for now. I've linked with corresponding issue and will close it for now.

PR #959, opened 2015-09-26
Implemented creature skills

hwoarangmy:


Danimal696:
i dont understand a few of the new concepts(based on new release on forums):

> FightIdleDist       Distance from the nearest enemy creature that the creature will try to let when no skill can be used

Is that some kind of foe check? if farther than 1 (default value), it wont use skills?
Where are the damage and distance values now?

Also, there is a small bug, the stunned value and KO value seems the be same, so i cant capture creatures to jail since they wake up too fast after being KO´ed

hwoarangmy:
> Is that some kind of foe check? if farther than 1 (default value), it wont use skills?

No. That just means that if there is no skill available (the creature have used everything and the cooldown is running on for each), the creature will try to be N tiles from the closest foe. If 1, the creature will stay at melee range even during its attack cooldown. That allows to do some hit and run creatures. For my tests, I've achieved good results by increasing wizards speed and lowering warmup. I could kill a troll with this wizard ^^

> Where are the damage and distance values now?

In the corresponding skill (melee or ranged) in the creature definition. I let the same values as it was.
https://github.com/hwoarangmy/OpenDungeons/blob/development/config/creatures.cfg#L231

> Also, there is a small bug, the stunned value and KO value seems the be same, so i cant capture creatures to jail since they wake up too fast after being KO´ed

The stunned value is defined in the creature definition. The KO value is defined in global.cfg. I've checked and they are correctly used. ATM, KO value is 120 turns which is more or less 1:20 minutes.
However, I don't understand your problem. If the creature wakes up, that means that it is  stunned, not KO. If it is KO and you cannot take it to jail, it should die, not wake up. If you are talking about your own creatures, you cannot put them in jail ATM.

Danimal696:
no, it was with enemy creatures, they were ko´ed, but my kobolds didnt have any time to get them to prison, they wake up real fast, but it migth be a problem of the Stonekeep level, ill try later other level.

hwoarangmy:
If they are KO and wake up alone, that's because an enemy keeper dropped them. Note that AI tries to save wounded creatures so it might not be that easy to KO them...

Danimal696:
they were ferals tentacles, the ones at the rigth of were you start, near the lava pool with dragons.

hwoarangmy:
Well, I've just tried the map and had the expected result. After my creatures KO the tentacles, I realized I had forgotten to build the prison room. So, I even had time to dig a 5x5 tiles, use the console command to unlock the rooms, build the prison and drop kobolds to take the prisoners to the jail.
Then, the other tentacles started to die (none woke up). I don't know what is happening to you but I can't  reproduce.

Danimal696:
would you believe? i cant reproduce, it works fine now 2 out of 2, so forget about it

hwoarangmy:
ok cool ^^

Danimal696:
Since im evil and you shamed me im presenting another bug to you, after saving and reloading, prisioners forget they are jailed and do a prison break! the nerve!

Related to this, would this icon be better for prisioner creatures?
http://game-icons.net/lorc/originals/manacles.html

hwoarangmy:
> after saving and reloading, prisioners forget they are jailed and do a prison break! the nerve!

Nice catch. Many stuff is not saved ATM. But I guess this one is worst than most others ^^
I will open an issue

> Related to this, would this icon be better for prisioner creatures?

It's pretty close to the one I had originally used. But the current one is not bad, too. Not sure it is worth the effort:
http://game-icons.net/lorc/originals/handcuffs.html

Bertram25:
Leave me one day and I'll do the code review. :)

Bertram25:
There are a few things but wow, overall the feature list is impressive. :D I'll test asap.

hwoarangmy:
Also note that by consequence, melee and ranged attack now have cooldown and warmup ^^

akien-mga:
Code looks very good to me, I'll try to test it tonight.

Bertram25:
Will test tonight. :)

akien-mga:
It simply works great :D It already feels like battles are way more dynamic :)
Let's merge the beast!

Bertram25:
I just finished a game. Seems @akien-mga has beaten by ~ 40min ;)
Great to see this merged.

PR #964, opened 2015-10-03
Add Roach creature in Keeper faction

akien-mga:
Currently has the exact same definition as the Spider (including MoodSpider)
but we should probably make it a bit more specific later on.


Bertram25:
tested and it looks good. :) Let's add the beast as a nice variety.

PR #970, opened 2015-10-04
Made the settings button on the main menu more visible.

### LINKED ISSUE #932: Settings should be made easier to find in the main menu
I've been looking for some time for a way to make the game windowed... Until I've figured out that I have to start the game to do so. I think it's good to have settings menu in main screen, just after the game has started.

Bertram25:
Fixes #932

### DIFF
--- a/gui/MenuMain.layout
+++ b/gui/MenuMain.layout
@@ -50,13 +50,10 @@
             <Property name="Area" value="{{0.5,170},{0.5,40},{0.5,310},{0.5,120}}" />
             <Property name="Text" value="Quit" />
         </Window>
-        <Window type="OD/ImageButton" name="SettingsButton" >
-            <Property name="Area" value="{{1,-27},{0,5},{1,-2},{0,30}}" />
-            <Property name="ButtonImage" value="OpenDungeonsIcons/OptionsIcon" />
-            <Property name="AlwaysOnTop" value="True" />
-            <Property name="TooltipText" value="Options" />
-            <Property name="ClippedByParent" value="False" />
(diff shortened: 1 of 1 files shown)


Bertram25:
I finally did it. ;)
Feel free to comment on whether it is enough.

akien-mga:
Looks good to me. I still have the overall impression (on a 1920x1080 monitor, even though I play in 1600x900, which is still quite hi-dpi) that we don't really fill the available space, and the tiny "settings" button looks... tiny :)

But I think it's good enough for the time being. In the future, I'd like to have us think of a potentially cooler design for the main menu, e.g. with animations from the game, etc.. A bit like what Sumwars has: http://sumwars.org/wiki/File:Sumwars_0.5.5_mainmenu.jpg (also OGRE + CEGUI)

Bertram25:
> But I think it's good enough for the time being. In the future, I'd like to have us think of a potentially cooler design for the main menu, e.g. with animations from the game, etc.. A bit like what Sumwars has: http://sumwars.org/wiki/File:Sumwars_0.5.5_mainmenu.jpg (also OGRE + CEGUI)

Honestly, I completely agree the main menu needs a revamp in a cooler way. I even think we should linke the first screen to the campaign. The best idea I can come up with, would be to display a big boss on its throne, preparing to fight, or something. and the usual buttons in corners. But that's some work, so maybe this can wait until we have more on that part?
We could also hunt a few other background to make some transition there?

hwoarangmy:
Well, for my part, I think the actual main menu is not that bad. So, until we have a final idea of what we want, I would be to not loose time on changing it.

PR #979, opened 2015-10-05
Read creature mood like all the other config parameters

hwoarangmy:


hwoarangmy:
@Bertram25 @akien-mga WDYT of the polymorphic factory vs function pointer we use ATM ?

Bertram25:
> @Bertram25 @akien-mga WDYT of the polymorphic factory vs function pointer we use ATM ?

Not sure to get what you want me to say here, in fact. :)

hwoarangmy:
@Bertram25 I'm saying: do you think the polymorphic factory like done here is better than what we did for the other factories: use function pointers like in 
https://github.com/OpenDungeons/OpenDungeons/blob/development/source/spells/SpellManager.h#L54

For my part, I prefer the polymorphic way. It is more c++ish and it would allow to use the managers in the modes in a better way. For example to remember the exact tiles hovered while moving the mouse (to build a bridge at once). That's stuff we can do at the moment by using ugly static variables but using a factory will allow to gather all building stuff in the same class as well as using polymorphism to handle common cases

Bertram25:
I do also prefer the polymorphic way, as it's a more common way to code in C++ indeed, but also because it simply avoids pointers, whatever the kind of pointer, just for one thing.
The factory may also have other benefits. :)

Bertram25:
@hwoarangmy Mind if I wait for the awakeness fix to test?

hwoarangmy:
> Mind if I wait for the awakeness fix to test?

@Bertram25 I would prefer to open an issue because I'm on something else. And since we are going to change the word, I guess there is no need to fix the typo

hwoarangmy:
@Bertram25 Done in #983

Bertram25:
ok, cool :)

hwoarangmy:
@Bertram25 Are you testing ? If not, I will add some more commits

Bertram25:
@hwoarangmy Go ahead. I'm finishing a patch for VT. :) I'll test after that.

akien-mga:
BTW, I won't be around much until next tuesday. I'll catch up later :)

hwoarangmy:
> BTW, I won't be around much until next tuesday. I'll catch up later :)

ok. Have a nice trip ;-)

hwoarangmy:
@Bertram25 I've added the commits (not very big this ones). The next one will be bigger (I will wait for this one to be merged). I'm trying to allow creature behaviour to be customizable by adding plugins like creature skills or room afinity. I'm not done yet but it is already not bad.

Bertram25:
> BTW, I won't be around much until next tuesday. I'll catch up later :)

Have fun :)

Bertram25:
I'll test asap.

akien-mga:
> I'll test asap.

Me too. Hopefully sooner :P

akien-mga:
I'll merge as I know @hwoarangmy has some niceties cooking in his local repo. As both @Bertram25 and I seem to be somewhat busy lately, I guess we shouldn't be too strict about merging quality pull requests such as this one once the code has been reviewed; we'll still spot potential regressions easily enough once it has been merged I guess :)

hwoarangmy:
> we'll still spot potential regressions easily enough once it has been merged I guess

Note that this PR as been tested by @Danimal696 since it is included in the win32 de binaries

PR #992, opened 2015-10-14
Implemented creature behaviour + 2 new creatures + many stuff

hwoarangmy:
Creature behaviour allows to customize what creatures do in the dungeon by adding plugins in creature config files (like the skills or the room affinities).


akien-mga:
Woot, I knew we shouldn't let PRs idle for too long :P

hwoarangmy:
Most commits here are small. But some are not ^^

akien-mga:
@hwoarangmy: Could you give us (or maybe mostly "me" since @Bertram25 is probably familiar with the concept) some details about how the new "factories" system work compared to how it was done previously, and what are the inherent advantages? I'll admit there are too many code changes for me to clearly understand how it will work :-D

akien-mga:
Well, 27 commits, that was a big one :)
@Bertram25 We should test and merge it asap before @hwoarangmy adds another 50 commits on top of it, including the 0.7.0 release tag :P

hwoarangmy:
> Could you give us (or maybe mostly "me" since @Bertram25 is probably familiar with the concept) some details about how the new "factories" system work compared to how it was done previously, and what are the inherent advantages?

First, it allows to get ride of th function pointers. Moreover, it makes things clearer. Previously, you had to see in the manager implementation to check which are the static functions (the equivalent of what is done in the factories). Now, you know exactly what you need to implement by looking at the factory definition. Lastely, you can benefit from polymorphism to share code.

To make things clear, most of that is done with the factory could have been done with static functions/variable. However, using factories help to follow an object rule: each object should have 1 and only 1 thing to do. Here, the factory is in charge of creating the creature behaviour and the creature behaviour is in charge of tuning the creature.

Bertram25:
If you can wait up to sunday evening. I'll do a fulle code review at that time + tests. :)

Bertram25:
@hwoarangmy How many people are you? ;P

hwoarangmy:
> If you can wait up to sunday evening. I'll do a fulle code review at that time + tests. :)

ok ^^

> @hwoarangmy How many people are you? ;P

Many are my names in many countries, Mithrandir among the elves, Tharkûn among the dwarves; Olórin I was in my youth in the west that is forgotten, in the south Incánus, in the north Gandalf; to the east I go not.

akien-mga:
It's growing!! Sauve qui peut !

@Bertram25 I'll try to find some time to test it all tonight, but IMO we could also merge the beast just based on our knowledge that @hwoarangmy also does quality stuff. We are still deep in the development phase, so there is still time to iron out every possible regression or bug later on.

hwoarangmy:
> there is still time to iron out every possible regression or bug later on

There might be about creature behaviour. But I think we will see them with time while playing.

akien-mga:
I had a very quick test and did not spot any huge regression, so let's merge. First issue I saw so far is that saved games can't be loaded apparently, I guess with all the recent level format changes something may have been forgotten.

akien-mga:
@OpenDungeons/developers We just reached 3000 commits, 3001 with the merge notification :) Congrats! :D

hwoarangmy:
> First issue I saw so far is that saved games can't be loaded

That's because we added new tile types. But note that before the 1.0, we shouldn't bother about saved games. After each commit, saved games/replays may be broken. That will be very hard to maintain (especially without automatic tests) so we should start to think about that only when we will have the need.

akien-mga:
> But note that before the 1.0, we shouldn't bother about saved games. After each commit, saved games/replays may be broken.

Yes I agree. If we can manage to get it working when stabilising for 0.7.0, it would still be a nice-to-have for players though. But it should indeed be a strong requirement only after 1.0.

akien-mga:
Checking this PR for the changelog. God it was huge!! :D

PR #993, opened 2015-10-14
Editor: activate newly-placed traps by default

### LINKED ISSUE #988: Traps layed in editor mode are unbuilt in level
their default activation value is 0 instead of 1 in level file, so all of them dont appear in level

akien-mga:
Fixes #988.

I'm not 100% sure this is the best way to do it, so I'm open to critics :)


hwoarangmy:
SF

PR #994, opened 2015-10-19
Refactor logging3

knight666:
I've added new logging macros, which use printf syntax and add a module parameter.

Old syntax:

```
OD_LOG_INF("OpenDungeons" + OD_VERSION + " rocks!");
```

New syntax:

```
OD_TRACE_INF("OpenDungeons %s rocks!", OD_VERSION);
```

Old output:

```
[20151019_160313] OpenDungeons 1.7.0 rocks!
```

New output:

```
(16:03:13) (odapplication) [TRIVIAL] OpenDungeons 1.7.0 rocks!
```

I've also introduced the concept of "logging sinks", which can be used to direct logging output to files, consoles or anything else. This change paves the way to allowing logging to be turned on or off per module at runtime, but will require a large refactor (all logging calls will have to be changed).


CMakeLists.txt
```diff
@@ -389,8 +389,15 @@ set(OD_SOURCEFILES
     ${SRC}/utils/FrameRateLimiter.cpp
     ${SRC}/utils/Helper.cpp
     ${SRC}/utils/LogManager.cpp
-    ${SRC}/utils/LogManagerFile.cpp
-    ${SRC}/utils/LogManagerOgre.cpp
+    ${SRC}/utils/LogManager.h
+    ${SRC}/utils/LogMessageLevel.h
```
hwoarangmy: You should not include headers in the source list

source/ODApplication.cpp
```diff
@@ -49,6 +51,11 @@
 void ODApplication::startGame(boost::program_options::variables_map& options)
 {
     ResourceManager resMgr(options);
+
+    LogManager logMgr;
+    logMgr.addSink(std::shared_ptr<LogSink>(new LogSinkConsole()));
```
hwoarangmy: interesting idea. But wouldn't a unique_ptr be better ?

source/ODApplication.cpp
```diff
@@ -87,8 +92,6 @@ void ODApplication::startServer()
 void ODApplication::startClient()
 {
     ResourceManager& resMgr = ResourceManager::getSingleton();
-    LogManagerOgre logManager(resMgr.getLogFile());
-    logManager.setLogDetail(LogMessageLevel::NORMAL);
```
hwoarangmy, line 91: I will check but if the log manager was launched here too, I think that's because it is not launched the same way in client (normal way) or in server (using a command line, allows to start a server where the clients will connect to play the game)

source/utils/Helper.h
```diff
@@ -30,6 +30,10 @@
 #include <vector>
 #include <cstdint>
 
+#ifdef _MSC_VER
+    #define snprintf _snprintf
```
hwoarangmy: In the coding guidelines, we try to avoid C code as much as possible. That's why we don't use printf and its family

source/utils/LogManager.cpp
```diff
@@ -21,3 +21,60 @@ template<> LogManager* Ogre::Singleton<LogManager>::msSingleton = nullptr;
...
+        31,
+        "%02d:%02d:%02d",
+        now->tm_hour,
+        now->tm_min,
+        now->tm_sec);
+
+    // filename
+
+#if WIN32 || _WINDOWS
+    const char* path_last_slash = strrchr(filepath, '\\');
```
hwoarangmy: Usually, we use boost for handling file path. That would avoid this kind of check that can lead to errors (for my part, I use '/' on windows as path separators and it works).

source/utils/LogManager.cpp
```diff
@@ -21,3 +21,60 @@ template<> LogManager* Ogre::Singleton<LogManager>::msSingleton = nullptr;
...
+#endif
+    const char* filename = (path_last_slash != nullptr) ? (path_last_slash + 1) : filepath;
+
+    // message
+
+    char message_formatted[1024] = { 0 };
+
+    va_list arguments;
+    va_start(arguments, format);
+    vsprintf(message_formatted, format, arguments);
```
hwoarangmy: you should not use vsprintf. Especially in a log function where unexpected data might be used (like a very long nickname or something). vsnprintf would be used here to avoid memory corruption.

source/utils/LogManager.cpp
```diff
@@ -21,3 +21,60 @@ template<> LogManager* Ogre::Singleton<LogManager>::msSingleton = nullptr;
...
+    va_end(arguments);
+
+    for (const auto& sink : mSinks)
+    {
+        sink->write(level, module, timestamp, filename, line, message_formatted);
+    }
+}
```
hwoarangmy: Missing a new line here

CMakeLists.txt
knight666: Then how can I force CMake to include the headers in the generated vcxproj? I rely on Visual Assist X's "Open File in Solution" dialog to jump between files in a project.

source/ODApplication.cpp
knight666: That does sound sensible. I'll give it a try in the morning. :)

EDIT: Experiment result: failure. It's _possible_ to store unique_ptr's in a vector, but it's recommended against. See here: http://stackoverflow.com/questions/3283778/why-can-i-not-push-back-a-unique-ptr-into-a-vector

source/ODApplication.cpp
knight666, line 91: "startGame" calls both "startServer" and "startClient". Both implementations can now use the same logging mechanism, using the sinks. I added a LogSinkOgre as well, but only realized after the fact that it writes to a similar log file.

source/utils/LogManager.cpp
knight666: Good advice! I'll take a look at how it's done in other parts of the codebase.

source/utils/LogManager.h
```diff
@@ -18,45 +18,64 @@
...
+    // DEPRECATED: Use OD_TRACE_DBG instead.
+    #define OD_LOG_DBG(message)                      LogManager::getSingleton().logMessage(LogMessageLevel::TRIVIAL, "", __FILE__, __LINE__, (std::string("") + message).c_str(), 0)
+
+    // DEPRECATED: Use OD_ASSERT instead.
+    #define OD_ASSERT_TRUE(condition)                if (!(condition)) LogManager::getSingleton().logMessage(LogMessageLevel::CRITICAL, "", __FILE__, __LINE__, #condition, 0)
+    // DEPRECATED: Use OD_ASSERT_MSG instead.
+    #define OD_ASSERT_TRUE_MSG(condition, message)   if (!(condition)) LogManager::getSingleton().logMessage(LogMessageLevel::CRITICAL, "", __FILE__, __LINE__, (std::string("") + message).c_str(), 0)
+#endif
+
+#define OD_TRACE_ERR(_module, _message, ...)         LogManager::getSingleton().logMessage(LogMessageLevel::CRITICAL, #_module, __FILE__, __LINE__, _message, __VA_ARGS__)
```
hwoarangmy: Good idea to add a module here. But IMHO, it would be better to use an automatic name (ie the source filename) so the logs can be copied/paste through the files (otherwise, we will always end up with wrong modules) or something more evoluted alowing to have different log levels per module (that can be changed through the console to have more output when we see a bug at runtime)

source/utils/LogManager.h
```diff
@@ -18,45 +18,64 @@
...
+    ~LogManager();
+
+    //! \brief Add a sink for log messages.
+    void addSink(const std::shared_ptr<LogSink>& sink);
 
-    //! \brief Set the log detail level.
-    //Messages lower than this level will be filtered.
-    virtual void setLogDetail(LogMessageLevel ll) = 0;
+    //! \brief Log a message to the sinks.
+    void logMessage(LogMessageLevel level, const char* module, const char* filepath, int line, const char* format, ...);
```
hwoarangmy: Here, you have changed too much the API IMHO. The basic idea was to have debug logs that can be activated when we want to test something in particular.

source/utils/LogMessageLevel.h
```diff
@@ -0,0 +1,40 @@
...
+    static const char* LevelToStringMap[] = {
+        "TRIVIAL",
+        "NORMAL",
+        "WARNING",
+        "CRITICAL"
+    };
+    return LevelToStringMap[(size_t)level];
+}
+
+#endif // _LOGMESSAGELEVEL_H_
```
hwoarangmy: Missing a new line

source/utils/LogMessageLevel.h
```diff
@@ -0,0 +1,40 @@
...
+
+inline const char* LogMessageLevelToString(LogMessageLevel level)
+{
+    static const char* LevelToStringMap[] = {
+        "TRIVIAL",
+        "NORMAL",
+        "WARNING",
+        "CRITICAL"
+    };
+    return LevelToStringMap[(size_t)level];
```
hwoarangmy: You should test that the level value is < to the array size

source/utils/LogSink.h
```diff
@@ -0,0 +1,34 @@
...
+
+class LogSink
+{
+public:
+    virtual ~LogSink() { }
+
+    virtual void write(LogMessageLevel level, const char* module, const char* timestamp, const char* filename, int line, const char* message) = 0;
+};
+
+#endif // _LOGSINK_H_
```
hwoarangmy: Missing new line

source/utils/LogSinkConsole.cpp
```diff
@@ -0,0 +1,65 @@
...
+    {
+        std::cerr << ss.str();
+    }
+    else
+    {
+        std::cout << ss.str();
+    }
+
+#if (WIN32 || _WINDOWS) && OD_DEBUG
+    ::OutputDebugStringA(ss.str().c_str());
```
hwoarangmy, line 63: I didn't knew this one. Nice ^^

source/utils/LogSinkConsole.h
```diff
@@ -0,0 +1,32 @@
...
+class LogSinkConsole : public LogSink
+{
+public:
+    LogSinkConsole();
+    ~LogSinkConsole();
+
+    virtual void write(LogMessageLevel level, const char* module, const char* timestamp, const char* filename, int line, const char* message) override;
+};
+
+#endif // _LOGSINKCONSOLE_H_
```
hwoarangmy: missing new line

source/utils/LogSinkFile.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+#include "utils/Helper.h"
+
+LogSinkFile::LogSinkFile(const std::string& filepath)
+    : mFilePath(filepath)
+{
+    mFile.open(mFilePath.c_str(), std::ios::out | std::ios::trunc);
+    if (mFile.is_open())
+    {
+        mFile.close();
```
hwoarangmy: Are you sure it is a good idea to close the file each time you write ? I'm not an expert but when I've done the file logging manager, I've opened the file in the constructor and closed in the destructor. And that because that's how it is done in Ogre. And I would think that if they do like that for such a feature, that's probably the best way to do it.
See:
https://github.com/ehsan/ogre/blob/master/OgreMain/src/OgreLog.cpp#L51

source/utils/LogSinkOgre.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+}
+
+void LogSinkOgre::write(LogMessageLevel level, const char* module, const char* timestamp, const char* filename, int line, const char* message)
+{
+#if !OGRE_THREAD_PROVIDER
+    std::lock_guard<std::mutex> lock(mLogLockMutex);
+#endif
+    mGameLog->logMessage(Ogre::String(message), toOgreLml(level));
+}
```
hwoarangmy: missing new line

CMakeLists.txt
hwoarangmy: They will automatically be added be the cpp file that include them. Look at the source list. There is no header at all there.

source/ODApplication.cpp
hwoarangmy, line 91: ok. Nice

source/utils/LogManager.cpp
hwoarangmy: You can have a look here:
https://github.com/OpenDungeons/OpenDungeons/blob/development/source/network/ODServer.cpp#L1378

source/utils/LogManager.h
knight666: I get what you were going for, but setLogDetail wasn't actually being used. I've added a mechanism to set the LogMessageLevel per module (i.e. stripped filename), is that a worthwhile replacement?

source/utils/Helper.h
knight666: I have to disagree here. I'm using snprintf to write a timestamp, while the old code created a static time locale, a stringstream, imbued the stream with the locale, used that to format a timestamp and then passed the resulting string along. This introduces several extra dynamic memory allocations in a method that should be as low overhead as possible.

source/utils/LogSinkFile.cpp
knight666: Your suspicions were correct. What's vitally important when logging to a file is that no messages are lost due to I/O operations. This can be accomplished by opening and closing the file for every message, but flushing the stream works just as well.

source/utils/LogManager.h
knight666: Implemented. ;) The module name is now the filename with the extension, so log messages can be copied between files without prejudice.

source/utils/LogManager.cpp
knight666: I have only thing to add here: doh! I should really have known better.

source/utils/LogManager.cpp
knight666: Implemented. :)

CMakeLists.txt
knight666: The headers are not added to the generated vcxproj automatically, but must be added to the source list manually. I've removed the headers from the source list (as per the guideline), but I want to make a note that it's very annoying not to have the headers available in the solution explorer when working in Visual Studio.

source/ODApplication.cpp
hwoarangmy: It is pretty widely used in our code by a retired member @oyvindln. He was quite skilled with c++ so I guess we can rely on his work. IIRC, he used something like Utils::makeUniquePtr (or something like that).
However, I don't really care about using unique_ptr or shared_ptr but I think the last is misleading. In your case, the wanted behaviour clearly is to use only one instance and a unique_ptr seems more adapted. For my part, if I see a shared_ptr, I would expect it to be used somewhere else.

source/utils/Helper.h
hwoarangmy: Yes. I think we should make a choive here ^^
I think the best way would be to have a look at how other open source projects do. I will try to have a look.

source/utils/LogManager.h
hwoarangmy: > is that a worthwhile replacement

IMHO, that would be a good reason to change the API. So, I would say yes.

hwoarangmy:
Great work. There are many improvements on the logging and the code is pretty well done. But since the new logging system breaks the non-C rule (even with arguments), I will wait for @Bertram25 to also review this one so that we all agree on the vsnprintf use.
@akien-mga feel free to give your opinion if you want.

source/utils/Helper.h
hwoarangmy: I've had a look and found how CEGUI and Ogre does. You can see it here:
https://github.com/OpenTechEngine/CEGUI/blob/master/cegui/src/DefaultLogger.cpp#L91
https://github.com/ehsan/ogre/blob/master/OgreMain/src/OgreLog.cpp#L91

I guess both are reliable and do more or less the same. IMHO, it would be wise to also do that. WDYT ?

knight666: They're functionally the same, but snprintf allows you to do it on the stack instead of on the heap. However, if the "No C" rule prevails, I'm perfectly willing to use the stringstream version instead.

hwoarangmy: When I have this kind of existential issue, I usually look at what is done in references sources. And IMHO, Ogre and CEGUI are good references. If they do the same, I would go for doing the same, too.

Bertram25:
I'll be reviewing now. :)

source/ODApplication.cpp
Bertram25: Yes, you have to use std::move to, well, move a unique ptr from a container to another. But it's feasible. Care to try?

source/utils/Helper.h
Bertram25: Yes, please, use the C++ example of CEGUI for instance. I've got nothing against C, but here it's not worth the exception.

source/utils/LogManager.h
```diff
@@ -18,45 +18,72 @@
...
 
-enum class LogMessageLevel
-{
-    TRIVIAL = 1,
-    NORMAL,
-    WARNING,
-    CRITICAL
-};
+#if OD_LOG_ALLOW_DEPRECATED
+    // DEPRECATED: Use OD_TRACE_ERR instead.
```
Bertram25: Should we really keep the old ones? Keeping both will lead to bad code practices by newcomers IMHO.

source/utils/LogMessageLevel.h
Bertram25: +1 it would be too bad to crash for a log, right? ;)

source/utils/LogSinkConsole.cpp
Bertram25, line 63: Cool, indeed. :)

Bertram25:
Arg, the overall diff doesn't display every changes. I'll have a look commit per commit.

Bertram25:
As told in the corresponding commit, yes, let's use a more C++-ish way, please, instead of vsnprintf.
As I said there, I have nothing against C, but Cegui gives a rather good enough example, for instance.

Bertram25:
@knight666 Thanks a lot for the job done, and... Welcome? ;)

knight666:
Thanks for the review! I'll remove the usage of snprintf and will upload a new version in the morning.

Thanks for having me! I was planning to do UI art, though, but I guess I got distracted. ;)

hwoarangmy:
> I was planning to do UI art, though, but I guess I got distracted. ;)

No problem. We are in a spare time project. You can follow your heart ^^

source/utils/LogManager.h
hwoarangmy: IMHO, if we go for c++ish logs as done in CEGUI and Ogre, the API will be the same between OD_LOG_XXX and OD_TRACE_XXX. The easiest way will be to replace the old one and that will avoid to change everything in code ^^

knight666: @Bertram25 Yes.The reason I want to keep them is that I wanted to prevent creating a change that affected _all_ files in the source tree. I agree with you that it's not ideal, however.

@hwoarangmy Unfortunately, the API isn't the same between the new and the old style. The old style creates an `std::string` that is passed to the `logMessage` function, while the new style uses variable arguments to create a `printf`-style logging interface.

I have written at least a dozen logging interfaces throughout my career, both using streams and using `printf`. There are clear advantages and disadvantages to both.

Streams:

```
+ Type-safety. A string will never be interpreted as a number.
+ Extensible. Easy to create a custom "operator <<" overload for a specific type, so you can just write "ss << "position: " << vec3(1.0f, 1.0, 1.0f);" for example.
- Slow. Multiple dynamic memory allocations means they're not ideal for high-performance logging.
- Cumbersome. Ever tried writing a hex decimal to a stringstream? "ss << "0x" << std::hex << std::setfill('0') << std::setw(8) << number;" Compare that to "sprintf_s(buffer, "0x%08x", number);".
- Can't be compiled out. Logging interfaces tend to become cumbersome because logging can no longer be disabled using the preprocessor.
```

Printf:

```
+ Familiar to C users. If you know printf, you know what to expect when you see "LOG("Stuff %d", thangs);". If not, you can look up the syntax.
+ Compact format. Printf syntax is like a mini-language where you can specify exactly what the output should be.
- No type-safety. Put a string in a "%d" and watch the fireworks!
- Don't try to move arguments around. "LOG("(%d) (%s:%i) %s: %s", timestamp, filename, line, message, groceries");" No, wait, the filename should be at the end...
```

The best logging system I've ever seen is part of a AAA game engine. It uses a `printf` syntax, but the arguments are _numbered_. For example:

```
LOG("What a {0} day!", m_AIMood);
```

This gives you the compactness of `printf`, the extensibility of `std::stringstream`, and type-safety! Unfortunately, in typical AAA game engine fashion, it's horribly complicated and only feasible if you can afford to have someone working on low-level things like that _full time_.

hwoarangmy: @knight666 I understand your point. In my career, I've also used C, Java, C++, C# so I'm pretty used to printf and its friends. However, as you already said, using wrong arguments or using badly printf can lead to pretty strange behaviours. And it can be hard to debug and not that easy to see during code review.
As a professional, you are used to work with other professionals who are all used to printf and know how to use it. Here, we are working on a spare time project and we some contributors are not professionals.
That's why we need to use the most straightforward (readable) code and avoid code breakers as long as we can. IMHO, the best approach for this kind of problem is to look at what other comparable (open source) project does. And, clearly, CEGUI and Ogre (2 projects pretty big that cannot be qualified of amateur - they are used in many commercial games) use the same syntax (pretty close to what we had). That's why we should go for the same IMO.

source/utils/LogMessageLevel.h
hwoarangmy: Moreover, you shouldn't use c-style cast (build would fail with Linux). You should use c++ syntax:

```
static_cast<size_t>(level)
```

hwoarangmy:
@knight666 Note that the CI (appVeyor and Travis) failed with your build. It seems like MinGW32 do not like your mutex. The easiest for that is to use SFML (which is already included). You can have a look at utils/StackTraceMinGW.cpp to see how to use it.
Moreover, old C-style casts are not allowed. That's why Travis CI fails.

knight666:
@hwoarangmy Saw it, already checking if my fix compiles. ;) Do you have continuous integration for any other targets besides Windows?

Also, funny how these aren't build errors, even though they're far more dangerous:

```
2>e:\source\opendungeons-fork\source\entities\creature.cpp(2451): warning C4244: 'argument' : conversion from 'int' to 'const Ogre::Real', possible loss of data
2>e:\source\opendungeons-fork\source\entities\creature.cpp(4891): warning C4244: '+=' : conversion from 'double' to 'Ogre::Real', possible loss of data
2>e:\source\opendungeons-fork\source\entities\creature.cpp(4894): warning C4244: '+=' : conversion from 'double' to 'Ogre::Real', possible loss of data
2>e:\source\opendungeons-fork\source\entities\creature.cpp(4897): warning C4244: '+=' : conversion from 'double' to 'Ogre::Real', possible loss of data
2>e:\source\opendungeons-fork\source\gamemap\pathfinding.h(29): warning C4244: 'argument' : conversion from 'int' to 'float', possible loss of data
2>          e:\source\opendungeons-fork\source\entities\creature.cpp(2520) : see reference to function template instantiation 'float Pathfinding::distanceTile<Tile>(const T &,const T &)' being compiled
2>          with
2>          [
2>              T=Tile
2>          ]
```

source/utils/LogManager.h
knight666: So... do you want me to remove the new macros? Because that's fine, it'll just be a bit more work.

hwoarangmy:
> Do you have continuous integration for any other targets besides Windows?

Yes. appVeyor checks windows with MinGW32 and MSVS2013 (I will probably do the same for MinGW64 as it is what I mostly use)
Travis checks for Linux build with gcc 4.8. Note that Travis could also check compilation for OSX but because we have never tried yet if it works (as none of us have OSX), we didn't test.

> Also, funny how these aren't build errors, even though they're far more dangerous

Yes. That's because theses warnings are triggered by MSVS and not gcc or MinGW by default. And since there are a lot of them, I was too lazy to fix it.

Bertram25:
> Also, funny how these aren't build errors, even though they're far more dangerous:

Note also that strict conversion check is a warning type enabled in MSVS, but not a dangerous error per se. Those can be fixed but with a lot of pain for almost no value.
It's up to you to fix those, but as long as gcc4/5 don't display more fatal warnings or bad practices, I'm fine personally. :)

source/utils/LogManager.h
```diff
@@ -18,45 +18,55 @@
 #ifndef LOGMANAGER_H
 #define LOGMANAGER_H
 
+#include <memory>
```
hwoarangmy, line 21: Are theses needed ? Moreover, in our coding guide, standard headers should be included at last. The order should be:
1 - Class header (for cpp files)
2 - OD headers
3 - Library headers (ogre, cegui, sfml, ...)
4 - standard headers

This do not prevent merge but if you want to go for something else, it's good to know ^^

source/utils/LogMessageLevel.h
```diff
@@ -0,0 +1,45 @@
...
+*  but WITHOUT ANY WARRANTY; without even the implied warranty of
+*  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+*  GNU General Public License for more details.
+*
+*  You should have received a copy of the GNU General Public License
+*  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#ifndef _LOGMESSAGELEVEL_H_
+#define _LOGMESSAGELEVEL_H_
```
hwoarangmy, line 19: For headers, we do not use _ at begin and start of the word. Not blocking

knight666, line 19: Sorry about that, that's just how my macro is set up. ;)

hwoarangmy, line 19: no pb. It's just for the next one ;-)

hwoarangmy:
I will try that tonight

hwoarangmy:
@knight666 Thanks for your work. I've tested and it works well. Note that for critical messages, classname and line number are displayed before the line which seems weird (I would expect that after time + log lovel). However, it is already a nice improvement so I will open an issue. If you don't want to handle that, I will. Just tell me in the issue if possible.

hwoarangmy:
@knight666 Its open: #998

knight666:
@hwoarangmy Thank you for being so strict in your reviews. It helps newcomers like me to get acquainted with a project. ;)

Now that the logging refactor has been merged, it's time to tackle the assert handling... _evil laughter_

hwoarangmy:
> It helps newcomers like me to get acquainted with a project. ;)

I'm such an old bear... Only @Bertram25 can be worst than me ;-)

> Now that the logging refactor has been merged, it's time to tackle the assert handling

It looks promizing. Can't wait to see it ingame ^^

akien-mga:
> I'm such an old bear... Only @Bertram25 can be worst than me ;-)

Just wait for my nit-picking about trailing spaces, indentation, spelling or line endings :p

hwoarangmy:
> Just wait for my nit-picking about trailing spaces, indentation, spelling or line endings :p

That's right, we are such an old bear team ^^

PR #995, opened 2015-10-20
Add "_d" suffix to game executable on Windows.

knight666:
One thing that bugged me while working in Visual Studio was that while the output directory had direct links to the media folders (which is great!), the game's executable would be output to "Debug" and "Release" folders. This meant I had to copy the required DLLs to both Debug and Release folders.

My fix is to always output the executable to the same build directory, but differentiate between debug and release builds using a "_d" suffix.

I copied this change from CEGUI, take that as you will. ;)


CMakeLists.txt
```diff
@@ -508,6 +508,23 @@ endif()
 #### Binary ######################
 ##################################
 
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}")
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY_DEBUG ${CMAKE_RUNTIME_OUTPUT_DIRECTORY})
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY_RELEASE ${CMAKE_RUNTIME_OUTPUT_DIRECTORY})
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY_MINSIZEREL ${CMAKE_RUNTIME_OUTPUT_DIRECTORY})
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY_RELWITHDEBINFO ${CMAKE_RUNTIME_OUTPUT_DIRECTORY})
+
+if(WIN32 OR APPLE)
```
hwoarangmy, line 517: Do you have an apple computer ? That would be interesting as none of us have

hwoarangmy:
looks good. I will try that tonight

CMakeLists.txt
knight666, line 517: Unfortunately, I do not. This line was copied from CEGUI.

CMakeLists.txt
```diff
@@ -508,6 +508,23 @@ endif()
...
+    set(OD_BUILD_SUFFIX "_d" CACHE STRING "String holding a suffix appended to the name of output binaries (under CMake build, only used for debug).")
+else()
+    set(OD_BUILD_SUFFIX "" CACHE STRING "String holding a suffix appended to the name of output binaries (under CMake build, only used for debug).")
+endif()
+
+if(OD_BUILD_SUFFIX)
+    set(OD_HAS_BUILD_SUFFIX TRUE)
+    set(CMAKE_DEBUG_POSTFIX ${OD_BUILD_SUFFIX})
```
hwoarangmy, line 525: Is CMAKE_DEBUG_POSTFIX used ?

CMakeLists.txt
```diff
@@ -508,6 +508,23 @@ endif()
...
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY_RELWITHDEBINFO ${CMAKE_RUNTIME_OUTPUT_DIRECTORY})
+
+if(WIN32 OR APPLE)
+    set(OD_BUILD_SUFFIX "_d" CACHE STRING "String holding a suffix appended to the name of output binaries (under CMake build, only used for debug).")
+else()
+    set(OD_BUILD_SUFFIX "" CACHE STRING "String holding a suffix appended to the name of output binaries (under CMake build, only used for debug).")
+endif()
+
+if(OD_BUILD_SUFFIX)
+    set(OD_HAS_BUILD_SUFFIX TRUE)
```
hwoarangmy, line 524: FMI, why do you use a boolean here after testing OD_BUILD_SUFFIX ? Why don't you test OD_BUILD_SUFFIX later instead of the boolean ?

CMakeLists.txt
knight666, line 525: Yes, that's the variable used by CMake for actually setting the suffix.

hwoarangmy:
> Do we really need to redefine CMAKE_RUNTIME_OUTPUT_DIRECTORY*

I've faced the same problem for another open source project. By default, MSVS outputs binaries to "PROJECT/BUILD_TYPE". For example, OD/Release/OpenDungeons.exe.

And if you want to force the binary output directory to be in OD, you need to set it for every build type (there is no way to do that once and for all).

hwoarangmy:
Good to merge for me. Nice work

PR #997, opened 2015-10-21
Fixed Travis compilation + small improvements

hwoarangmy:


Bertram25:
Excellent to see CI integration live here. :)

hwoarangmy:
> Excellent to see CI integration live here. :)

Yes. And once we have decided how to do tests, I guess we will have done a great step to stability ^^

Bertram25:
Ah, btw: http://docs.travis-ci.com/user/migrating-from-legacy/?utm_source=legacy-notice&utm_medium=banner&utm_campaign=legacy-upgrade

hwoarangmy:
> Ah, btw: http://docs.travis-ci.com/user/migrating-from-legacy/?utm_source=legacy-notice&utm_medium=banner&utm_campaign=legacy-upgrade

I had seen that but honestly, I don't see how we can do anything without using sudo. The white repository list is really poor, only gcc 4.6 is installed (OD do not compile with it as it do not handle c++11 stuff), cmake also do not handle all the commands inside our makefile...
Maybe later, when they will get a default environment that is worth that name

Bertram25:
Ah, I realize I wasn't that explanatory. I completely agree, their default environment is outdated to the least. I just wanted to mention their migration, but it means no additional work on (y)our side, of course. :)

akien-mga:
Builds and works fine, and @Bertram25 seems happy with the code changes so let's merge :)

Currently the gold is displayed "2437 / 5000" while the creatures number is "4/10", we might want to homogenise this :) (either spaces around the slash like for gold, or none like for creatures).

hwoarangmy:
I would say none as it takesless space (and it may be pretty high for gold) ^^

Bertram25:
Yup, I'm fine with no spaces before and after '/'

PR #1003, opened 2015-10-30
Proposal for the overcrowded rooms tab

akien-mga:
Since the addition of the bridges, there were too many icons in the rooms tab to fit in the window (especially in the editor at small resolutions), so @hwoarangmy reduced the size of the icons and put them on two lines.

I'm not too fond of the inconsistency that creates between the rooms tabs and the other ones though, and in higher resolutions there's a lot of lost space on the right for a very crowded panel on the left.

So here's a work-in-progress proposal to make the tabs scrollable panes by default. I find it works relatively well in the editor (test it e.g. in 1024x768 to be sure that is will need more space than available). On the other hand in-game the horizontal scrollbar is still visible due to the presence of hidden icons, and that looks somewhat messy.


hwoarangmy:
I've tried the PR. Indeed, it seems pretty weird to have the scrollbars always visible (even at game start when you only have a few rooms). It might be a good idea to try to resize the pane according to the icons.

Bertram25:
Sorry guys, It took me some time to kill a nasty bug, but I'll be around a bit more the next few days. :)

As for the PR and the related feature, I'll test it asap, but I wondered a few things:
- Since we're adding new rooms/traps/spells depending on research, would it be good to start adding button on the fly in the corresponding tabs. (A bit like we do for the settings video options but horizontally.)
  Doing this would sort the scroll bar problem, IMHO, and corresponds more to the player logic.
- If so, would we want to sort the button when updating them? Keeping the same logic whatever the research order is?

WDYT?

hwoarangmy:
@Bertram25 Nice to see you back ^^

> Since we're adding new rooms/traps/spells depending on research, would it be good to start adding button on the fly in the corresponding tabs.

As a player, I would expect the icons to always be at the same place, no matter which researches are done or not. But if we follow this logic, we might end up with the room tab with a few rooms on the left, nothing in the middle and some more on the right (and they might be not visible without moving the scrollbar.
Note that we could also disable not available rooms/traps instead of hiding them. That would allow to have the same display whatever is enabled or not (and maybe adding the editor specific rooms to some specific tab).
However, using a scrollbar doesn't sound very user friendly. Using 2 tabs for basic/advanced rooms might be a solution...

Bertram25:
> Note that we could also disable not available rooms/traps instead of hiding them. That would allow to have the same display whatever is enabled or not (and maybe adding the editor specific rooms to some specific tab).

Yep, also feasible. Both are fine to me.

> However, using a scrollbar doesn't sound very user friendly. Using 2 tabs for basic/advanced rooms might be a solution...

Also a good solution. What would you see in basic and advanced rooms for a start?

Bertram25:
> @Bertram25 Nice to see you back ^^

@hwoarangmy Thanks :) The latest commit rock BTW. :D This is really going to be something.

hwoarangmy:
> Also a good solution. What would you see in basic and advanced rooms for a start?

Well, for now, I would split editor specific rooms. And I would say all rooms we have are basic except prison and crypt which are advanced.

@OpenDungeons/developers WDYT ?

akien-mga:
> Well, for now, I would split editor specific rooms. And I would say all rooms we have are basic except prison and crypt which are advanced.

Sounds good to me. Actually if we move the portal and temple icons out of the main tab I think it will have enough space for the rest in 1024x768. If we want to keep supporting lower resolutions without changing the UI too much, we can then keep this scrollable pane solution as a fallback (the scrollbar would only show when potentially needed, i.e. at very low resolutions).

Another solution could be to do a complete redesign of the bottom UI, using smaller icons for all tabs (what motivated my change was mainly the fact that having different icon sizes depending on the tabs look weird).

Using smaller icons means that we would have plenty of free space to allocate for different stuff; we might want to have a look at what similar games do in terms of UI to get ideas. For example we could make a distinction between an "action" panel on the right (spells, creature management) and a "building" panel (rooms, traps) on the left. Just some random thoughts, more for a long-term UI evolution.

hwoarangmy:
Yes, thinking about the UI in a more global way may be a good idea. Maybe @knight666 will also have some ideas here.

Bertram25:
Is this one still desired?

akien-mga:
Nope. We still need to find a solution for the UI before the release though.

Bertram25:
Ah? I thought the solution chosen was to split the rooms tabs into basic and advanced?

hwoarangmy:
> Ah? I thought the solution chosen was to split the rooms tabs into basic and advanced?

I guess that's the best idea.

PR #1006, opened 2015-11-07
Fix icon size of new spells

akien-mga:
Cherry-picked from #1003 as the latter is not really meant to be merged, only a test for scrollable panes.

### DIFF
--- a/gui/WindowResearchTree.layout
+++ b/gui/WindowResearchTree.layout
@@ -379,7 +379,7 @@
                     </Window>
                 </Window>
                 <Window type="OD/ImageButton" name="CreatureStrengthButton" >
-                    <Property name="Area" value="{{0.7,-25},{0,195},{0.7,25},{0,255}}" />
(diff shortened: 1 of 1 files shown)


akien-mga:
I notice that appveyor and travis-ci go wild even though I only touched an XML file in gui. Is there a way to configure it to test rebuilding everything only when a file that needs to be compiled is modified?

It doesn't hurt much, but it's not dungeon-environmentally-friendly :P

hwoarangmy:
> I notice that appveyor and travis-ci go wild even though I only touched an XML file in gui

AFAIK, they compile as soon as there is a new PR. Note that xml files might also describe some source architecture. That means that checking for an xml file is not enough to know if the sources should be compiled or not. For my part, I prefer the they check too much instead of not enough ^^

akien-mga:
> For my part, I prefer the they check too much instead of not enough ^^

Fair enough :)

Issue #578, opened 2015-03-24
Improve test coverage


oyvindln:
Right now, only the consoleinterface class has tests. We should be aiming for having as much as possible covered by tests, but it will require some redesign. TileContainer seems to be a good place to start.

hwoarangmy:
Well, I've already thought about that and even tried to add tests to TileContainer but there are so many dependencies that you need to include the whole project. Basically, because TileContainer uses Tile. And Tile uses GameMap. And from one to the other, we end up including everything. That's the bad point of not using abstract layers ^^

hwoarangmy:
I've been thinking a bit more about unit tests. IMHO, our architecture is not adapted to unit tests as @oyvindln tried to implement. Almost every interesting class relies on GameMap and compiling a test means more or less to compile the whole game.
Except testing very simple things, like checking if our Random class works, I guess we cannot do much this way. And certainly not the real interesting tests we might want to do (tests on floodfill, pathfinding, creature behaviour, ...)

That being said, our client-server architecture is very test friendly to test independantly the client and the server (by stubbing one or the other - just like the replay mode does).

My proposition is:
- Merge again the files splitted by @oyvindln so that we have 1 cpp per header file (I think splitting makes easy to miss something when dupplicating code)
- Do not have 1 executable per test. But find a way to start OD in a special mode or something similar to run unit tests. That might be by using scripts (AS? LUA?) or by loading a dll (to avoid embedding unit tests within OD).

@oyvindln your opinion would be welcome if you are still around.
@OpenDungeons/developers WDYT ?

hwoarangmy:
I've been thinking that the most straightforward way to tests would be use the normal executable, launch it on server mode and go for a client stub to make tests (using command consoles for tests needs for exemple for wrapping creatures and this kind of stuff). That would allow to test the server easily and very effectively (since we test the real executable).
However, testing the client side might be a little bit more tricky. But it may be possible by grepping some stuff in the logs to see if it behaves as expected.

@Bertram25 WDYT about merging the sources back as I've proposed ? Any other idea for testing ?

Bertram25:
> Merge again the files splitted by @oyvindln so that we have 1 cpp per header file (I think splitting makes easy to miss something when dupplicating code)

I must say I completely agree. Duplicating cpp file vs 1 header, except for good reason, is considered a bad practice AFAIK.
[...]

PR #1008, opened 2015-11-08
Enabled tests in appVeyor + added a test that ensures that game can be launched

hwoarangmy:
part of #578


akien-mga:
I get a linking issue with boosttest-source_tests-Goal:

```
Linking CXX executable ../../boosttest-source_tests-Goal
CMakeFiles/boosttest-source_tests-Goal.dir/__/utils/LogManager.cpp.o: In function `LogManager::logMessage(LogMessageLevel, char const*, int, std::string const&)':
/home/akien/Projects/opendungeons/OpenDungeons/source/utils/LogManager.cpp:60: undefined reference to `boost::filesystem::path::stem() const'
/home/akien/Projects/opendungeons/OpenDungeons/source/utils/LogManager.cpp:83: undefined reference to `boost::filesystem::path::filename() const'
CMakeFiles/boosttest-source_tests-Goal.dir/__/utils/LogManager.cpp.o: In function `__static_initialization_and_destruction_0':
/usr/include/boost/system/error_code.hpp:222: undefined reference to `boost::system::generic_category()'
/usr/include/boost/system/error_code.hpp:223: undefined reference to `boost::system::generic_category()'
/usr/include/boost/system/error_code.hpp:224: undefined reference to `boost::system::system_category()'
collect2: error: ld returned 1 exit status
source/tests/CMakeFiles/boosttest-source_tests-Goal.dir/build.make:164: recipe for target 'boosttest-source_tests-Goal' failed
```

Same thing for the LaunchGame test with missing references to boost_system and boost_filesystem.

akien-mga:
It builds fine with:

```
diff --git a/source/tests/CMakeLists.txt b/source/tests/CMakeLists.txt
index 6bacb78..d3a709e 100644
--- a/source/tests/CMakeLists.txt
+++ b/source/tests/CMakeLists.txt
@@ -30,6 +30,7 @@ add_boost_test(Goal
         "${SRC}/utils/LogManager.cpp"
         "${SRC}/utils/LogSinkConsole.cpp"
         LIBRARIES
+        ${Boost_LIBRARIES}
         ${SFML_LIBRARIES})

 add_boost_test(Pathfinding
@@ -50,4 +51,5 @@ add_boost_test(LaunchGame
         ${SRC}/utils/LogSinkConsole.cpp
         test_LaunchGame.cpp
         LIBRARIES
-        ${SFML_LIBRARIES})
\ No newline at end of file
+        ${Boost_LIBRARIES}
+        ${SFML_LIBRARIES})
```

I don't know though if linking against the whole pack of Boost_LIBRARIES leads to overlinking or if it's fine this way.

hwoarangmy:
> I don't know though if linking against the whole pack of Boost_LIBRARIES leads to overlinking or if it's fine this way.

That's probably a bug introduced by the change in logging logic from @knight666 patch. We can always link against the needed libraries only. I think the whole thing will be fixed when we will enable unit test on appVeyor + Travis. I will have a look.

> Could you name the file "TestUnitTest.level" to match the other test levels?

Well, I had something else in mind. First, when unit test will work, I thought about using a special folder as we don't want unit test maps to be changed (and a new program option to launch them from the good folder).
Secondly, I thought about coding in the test name the testing process (for example, if it requires to launch a server with the specific test map) so that we can launch a script that would list all boosttest-source_tests-\* and will do required stuff. This way, adding a new unit test would not require to change the tests scripts (because it would be easy when testing on windows, to not update the linux script and an error would not be visible)

akien-mga:
> Well, I had something else in mind. First, when unit test will work, I thought about using a special folder as we don't want unit test maps to be changed (and a new program option to launch them from the good folder).

Sounds good yes.

> Secondly, I thought about coding in the test name the testing process (for example, if it requires to launch a server with the specific test map) so that we can launch a script that would list all boosttest-source_tests-\* and will do required stuff. This way, adding a new unit test would not require to change the tests scripts (because it would be easy when testing on windows, to not update the linux script and an error would not be visible)

That would indeed prevent us from having to maintain concurrent testing scripts for all platforms, etc.. In the meantime, I've made a linux shell script that matches your batch script:

```
#!/bin/sh
# Shell script to run the unit tests for OpenDungeons

OD_BINARY="opendungeons"
UNIT_TESTS="ConsoleInterface Random Goal ODPacket Pathfinding LaunchGame"

if [ ! -x $(pwd)/${OD_BINARY} ]; then
    echo "Can't find the ${OD_BINARY} binary in the current directory, aborting."
    exit 0
fi

TESTS_PASSED=""
TESTS_FAILED=""

for test in ${UNIT_TESTS}; do
    echo -e "\n### Unit test: ${test}\n"

    if [ ${test} == "LaunchGame" ]; then
        ./${OD_BINARY} --server UnitTest.level --port 32222 --log srvLog.txt &
    fi
    ./boosttest-source_tests-${test}

    if [ $? == 0 ]; then
        TESTS_PASSED+=" ${test}"
    else
        TESTS_FAILED+=" ${test}"
    fi
done

echo -e "\n#####################################"
echo "The following tests were PASSED: ${TESTS_PASSED}"
echo "The following tests were FAILED: ${TESTS_FAILED}"
echo -e "#####################################\n"
```

(If it looks good to you I'll do a PR once yours is merged).

hwoarangmy:
> I've made a linux shell script that matches your batch script

Cool ^^

Like in yours, I first thought about running all tests so that we know which are ok or not. However, the problem is that if one fails, we are not sure the others can be run. That's because we launch OD in server mode. And since a failed test might not launch the game (and stop the server once it is done), it might make a test launch the wrong test map or something similar. That's why I decided to stop at the first failing test (so that it is clear which one first failed as it might be the root to another failing test).

hwoarangmy:
@akien-mga I've enabled tests on travis (mostly because I needed to trigger a build because the error I had was because of a missing library in the win32 dependencies). If everything goes well with appVeyor, could you please handle the travis part (adding the still missing boost libraries + your script) ?

Thanks

hwoarangmy:
@akien-mga Everything goes well on appVeyor. If you want to handle the travis part, please merge and play with the scripts ^^
If not, let me know and I will do that here.

akien-mga:
Sure, I'll have a go at it.

PR #1009, opened 2015-11-08
Enable unit tests on Linux in travis

akien-mga:


hwoarangmy:
Thanks for taking care of that.
I've had a look at the logs but it seems there are some errors. Like
https://travis-ci.org/OpenDungeons/OpenDungeons/builds/89947751#L2423

Moreover, the tests are not launched. We need the script to return a value != 0 when there is an error so that we know when the tests fail.

akien-mga:
Hm it looks like the tests were properly launched, see all "Running 1 test case..." and "**\* No errors detected" mentions. There are errors due to Ubuntu having a weird shell implementation, I'll have another look to fix my script.

The script checks the return value of the last command (`$?`) to see if the test succeeded. But right now it's the string management afterwards that seems not to work on Ubuntu shell.

hwoarangmy:
> it looks like the tests were properly launched

Yes. I would guess that it's the string comparaison that fails (and that would probably explain why the server is not launched and, thus, the failing of launch game). However, we would expect the tests to fail when there is an error in any case. I will have a look in my ubuntu VM.

akien-mga:
@hwoarangmy Well there wasn't an error as far as the scripts were concerned, only in the cosmetic aspects of my shell script (i.e. listing all FAILED and PASSED tests in the end). The only test that failed was indeed the LaunchGame one since the broken string comparison prevented starting the server.

It should hopefully be fixed now as I've switched the shebang to enforce bash over Ubuntu's funky dash. At least it works fine in my Ubuntu 12.04 LTS VM.

hwoarangmy:
> It should hopefully be fixed now as I've switched the shebang to enforce bash over Ubuntu's funky dash

Cool. One last thing, could you make sure your script returns != 0 when a test fails ? The idea is to make travis fail when there is a problem to make sure we only merge stuff that runs through the tests ^^

akien-mga:
> One last thing, could you make sure your script returns != 0 when a test fails ? The idea is to make travis fail when there is a problem to make sure we only merge stuff that runs through the tests ^^

Yes good idea. I understand what you said earlier about the fact that one failed test might mean that the subsequent test results might not be so relevant, but since they run pretty fast, what do you think about only returning != 0 after all tests after been done, if `TEST_FAILED!=""`? It might still be useful to see what happens to tests 2 to N after test 1 has failed.

akien-mga:
It _should_ now work, but we'll have to make a test fail to confirm :-P

We could likely do that in a dummy PR once this one has been merged.

hwoarangmy:
> what do you think about only returning != 0 after all tests after been done, if TEST_FAILED!=""?

No problem

> It should now work, but we'll have to make a test fail to confirm :-P

If you have something to do in mind, you can have a go. For my part, I will build a new test that will check that allied creatures do not fight and that they engage enemy. If you have not done it before, I will make a test fail ^^

akien-mga:
> If you have something to do in mind, you can have a go. For my part, I will build a new test that will check that allied creatures do not fight and that they engage enemy. If you have not done it before, I will make a test fail ^^

Well I was just thinking to comment out the command that starts the server to make the LaunchGame test fail :) But I want to do this in a separate PR to avoid polluting this one with silly commits :P

hwoarangmy:
ok. Then I will let you do it as your instinct tells ^^

PR #1012, opened 2015-11-09
Make a test fail to see how travis behaves

akien-mga:
Mwahaha.

BTW @hwoarangmy, I'm not sure if scripts/ is the best place for those scripts, it seems that it was more meant for game-relevant scripts (though we don't use any so far). So when we start using in-game scripts we might want to move those CI scripts to tools/ for example.

### DIFF
--- a/scripts/unix/run_unit_tests.sh
+++ b/scripts/unix/run_unit_tests.sh
@@ -15,9 +15,9 @@ TESTS_FAILED=""
 for test in ${UNIT_TESTS}; do
     echo -e "\n### Unit test: ${test}\n"
 
-    if [ "${test}" == "LaunchGame" ]; then
-        ./${OD_BINARY} --server UnitTest.level --port 32222 --log srvLog.txt &
-    fi
+    # if [ "${test}" == "LaunchGame" ]; then
+    #     ./${OD_BINARY} --server UnitTest.level --port 32222 --log srvLog.txt &
+    # fi
     ./boosttest-source_tests-${test}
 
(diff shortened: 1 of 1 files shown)


hwoarangmy:
> BTW @hwoarangmy, I'm not sure if scripts/ is the best place for those scripts, it seems that it was more meant for game-relevant scripts

Well, if we really use scripts for the levels, IMHO, the most relevant place would be in the same folder (maybe with the same level name) as I don't think we will want to share the same script for several levels.

> we might want to move those CI scripts to tools/ for example

Doesn't sound like a bad place either. If you prefer to move the scripts, I've nothing against it.

akien-mga:
The travis-ci build really takes a while, I'll see what I can do to make us use dependencies from GetDeb to prevent having to compile the whole cegui ourselves.

_Edit:_ Though I guess travis-ci uses Ubuntu 12.04 LTS or similar? Not sure if GetDeb has packaged cegui 0.8.x for this old release. If not we could maybe also setup a PPA with the dependencies we need, and/or provide a zipfile with pre-compiled dependencies as was also suggested.

akien-mga:
> Well, if we really use scripts for the levels, IMHO, the most relevant place would be in the same folder (maybe with the same level name) as I don't think we will want to share the same script for several levels.

Yeah, for the time being as we're not using scripts at all it's a bit early to decide where they would be best stored. Time will tell :)

akien-mga:
So, it seems to fail fine :)

PR #1013, opened 2015-11-09
Improved test coverage

### LINKED ISSUE #578: Improve test coverage

hwoarangmy:
fixes #578


hwoarangmy:
@akien-mga Could you please update the linux script according to the batch ?

akien-mga:
> Could you please update the linux script according to the batch ?

Sure, btw the change is pretty clever :)

hwoarangmy:
thanks ^^

hwoarangmy:
@akien-mga Note that a test failed in travis but the test is considered ok (I forgot to change the level name in the script)

akien-mga:
>  Note that a test failed in travis but the test is considered ok (I forgot to change the level name in the script)

Ah right, actually that shows one good and one bad thing:
- The good: If the test name is wrong, it is properly considered as a failed test in the "$?" check, so that's cool.
- The bad: I'm a noob at bash and forgot some quotes, this will fix the issue that prevents having "exit 2" after more than one failed tests:

```
diff --git a/scripts/unix/run_unit_tests.sh b/scripts/unix/run_unit_tests.sh
index 614a24d..c954331 100755
--- a/scripts/unix/run_unit_tests.sh
+++ b/scripts/unix/run_unit_tests.sh
@@ -32,6 +32,6 @@ echo "The following tests were PASSED: ${TESTS_PASSED}"
 echo "The following tests were FAILED: ${TESTS_FAILED}"
 echo -e "#####################################\n"

-if [ ! -z ${TESTS_FAILED} ]; then
+if [ ! -z "${TESTS_FAILED}" ]; then
     exit 2
 fi
```

hwoarangmy:
> this will fix the issue that prevents having "exit 2" after more than one failed tests

Seems like I also forgot to rename a test ^^ Can you handle that once this one is merged ?

akien-mga:
> Seems like I also forgot to rename a test ^^ Can you handle that once this one is merged ?

I won't need it as I'll use the same logic as you to determine test names dynamically :)

akien-mga:
@hwoarangmy I don't know how batch scripts work on Windows, but you might have to check if you won't get issues when more than one test wants a server. For Linux I made a system to store the PID after launching the server and kill it after the test was run, to prevent trying to start a server on an already used port due to the previous test.

hwoarangmy:
> For Linux I made a system to store the PID after launching the server and kill it after the test was run, to prevent trying to start a server on an already used port due to the previous test.

That's a good idea but I'm not sure it is feasible on windows. I'll have a look

akien-mga:
Looks fine to merge, and I want to push my updated linux script, so let's go :)

PR #1014, opened 2015-11-09
Improve BASH unit test script to remove hardcoded test names

akien-mga:
The tests are now named with a $basename-LL-$test convention where LL is the name of the level that should be run in the server, if any (00 if none).
Also fixed final check to know if some tests failed that would not work on strings with spaces (i.e. more than one failed test).

### DIFF
--- a/scripts/unix/run_unit_tests.sh
+++ b/scripts/unix/run_unit_tests.sh
@@ -2,7 +2,7 @@
 # Shell script to run the unit tests for OpenDungeons
 
 OD_BINARY="opendungeons"
-UNIT_TESTS="00-ConsoleInterface 00-Random Goal 00-ODPacket 00-Pathfinding aa-LaunchGame"
+TEST_BASENAME="boosttest-source_tests-"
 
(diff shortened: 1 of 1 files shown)


scripts/unix/run_unit_tests.sh
```diff
@@ -12,19 +12,32 @@ fi
 TESTS_PASSED=""
 TESTS_FAILED=""
 
-for test in ${UNIT_TESTS}; do
+# The boost test filenames are supposed to be as follows: ${TEST_BASENAME}LL-*
+# LL=name of the level server the game should launch (00 if none).
+for testbin in boosttest-source_tests-*; do
```
hwoarangmy, line 17: Why didn't you use TEST_BASENAME here ?

hwoarangmy:
perfect ^^

scripts/unix/run_unit_tests.sh
akien-mga, line 17: I forgot :D

PR #1015, opened 2015-11-10
Added unit tests for creatures and rooms

hwoarangmy:


akien-mga:
In the travis-ci log I get:

```
./scripts/unix/run_unit_tests.sh: line 38: kill: (32027) - No such process
```

So I guess my trick to make sure to terminate the server process after the test was run is not necessary, having the tests leave the server seem to be enough to stop it?

hwoarangmy:
> So I guess my trick to make sure to terminate the server process after the test was run is not necessary, having the tests leave the server seem to be enough to stop it?

It should be. The sleeps would be better in the script but I couldn't find a way to use "timeout" or "ping" so I moved the sleeps there.

akien-mga:
Looks fine.

PR #1016, opened 2015-11-12
Dark elf + new Prison texture

Danimal696:
Still missing normal elf and new ranged weapons


Bertram25:
Whoohoo, can't wait to see him in game. :D

hwoarangmy:
Cool. I will try that tonight ^^

@Danimal696 Note that you can gather all materials in one file to avoid having several files for the same model

hwoarangmy:
works well

Danimal696:
size it 0.11 for better results, the walk and attack speed also needs to be adjusted

hwoarangmy:
Note that there will be a problem of synchronization when we will have the proper missile. The architecture makes that the missile is thrown at animation start which is not exactly what will be expected from what I've seen in the animation.

Danimal696:
i can cut the animation to jump at shoot point from the start, should i?

hwoarangmy:
I think it would be better. To say that they have a weapon ready for firing and the warmup would be used to reload.

Danimal696:
okay

PR #1026, opened 2015-11-27
Implemented master server + few technical improvements

hwoarangmy:
Note that I made a mistake and pushed changes to the master server in the main repo (I thought I had forked but I had not).
Anyway, I've created a new account for OD in OpenShift (as @alketii did but I didn't had the logins). The master server can be tested and should already work. It can be used for games launched in server mode or with UI.


alketii:
Awesome! Will test tomorrow.

Bertram25:
Whoohoo, will test tonight. (Btw, will start working on user map support soon.)

akien-mga:
I'm testing now :)

Just FYI, it's not related to this PR, but I'm quite active in the development of Godot Engine lately (https://github.com/godotengine/godot) in the bug triage and to some extent project management, so my free dev time is well filled :) But I'm still very much interested in OD of course, it was just to explain why I'm now testing PRs more in 3-4 days instead of 12h some time ago :p

akien-mga:
IT LIVES!
odms

I didn't even have to configure anything port-wise; but then I haven't tested yet if external clients can connect to my game :D

_Edit:_ Well without configuring port forwarding I couldn't connect to my own server, but I've finally managed to configure it ;)

alketii:
Finally I can play OD :')

hwoarangmy:
> Edit: Well without configuring port forwarding I couldn't connect to my own server, but I've finally managed to configure it ;)

The master server do not change anything about needed network configuration. And sometimes, it is not very straight forward.

> Finally I can play OD :')

What was the problem before ? From the network point of view, it changes nothing. The only difference is that we won't have to tell each other the IP of the server. It will be done automatically. But apart from that, there is no differences.

akien-mga:
It works great so far. Nice work!

Some minor issues:
- When trying to connect to a server running "Ruins of the Confluent", the first letter of the description is missing. Same with "Scream in the dark...", and both descriptions start with a "A".
- Trying to join a server using the same nickname as the creator obviously does not work, but it also does not error out cleanly. I get this in the logs `(MenuModeMasterServerJoin) [CRITICAL] (MenuModeMasterServerJoin.cpp:126) index too high=0, size=0` and it seems to create new columns in the server join window (i.e. new pairs of creator/level, but empty).
- I had once an issue when starting the lobby for a server, then going back and starting a new level. The client with which I had listed available servers after starting the first level would still see the first level. I'm not 100% sure how to reproduce yet, but we'll iron out things little by little, the current status is already mega awesome.

akien-mga:
> > Finally I can play OD :')
> 
> What was the problem before ?

I think none, but @alketii is lazy :D

akien-mga:
Good to merge for me :)
Now we need a multiplayer game to test the master server.

akien-mga:
I just hacked myself into @hwoarangmy's test server and therefore found another issue:
- If the server disconnects while clients are in the lobby, they are not notified about it in the GUI (there is however `(ODSocketClient) [WARNING] (ODSocketClient.cpp:180) Socket disconnected` in the console). The "creator player" is shown as having left the lobby though.

Another small issue:
- Timestamps in the masterserver seem to be UTC-5. It would be nice to use UTC IMO.

hwoarangmy:
> Trying to join a server using the same nickname as the creator obviously does not work, but it also does not error out cleanly

Note that it doesn't work in local too. That's not related to this PR, using the same nick for several players never worked. I've never looked why, though. Might be worth opening an issue to not forget.

> it seems to create new columns in the server join window

I've just tested and seen each time we close/enter join menu, columns are re-added. I will fix that.

hwoarangmy:
I've fixed the columns and the chars not being displayed. concerning using UTC time, I propose to try the master server and see if there is more to fix.

akien-mga:
I'll merge, but if @Bertram25 still wants to try it tonight, we can try to set up a multiplayer game via the masterserver, I should be around.

hwoarangmy:
@akien-mga Since you want to have a game tonight, I'm uploading win32 binaries. If you want to go for similar linux binaries, I'm based on the main repo.

PR #1036, opened 2015-11-29
Fixed segfault about hatchery absorbtion

hwoarangmy:


akien-mga:
Travis failed but it's apparently a SF issue when downloading boost. Still need to take the time to find relevant Ubuntu repos to add so that we don't need to build all deps ourselves at each test.

PR looks good otherwise.

PR #1037, opened 2015-11-30
Make objective short notice brighter

akien-mga:
The dark blue was very hard to read.


akien-mga:
BTW while looking closely at the event notice messages, I saw that its is also somewhat flickering for me. And even by looking closely at the chat messages, they flicker a bit. So it's not just a driver limitation of @oyvindln.

hwoarangmy:
Yes. As we said during the game, I would guess that it comes from the minimap (like the flickering keeper hand). Probably because rendering the scene in the minimap is cpu consuming.

hwoarangmy:
Looks good

PR #1042, opened 2015-12-02
Implemented chat in configure seat menu + small fixes

hwoarangmy:


akien-mga:
Code looks good overall, and the chat seems to work just fine :)
BTW the chat is also shown when starting a singleplayer map; should we hide it, or do we consider that it's not a big deal?

hwoarangmy:
For my part, I prefer when it is shown. But we can later disable it if we want. It is pretty simple...

akien-mga:
I've played a bit more with it, good to go IMO.

PR #1045, opened 2015-12-04
Merged EntityBase into GameEntity

hwoarangmy:


akien-mga:
It seems to still work fine in game, and I trust your code to be good, so let's merge :)

PR #1046, opened 2015-12-10
Implemented Casino + fixed menu nick handling + some code cleaning

hwoarangmy:
For casino, I've gone for the lazy way by using ground mesh + spots from other rooms. I'm sure @Danimal696 will do something about that when he will have some time.

Basically, the casino needs 2 creatures per active spot (if there is only one, it will wander until another one comes). They bet up to 200 gold per creature and the keeper takes 10% of the bets. If there is no space on the treasuries, the gold is lost.

Whatever the configuration is, there should never be 2 creatures alone on 2 different spots. If it happens, it is a bug ^^
For example, if creatures A and B are playing while C is alone, if B leaves, A should start playing with C.


akien-mga:
Tested it in game and it seems to work pretty fine already, let's merge and fine tune little by little :)

akien-mga:
I'm going to make a linux binary with the OIS fork I mentioned in the other issue, to see if it improves input management.

PR #1047, opened 2015-12-12
Technical improvement of fighting behaviour in arena

hwoarangmy:
Used same code for casino and arena that allows to fight in a restricted area (and that can be interrupted if an enemy comes)


akien-mga:
Works fine :)

PR #1049, opened 2015-12-17
Allow to play sounds on some game events + sound improvements

hwoarangmy:


Bertram25:
Cool, will test now! :D

Bertram25:
I tried it and the progress on the game are excellent. The keeper voice is making my day.
@eugeneloza Your voice is in!!! :D 

On the counter part, I cannot hear any spatial sound on a linux build, but as I recently changed my laptop, @akien-mga could you confirm?
I would also fix the folder naming for relative/ -> Relative/ or make Spatial/ be spatial/

Cheers :D

hwoarangmy:
@Bertram25 I've fixed what you saw ^^

hwoarangmy:
oops, I've just seen I forgot to update creadits for @eugeneloza sounds. I will fix it ASAP in this PR or a next one

Bertram25:
I'll test asap :). The thing is, did you rename the relative/ folder to Relative/ as well?

hwoarangmy:
> The thing is, did you rename the relative/ folder to Relative/ as well?

yes. That's what I did on the 2 last commits. From what I've seen, the easiest way was to remove everything from the folder and re-add with the correct name.

akien-mga:
> yes. That's what I did on the 2 last commits. From what I've seen, the easiest way was to remove everything from the folder and re-add with the correct name.

Well the easiest way is `git mv` :)

hwoarangmy:
> Well the easiest way is git mv :)

I tried git mv but it doesn't work on non-case sentive OS. I would have to rename the folder to "relative2" and then to "Relative" in another commit which would result in removing/adding everything twice. For source files, it would be better because it would avoid to loose history (i guess) but for sound files, it is not as important, especially for us, since there is no history ATM.

Bertram25:
Testing...

Bertram25:
The game is working fine, but I cannot hear any sounds, except for the keeper voice.
@akien-mga Can you confirm?

hwoarangmy:
@Bertram25 I was thinking about some sensitiveness problem but couldn't see any problem. Could you please post a log file ?

Bertram25:
> @Bertram25 I was thinking about some sensitiveness problem but couldn't see any problem. Could you please post a log file ?

Sure. On it.

Bertram25:
@hwoarangmy Here it is:
http://pastebin.com/kG9HXRNy

hwoarangmy:
@Bertram25 From what I see in your log, there is no error (if a spatial sound was launched but not found - because of wrong path or anything it would print an error).
Could you please try to play with the sound parameters ?
https://github.com/OpenDungeons/OpenDungeons/blob/development/source/sound/SoundEffectsManager.cpp#L54

Bertram25:
Well, IIRC, I was the one adding that piece of code, so I'm surprised it's not working anymore. I'll check whether the sounds get culled or anything.

Bertram25:
> I was the one adding that piece of code, so I'm surprised it's not working anymore

(I don't mean I'm too good for that to happen of course, but I spent a lot of time on it, that's why. ;])

Bertram25:
Ok, I may be in need of sleep. ;)
This is working fine. there was two things:
1. My music volume config was somehow too low for me to hear sounds on the new laptop.
2. I somehow assumed there should be sounds in the Training Hall, and after fixing the sound volume, I realized that part rather fast.

Always check the plug, they say. ;)
I'll merge. Let's move on!

Bertram25:
@hwoarangmy @akien-mga Btw, now the game looks much nicer in term of ambiance sounds, is it a good idea to add:
- Fighting sounds in the arena/TRaining hall?
- Claw/roar sounds for most of the Keeper species, instead of the usual metal clung, we have for everyone?
- Step sounds?

I also think we should think about adding a time offset for the actual attack sound play. So we can tune it to make it sync with the attack animation of creatures. WDYT?

hwoarangmy:
I'm ok with that. Do not hesitate to open a new issue as it might interest also @Danimal696 and it will be a better place to discuss that ;-)

PR #1053, opened 2015-12-28
Added a new room object for casino + fire particle effect

hwoarangmy:


Bertram25:
Will test now. :)

hwoarangmy:
ok, I will do corresponding fixes

hwoarangmy:
@Bertram25 It should be good now ^^

Bertram25:
Lol, why 2 commits? I'll test in any case.

hwoarangmy:
> Lol, why 2 commits? I'll test in any case.

Because on non case sensitive OS, there is no choice for changing sensitiveness ^^
git sees no difference unless the file is renamed

Bertram25:
> Because on non case sensitive OS, there is no choice for changing sensitiveness ^^
> git sees no difference unless the file is renamed

Arg, soo true.

Bertram25:
Ah btw, I do think you could replace the fire spell particle effect with that new one right away. Just my two cents, though.

hwoarangmy:
> I do think you could replace the fire spell particle effect with that new one right away

I have another use for this one ;-)

Bertram25:
Merging!

akien-mga:
In such cases it might be nice to merge the two renaming commits into their parent with git rebase -i. :)

Bertram25:
> In such cases it might be nice to merge the two renaming commits into their parent with git rebase -i. :)

This might not work either on case insensitive OS, but I don't remember trying, though.

hwoarangmy:
> In such cases it might be nice to merge the two renaming commits into their parent with git rebase -i. :)

I've not tried but I would bet git would detect no difference between the files and would cancel the change

akien-mga:
I'm pretty sure it would work when merging commits B and C into A at once :)

hwoarangmy:
> I'm pretty sure it would work when merging commits B and C into A at once :)

OK, I will try next time ^^

PR #1057, opened 2015-12-31
Started adding support to create new maps.

Bertram25:
There are still a few things to do.

This is for discussion purpose. Maps created through the menu are saved correctly but not yet functional.


hwoarangmy:
@Bertram25 Do you know why Travis build failed ? It seems to have been canceled and I guess you made your changes on Linux so it's weird to have a problem here...

akien-mga:
It's the download of boost from SF that failed. We really need to find packages for Ubuntu to avoid building from scratch at each commit. Are we using Travis with Ubuntu 12.04 or 14.04?

Bertram25:
Happy new year guys! :)
Yes, it's a timeout encountered on Travis. I used to break different make step into several - instructions to avoid this.
Maybe we could start by splitting make and make install instructions? WDYT?
Otherwise, using precompiled boost libraries sounds like a much better option, indeed.

akien-mga:
It looks like a SF network issue more than a timeout though, there's not much we can do IMO.

Bertram25:
And what do you think about the current new map state?

I plan to make a minimal create map dialog work, and add more and more options (like seats, objectives, ...) in the next one. WDYT?

hwoarangmy:
Happy new year everybody ^^

@Bertram25 Testing now

hwoarangmy:
I've tested and it looks good. Things I've seen:
- Cancel/create buttons are not centered and too small (full label is not visible)
- When I clicked "create map", I got a crash (telling map is not found).

Have you thought about displaying the window for existing levels to allow editing ?

Bertram25:
First goal reached. You should now be able to create new maps and save official ones as custom ones.

> Cancel/create buttons are not centered and too small (full label is not visible)
> When I clicked "create map", I got a crash (telling map is not found).

Should be fixed. :)

> Have you thought about displaying the window for existing levels to allow editing ?

If you're speaking of their properties, yes.
At the moment, you should be able to create a basic map with one team, an load existing maps, as usual, but save official ones as custom ones, to permit edition on linux most notably where the game is installed in the system. (But also on Windows proper installations ...)

I left the new map creation window as slick as possible to permit focusing on the map properties once you're in the editor. I'll start adding map properties (seat, team, objectives, music, ...) in the next PR.

Feel free to comment. :)

hwoarangmy:
> At the moment, you should be able to create a basic map with one team, an load existing maps, as usual, but save official ones as custom ones

IIRC, that's what we said we wanted to do. And that makes sense IMHO. However, it might be a bit misleading to think that you open a map trought "Official maps", save it, and, to open the map you modified, you have to go to "custom maps". It would be nice to have some visual notice that the map has been modified and a similar custom map already exists (probably using level file name).

> I left the new map creation window as slick as possible to permit focusing on the map properties once you're in the editor. I'll start adding map properties (seat, team, objectives, music, ...) in the next PR.

Nice ^^
We are starting to have a decent map editor

hwoarangmy:
I've tested and got an error because the paths you use are not created. In the RessourceManager, you need to create levels/skirmish and levels/multiplayer.

However, I've tested after manually creating the paths and it works well ^^
I will try to review ASAP

akien-mga:
I also had a very quick test of the map saving (love the new UI!) and it said that the save was succesful (showing the path), but indeed I got no file as the path were not created. Two issues I guess:
- write error should be reported properly
- paths should be created if missing

Bertram25:
> - write error should be reported properly
> - paths should be created if missing
> - try to sort headers in new file.

k, on it. :)

akien-mga:
Code looks good overall :)

Bertram25:
Btw, I udpated our copy on gitlab: https://gitlab.com/groups/opendungeons

akien-mga:
> Btw, I udpated our copy on gitlab: https://gitlab.com/groups/opendungeons

Nice!

Bertram25:
> It would be nice to have some visual notice that the map has been modified and a similar custom map already exists (probably using level file name).

@hwoarangmy I'm not forgetting you. :)
I still have to think about it some more, but as a first concept, what about adding a star icon next to the official map name in that case?

hwoarangmy:
> I still have to think about it some more, but as a first concept, what about adding a star icon next to the official map name in that case?

IIRC, that's what you proposed last time. Seems like you agree with yourself ;-)

Bertram25:
> IIRC, that's what you proposed last time. Seems like you agree with yourself ;-)

That part is rather easy, I just hope others are, too. ;)

Btw, this is done:

> write error should be reported properly
> paths should be created if missing
> try to sort headers in new file.

I'm on the star icon feature now.

Bertram25:
Ok, all requests should be done now.

akien-mga:
It works great, I like the cog icon + explanation in the level desc that a custom level exists for it.
It created my ~/.local/share/opendungeons/levels path just fine and the subdirectories too. Good to merge as for me :)

Bertram25:
@hwoarangmy @akien-mga Btw, I would not add the same mechanism in the skirmish/multiplayer map menu, as it doesn't make sense to me there. WDYT?

akien-mga:
> Btw, I would not add the same mechanism in the skirmish/multiplayer map menu, as it doesn't make sense to me there. WDYT?

You mean in the menu to choose the map you want to play? If so I agree, the filters are enough, we don't need to tell users that they have a custom level of an official one if they selected the official levels filter to play.

hwoarangmy:
> Btw, I would not add the same mechanism in the skirmish/multiplayer map menu, as it doesn't make sense to me there. WDYT?

As a player, I would expect the same behaviour for editor or gamemode as we choose in both which kind of level we want to open (custom or official). However, when we started talking about using a different path to save custom levels and using a cog to make a difference, we didn't had the custom/official choice in the combo.

If I try to take some distance, I would say that I would expect to have to use "custom maps" to open custom maps and "offical maps" to open official maps. I know I was the one asking for this feature (and I'm sorry about that) but it might not be such a good idea.
After all, wouldn't an explicit message when opening an official map and saving it telling that it will be available under "custom maps" filter be sufficient ?

akien-mga:
Well IMO the cog when opening an official map for _edition_ is an important feature, as we don't want _modders_ to have to go back and forth ten times between the "official" and "custom" sections in the combobox to check which official levels have custom equivalents.

It's less needed for simpler _players_ as they are less likely to have custom maps that are the same as official maps - this situation would happen mostly for modders who want to propose us an enhanced version of an official map. If a player has such a map, he probably knows about it and will look for it directly in the custom section when selecting a level for playing it. So the cog icon is really just a convenience feature for modders to let them know that they have among their own custom levels an enhancement of an official map.

hwoarangmy:
> Well IMO the cog when opening an official map for edition is an important feature, as we don't want modders to have to go back and forth ten times between the "official" and "custom" sections in the combobox to check which official levels have custom equivalents.

They won't have to go back and forth. They will only have to open it once and, each time they will want to use it (for test as well as editing), they will have to go to the custom section. IMHO, it is more user friendly.
Moreover, ATM, I guess the map will be editable in both menus (official and custom), which sounds weird.
We have not talked about that but we might also want a feature to delete a custom map. Again, it will be easier to add a button in custom maps and nothing to do with official ones. If we keep the current behaviour, deleting a custom map would not be possible in the official menu but it would in custom. That's not user friendly again IMHO.

> It's less needed for simpler players as they are less likely to have custom maps that are the same as official maps

IMHO, we should try to keep the same behaviour in the menus as long as we can. If we want to allow choosing between official and custom maps in the game selection menu, I think we should have a similar behaviour in the editor.

Bertram25:
> ATM, I guess the map will be editable in both menus (official and custom), which sounds weird.

Should we add a notice when showing official maps in the editor mode to tell they will be saved as custom ones?
Otherwise, I won't understand as a user why there is a cog in official maps when just playing. (The less info, the less confusing, IMHO.)
But having the info when modding is relevant, so I'm with @akien-mga on this one. Did I miss something else?

hwoarangmy:
> Should we add a notice when showing official maps in the editor mode to tell they will be saved as custom ones?

IMHO, a message when the map is saved telling it is an official map and it will be saved in the custom maps would be enough. As you say, the simpliest, the better. But I will try right now and tell you how it feels

hwoarangmy:
I've tried and it works as expected. However, I confirm I find pretty weird to use the custom map when having the combo on "official maps". I agree the explanation and the cog icon makes pretty clear what's going on but this behaviour is not what I would have expected.

IMHO, the cog icon we have been talking about was a good idea if we didn't had the combos for selecting official/custom maps. But I believe your idea of using the combo is better than the cog icon but it makes the cog icon less relevant. But well, I know I was the one asking for the feature so if you want to keep it, I won't go against ^^

@Bertram25 Just tell if you want to add something more in this PR but it is mergeable for me

akien-mga:
I think it's good to merge too, so let's go :)

Bertram25:
> I confirm I find pretty weird to use the custom map when having the combo on "official maps".

Hey again, to clarify that point, when you are selecting the official map, it will load the official map, and when you select the custom one, it will load the custom one.
But, after loading an official map, it will save as a custom one. This is necessary to ensure official maps don't get "corrupted" and as it won't be possible anyway on system installation, especially on Linux.

The cog icon is here to tell there is already a custom map and that the modder may overwrite it if selecting the official one.

Is it clearer?

hwoarangmy:
> But, after loading an official map, it will save as a custom one. This is necessary to ensure official maps don't get "corrupted" and as it won't be possible anyway on system installation, especially on Linux.

Yes, I understand this point. I'm just saying that it seems more straight forward to me to open an official map selecting the official combo. If I save it, having a message telling me that it has been saved as a custom map and that to open it, I shall open it through custom maps.
Then, if I open the map through official maps, I will open the official map again (without my changes). And if I open it through custom maps, I will open the map I did.
This behaviour seems pretty straight forward to me and simpler. As you said, IMHO, the less info, the less confusing.

But again, if you prefer this implementation, I'm okay with it

akien-mga:
> Yes, I understand this point. I'm just saying that it seems more straight forward to me to open an official map selecting the official combo. If I save it, having a message telling me that it has been saved as a custom map and that to open it, I shall open it through custom maps.
> Then, if I open the map through official maps, I will open the official map again (without my changes). And if I open it through custom maps, I will open the map I did.
> This behaviour seems pretty straight forward to me and simpler. As you said, IMHO, the less info, the less confusing.

I'm still not sure we all understand the same behaviour. The behaviour you described is exactly what is implemented and what should be done. The cog icon only given an _information_ that there also exists a custom map for this official map. So if you load the official map nevertheless, you are then aware that it will override your existing custom map with the same name. The cog icon does not indicate that it will _load_ the custom map instead of the official one, just that this custom map _exists_, and it's good to know it as loading the official map will likely mean overriding the custom one.

Bertram25:
What @akien-mga said. :)

At least it's the intended implementation.

hwoarangmy:
ok. I had tested but got a strange behaviour. I saved a map after making a small change to check the behaviour but for some reason, I thought the map opened when selecting "official skirmish maps" was the one I modified. I don't kow if I got something wrong or if I made a mistake but I just retried and couldn't reproduce.

In reality, the cog icon is only informative so it is even better than I thought. Sorry for having discussed so much a non existing feature :-(

akien-mga:
Np, thanks for discussing it extensively so that we're sure that we implemented the right thing :-)

Bertram25:
No problem at all. I prefer this to be clear and acceptable for everyone. :)

PR #1062, opened 2016-01-03
Fixed crash when several creatures tried to carry the same dead body

hwoarangmy:


Bertram25:
So it's fixed. Let's merge.

PR #1063, opened 2016-01-04
Allow owned creatures to be put on prison and to display icons when jailed + few fixes

hwoarangmy:


hwoarangmy:
@akien-mga It should be ok now

akien-mga:
Builds and seems to work OK (though I did not test the additions extensively).

PR #1066, opened 2016-01-06
Background scene + made creature steal gold on the ground

hwoarangmy:


Bertram25:
I missed the previous one, but I'll definitely have a go for that one. :D Title looks great. !!!

akien-mga:
Ah ah that looks awesome! :D :dancer:

akien-mga:
Good to merge for me, but I know @Bertram25 will want to test it live too, so I keep it around for now :)

akien-mga:
BTW the new main menu scene kind of shows that our elf could use a better skin texture, it looks a bit too uniform IMO - the mesh is great overall though.

hwoarangmy:
> BTW the new main menu scene kind of shows that our elf could use a better skin texture, it looks a bit too uniform IMO - the mesh is great overall though.

I would say it also shows that our models are not bad as they look pretty cool on bigger view. IMHO, we should also try to allow the camera to zoom more to be allowed to see them like that in game ^^
I like the eye animation the troll has from time to time ^^

akien-mga:
> I would say it also shows that our models are not bad as they look pretty cool on bigger view. IMHO, we should also try to allow the camera to zoom more to be allowed to see them like that in game ^^
> I like the eye animation the troll has from time to time ^^

+1 - our models are awesome, and we're not far from being able to reproduce the battle scene that @Danimal696 made in Blender as an animated OGRE scene in the main menu :D

hwoarangmy:
> we're not far from being able to reproduce the battle scene that @Danimal696 made in Blender

We are not far, it's already done ;-)
We just have to add more models (but they would be smaller). It's only a matter of what we want

akien-mga:
> We are not far, it's already done ;-)
> We just have to add more models (but they would be smaller). It's only a matter of what we want

Well coolness-wise we still have some work to do to get that I think ;) http://forum.freegamedev.net/viewtopic.php?f=36&t=6540#p65823

hwoarangmy:
You are right, I forgot the light effects (which are pretty nice). But I don't think we should go for that as it is pretty CPU consuming

akien-mga:
> You are right, I forgot the light effects (which are pretty nice). But I don't think we should go for that as it is pretty CPU consuming

I meant more the carefully crafted battle stances shown in this image that were made in Blender and do not correspond to our models' battle animations - our current battles on closeup don't look that cool yet, though I'm pretty confident we will go towards that little by little :)

Bertram25:
Looks great for a first pass :D I'll merge!

A few things noticed that can be discussed for later:
- The scene has no light, making the model look flat, and show their main flaws.
  We could later add a mini-scenario, something like:
- I would add a corridor on the right of the scene, and why not make a knight come down, look and go away fleeing ;)
- The small scarab could look at the golem and dance for it. The golem could be placed slightly turn to the right or the left to make the scene look less orthogonal.
- ...

PR #1069, opened 2016-01-07
Replaced BigKnight with a new creature

hwoarangmy:
@OpenDungeons/developers If you all agree with the replacement, I will upload the original model on the media repo (found on opengameart already animated)

I only had to decimate it a bit to reach 4.5k tris and to export animations


Bertram25:
I will test now. :) My sole question so far is why was the PitDemon in particular chosen? Was there a problem with it (that I missed somehow)?

Bertram25:
I like the new creature. I miss the Pit demon though, but maybe there is a good reason.

hwoarangmy:
Because the creature looks like a tier 4.

Defender, Scarab, Orc and Troll are already final creatures. That only left PitDemon and Dragon (not good enough models compared to the rest). IMHO, Dragon is a little bit better than PitDemon so it got chosen ^^

Bertram25:
I see. We miss a very demonic strong creature for the dark side, but the models looks good. :)

hwoarangmy:
We could also replace BigKnight with this one (it can fit hero side IMHO)

Would you prefer ?

Bertram25:
> We could also replace BigKnight with this one (it can fit hero side IMHO)
> Would you prefer ?

Yes, definitely. Good idea. :)

hwoarangmy:
won't you miss too much the bigkight ?

Bertram25:
> won't you miss too much the bigkight ?

No, since we already have knights (not big ones) using the same keeper-nightmarish model. :)

hwoarangmy:
@Bertram25 @akien-mga It's done

akien-mga:
Looks great! :D

Bertram25:
Cool, thanks :)

PR #1075, opened 2016-01-11
Missing elf hair material

Danimal696:


akien-mga:
BTW, looks like the Elf.mesh is at the wrong place: https://github.com/OpenDungeons/OpenDungeons/blob/development/materials/scripts/Elf.mesh
That's probably why its skin looks bad? I guess you copied Elf.mesh instead of Elf.material.

hwoarangmy:
> BTW, looks like the Elf.mesh is at the wrong place: https://github.com/OpenDungeons/OpenDungeons/blob/development/materials/scripts/Elf.mesh

Nice catch, I had not seen this one ^^ Seems like I merged too fast.

@Danimal696 Also note that it might be good to merge all the materials from a model into 1 material file (Elf.material would contain every material used by the model for example). That would avoid to have too many material files in the directory.
I will open an issue for that.

Danimal696:
I have no idea why the skin looks that bad, tried changing material values but couldnt fix it, its got something to do with its normal map.

Bertram25:
I'll merge this for now, and I'll try to find better values for the hair.

PR #1078, opened 2016-01-11
Fixed Elf Hair Normal Appliance.

Bertram25:


akien-mga:
I've tried to fix ElfSkin.material, but I can't find any value anywhere to make the normal map more prominent. Changing values for the decalmap seem to have almost no effect :/

Bertram25:
> the decalmap seem to have almost no effect :/

I confirm. The base texture has to be righteously made in that case. Maybe adding light values within the base texture would help though.
I'll test.

PR #1080, opened 2016-01-12
Merged Elf/DarkElf materials into 1 file per model

hwoarangmy:


Bertram25:
Looks pretty SF.

akien-mga:
Ping @Danimal696 so that you can see how to have several materials defined in one file. Basically you just need to copy paste them after one another :)

Danimal696:
And how did you fix the materials call from the mesh?

hwoarangmy:
In reality, ogre opens all material files it finds from the material folder. The materials will be found whatever file name they have. The only relevant part is the material itself defined in the .material. Moreover, there can be more than 1 material defined in a file.
And since there is no change in the material names, there is no need to change the mesh.

Note that the particles also work the same. ATM, we use 1 particle file per particle effect but it would work as well to have only 1 particle file and every particle effect defined inside.

PR #1086, opened 2016-01-23
Main menu scene improvements

hwoarangmy:
With this PR, the main menu scene can be changed with the config file.
That will also help much testing the models for artists as it will allow to easily test the models and animations ingame without having to launch a game.

Note that most of the changes do not affect the game. The biggest impact is on modes navigation  (going forward/backward in the menus, going to menus after a game, ...).

Another goal of this PR is to allow to have more than 1 scene (for example for loosing/winning screens)

If this PR is not merged before, I might add some new commands to the script menu.


hwoarangmy:
For the record, I plan on implementing:
- "wait" command
- A "repeat" tag that would allow to go back to some scene in a scene group

Bertram25:
Great, I'll test all that asap. :)

Also, if you're into scripting mood, we may need to discuss a bit about how to do it, be it lua or something else?

hwoarangmy:
> Also, if you're into scripting mood, we may need to discuss a bit about how to do it, be it lua or something else?

Well, for now, I've used a script logic like for mood and other scripts we already use. It might have been a good opportunity to start looking at AS, LUA or something else but I don't know how to use them and I didn't feel like messing with it.

Bertram25:
> Well, for now, I've used a script logic like for mood and other scripts we already use. It might have been a good opportunity to start looking at AS, LUA or something else but I don't know how to use them and I didn't feel like messing with it.

Yep, fine by me. Especially since it seems we're near to ship, I guess now is not the time. But we can have a talk about it once at the beginning of 0.8 :)

Bertram25:
Code looks fine and game working great. :+1: 
The thing is: we can clearly see that the elf and the rat walking animation aren't fitting their walking speed. But that's something else.

Bertram25:
Mergeable as for me.

akien-mga:
I'll do a quick test and merge it tonight then :)

hwoarangmy:
> The thing is: we can clearly see that the elf and the rat walking animation aren't fitting their walking speed.

You can also clearly see that the roach is burning. Usually, burning roaches do not dance ;)

akien-mga:
> You can also clearly see that the roach is burning. Usually, burning roaches do not dance ;)

I like setting bugs on fire... :D

Bertram25:
>  Usually, burning roaches do not dance ;)

Except for their master :>

akien-mga:
Works fine, I merge :)
I'll review the commits themselves later, at least to know what was changed :D

PR #1101, opened 2016-02-28
Support to launch custom maps with master server + used new objects for casino room

hwoarangmy:


Bertram25:
Review time :D

Bertram25:
Everything works fine, and standard way to play still works, so I'll merge :)

It was uneasy to know about the save filename so I think this will stay a debug function.

PR #1102, opened 2016-03-01
Added support for selecting AI level + small fixes

hwoarangmy:


Bertram25:
Looks fine as for the code, can't wait to test that. :)

Bertram25:
N.B.: If the Travis builds are broken and cannot be fixed easily we should disable them for now IMHO.

hwoarangmy:
> N.B.: If the Travis builds are broken and cannot be fixed easily we should disable them for now IMHO.

It looks like libogre-1.9-dev has been moved from one of the repositories we had used. IMHO, we should try to fix it since it can avoid big regressions. I am not comfortable with the fact that it doesn't work anymore. It is pretty easy to break compilation on a non tested architecture. I was waiting a bit to see if it was going to be fixed in the repository but since it is not, we should try to fix it. I will have a go when this PR is merged.

Bertram25:
Working fine. Still needs to fix getAi() and I think it's mergeable.

hwoarangmy:
@Bertram25 It should be good now

PR #1106, opened 2016-03-08
Added synchronisation commands in menu scene

hwoarangmy:


akien-mga:
Looks good. Travis CI build failed for no obvious reason, but I'll have a look at improving the file further. We don't need e.g. to install gcc-4.8 manually as it's the default version.

PR #1107, opened 2016-03-09
Improve travis-ci build scripts

akien-mga:
- No need to build SFML, we install it from trusty
- GCC is already 4.8 (at least when requiring cpp language)
- Disable caching and addons as they don't work on non
  container-based platforms (i.e. we can't use caching and sudo)
- Cleanup a bit and install files to our path
- Install cegui-0.8 from PlayDeb repo

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,26 +19,20 @@ compiler:
     #- binutils-dev
 
 before_install:
-  - export HOME_DIR=`pwd`; gcc --version
+  - gcc --version
   - sudo apt-get update -qq
 # OD dependencies
-  - sudo apt-get install -y cmake libogre-1.9-dev libsfml-dev libois-dev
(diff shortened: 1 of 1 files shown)


akien-mga:
I'll see if I find a PPA with cegui so that we don't have to build it.

akien-mga:
Seems to work fine (now rebuilding again as I just force pushed a small cleanup).

akien-mga:
8 min build on travis :)

hwoarangmy:
IMHO, you could remove the commented lines. They are in the history if it breaks something in a near future and we don't need them if they break nothing ^^

hwoarangmy:
@akien-mga Great work, it is much better like that. I will wait a bit if @Bertram25 wants to have a look and merge soon

PR #1114, opened 2016-03-11
Changed main menu screen + minor improvements on console

hwoarangmy:


akien-mga:
Just had a quick glance at the code, but I'll merge it anyway. I still think our PR-based process is very good, but since we're all kind of busy I prefer not to let PRs sit around for too long (at least when coming from @hwoarangmy or @Bertram25 that always do code of very good quality).

Regardless, having PRs is great to be able to see what code changed, what are new features and API changes, etc., so we should keep this workflow :-D

akien-mga:
I love the scripted scene on the main menu :D
It takes a while to load at first though, we might need to have a look at adding a loading bar, or maybe if we can instance the scenery first, and load the rest of the meshes afterwards? (but I don't think we have support for that so far).

akien-mga:
Actually only the first loading was slow, then I think stuff stays in cache so it actually starts relatively fast. Having the first loading slow is OK IMO.

Bertram25:
@akien-mga Thanks for thinking I'm doing quality wirk. ;) I do think also that @hwoarangmy has a very good level. Yet, I would still enforce the rule saying the one making the code isn't the one merging it when possible. Even your quick glance at the code could prevent obvious flaws to happen.

akien-mga:
> Yet, I would still enforce the rule saying the one making the code isn't the one merging it when possible. Even your quick glance at the code could prevent obvious flaws to happen.

Yes, if only so that contributors that did not write the code can stay on top of the new developments and know what gets merged. It's easier to review code in pending PRs that to review a backlog of merged PRs ;)

hwoarangmy:
As you said, it is pretty nice to help following what happens and find what's new when we do snapshots. Concerning code review, I also like how we do ATM. We have examples of basic errors we have all done and it helps finding them. Moreover, it allows to discuss architecture choices and new code features ^^

PR #1115, opened 2016-03-12
Torture room object

Danimal696:
The current one is downscaled by code, please remember to delete that
when adding this.


Danimal696:
Im also adding a small test level i have been using for weapons show

akien-mga:
Woooot! :D

akien-mga:
I believe we still lack the CREDITS update for those objects @Danimal696 :)

PR #1123, opened 2016-03-12
Swords, hero version

Danimal696:


hwoarangmy:
I've tested but the image file "T2SwordGoodNormal.jpg" you uploaded is jpg while the one in the material is png. Thus, I get an exception. Could you please upload the png version ?

Danimal696:
my bad, that was the original map

hwoarangmy:
good

akien-mga:
Also missing credits update here.

PR #1126, opened 2016-03-18
Added a unit test for traps

hwoarangmy:
@akien-mga I promess I won't add another PR before we do the release ;-)

Note that I will go for the binaries tonight so it is available for the lan

### DIFF
--- a/source/tests/mocks/ODClientTest.cpp
+++ b/source/tests/mocks/ODClientTest.cpp
@@ -410,7 +410,33 @@ bool ODClientTest::processMessage(ServerNotificationType cmd, ODPacket& packetRe
                 BOOST_CHECK(packetReceived >> walkDirection);
             }
 
-            handleSetAnimationState(entityName, animState, loop, playIdleWhenAnimationEnds, shouldSetWalkDirection, walkDirection);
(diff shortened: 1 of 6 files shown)


source/tests/mocks/ODClientTest.cpp
```diff
@@ -410,7 +410,33 @@ bool ODClientTest::processMessage(ServerNotificationType cmd, ODPacket& packetRe
                 BOOST_CHECK(packetReceived >> walkDirection);
```
akien-mga, line 410: Again I can't find how to comment on a single commit, GH is failing me :°)

Just wanted to say that you should remember to make commit logs as "short title + newline + commit description", instead of a one liner as shown above. So this log should have been something like:

```
Pass information about the played animation in unit tests messages

In the unit tests, we should call a function telling the animation played by entities for
every message that can change the animation as it is heavily used to know what's happening.
```

(the empty line is important to separate the first "commit title" from the "commit description").

akien-mga:
Code looks good. I guess what's left is to add this test to the travis CI and appveyer build scripts?

akien-mga:
> I guess what's left is to add this test to the travis CI and appveyer build scripts?

Ah no we designed out scripts in a clever way already :D

PR #1155, opened 2016-03-31
Small technical improvements / fixes + 1 new spell (eye of evil)

hwoarangmy:


source/entities/Creature.cpp
```diff
@@ -1887,8 +1887,7 @@ void Creature::createStatsWindow()
                                         CEGUI::Event::Subscriber(&Creature::CloseStatsWindow, this));
 
     // Set the window title
-    childWindow = mStatsWindow->getChild("__auto_titlebar__");
-    childWindow->setText(getName() + " (" + getDefinition()->getClassName() + ")");
+    mStatsWindow->setText(getName() + " (" + getDefinition()->getClassName() + ")");
```
Bertram25, line 1891: good!

source/entities/Creature.cpp
```diff
@@ -1881,10 +1882,9 @@ void Creature::createStatsWindow()
     textWindow->setProperty("FrameEnabled", "False");
     textWindow->setProperty("BackgroundEnabled", "False");
 
-    // Search for the autoclose button and make it work
-    CEGUI::Window* childWindow = mStatsWindow->getChild("__auto_closebutton__");
-    childWindow->subscribeEvent(CEGUI::PushButton::EventClicked,
-                                        CEGUI::Event::Subscriber(&Creature::CloseStatsWindow, this));
+    // We want to close the window when the cross is clicked
```
Bertram25, line 1886: Good riddance of the auto close button ref. :)

source/entities/DoorEntity.cpp
```diff
@@ -49,7 +49,8 @@ DoorEntity::DoorEntity(GameMap* gameMap, Building& building, const std::string&
 }
 
 DoorEntity::DoorEntity(GameMap* gameMap) :
-    TrapEntity(gameMap)
+    TrapEntity(gameMap),
+    mBuilding(nullptr)
```
Bertram25, line 53: Ah, indeed.

source/network/ODServer.cpp
```diff
@@ -191,6 +193,7 @@ bool ODServer::startServer(const std::string& creator, const std::string& levelF
         if(!MasterServer::registerGame(ODApplication::VERSION, creator, port, label, descr, uuid))
         {
             OD_LOG_ERR("Could not register the game in the master server !!!");
+            stopServer();
             return false;
```
Bertram25, line 197: Looks good.

source/utils/StackTraceWinMinGW.cpp
```diff
@@ -17,6 +17,9 @@
```
Bertram25, line 17: That one looks good. Cannot really tell for the similar other ones, though,

hwoarangmy, line 17: Note that I've tested for the 3 and I confirm it works

hwoarangmy:
@OpenDungeons/developers I have been playing with blender trying rigging/animating and I found a simple model that could be good for evil eye. So, I've implemented it ^^
Note that when the evil eye is cast, the model is visible by everybody (but the tiles are visible by the caster and his allies only)

akien-mga:
Any idea why the MinGW builds fail on AppVeyor? https://ci.appveyor.com/project/hwoarangmy/opendungeons-crxx4/build/1.0.219/job/8hc4kn80jtng0n7j
It's not really verbose, sounds like it might be failing on the 7zip command.

source/entities/Creature.cpp
akien-mga, line 1891: Small git-fu notice: whenever possible, try to split git commit logs into a short title + a more lengthy description after an empty newline, e.g.:

```
Change creature window stats title by using setText

Previously, it was addressing built-in window which caused a crash when using CEGUI 0.8.5.
```

That would prevent a broken long title as we have here.

source/render/MovableTextOverlay.cpp
```diff
@@ -137,6 +137,14 @@ void ChildOverlay::update(Ogre::Real timeSincelastFrame)
     mOverlayContainer->hide();
 }
 
+void ChildOverlay::isOnScreen(bool onScreen)
```
akien-mga, line 140: Maybe `setOnScreen` would be a better name? We tend to use `isSomething` names for methods that return a boolean.

akien-mga:
I've had a quick look at all commits, tons of good stuff! :D
Let's merge as it all seems pretty solid and @Bertram25 also tested most of it already.

source/render/MovableTextOverlay.cpp
hwoarangmy, line 140: Yep, good idea

hwoarangmy:
> Any idea why the MinGW builds fail on AppVeyor?

No. Dependencies are correctly unzipped for msvs2013 so there should be no right problem. Now that this PR is in, I will try to have a look.

PR #1179, opened 2016-04-26
Creatures should go to the closest treasury room when getting their fee + resize all models to avoid ingame scaling

### LINKED ISSUE #1172: Creatures should use the closest room
I frequently see creatures walk past a room they need to go to another room of the same time. Often they turn angry in the process. They should always use the closest room, especially when they walk through it.

See this screenshot from the 20160402 build and see I dropped all my creatures next to a treasure room on payday only for them to walk through it to another treasure room:
bettergoldelsewhere

hwoarangmy:
fixes #1172


source/creatureaction/CreatureActionGetFee.cpp
```diff
@@ -93,14 +93,20 @@ bool CreatureActionGetFee::handleGetFee(Creature& creature)
         return true;
     }
 
-    uint32_t index = Random::Uint(0, availableTreasuries.size() - 1);
-    Room* room = availableTreasuries[index];
+    Tile* chosenTile = nullptr;
+    std::list<Tile*> tilePath = creature.getGameMap()->findBestPath(&creature, myTile,
```
Bertram25, line 97: Good! :)

config/creatures.cfg
```diff
@@ -13,9 +13,6 @@
 #                     workers will dig and claim tiles. When idle, fighters with a digRate > 0 can dig tiles as indicated
```
Bertram25, line 13: I have a problem with this commit: While this simplifies a lot of things, it isn't modder friendly. Is it at least documented in the wiki how to rescale a model for OD?

Bertram25:
Apart from my comment on https://github.com/OpenDungeons/OpenDungeons/pull/1179/commits/b5f601a5f15a1b63c299f36fa79b39a97ad07ed7 the code looks fine. :) I'll test this asap.

config/creatures.cfg
hwoarangmy, line 13: I have created the tools for windows and uploaded to our repo. I will go with the wiki when I can.
Anyway, I was planning on doing this for a while. We don't have much choice as it will inevitably create problems to have nodes with different scales for every creature. For example, you need to attach an entity to a bone (not a node) and, thus, the same weapon cannot be added to different models (at least if their size is different - that's the problem @Danimal696 had with the cultist). Moreover, I'm sure we will hit other problems with particle effects or whatever when we will add effects to models (as the effects will also be resized).

IMHO, the best is to have everything at the same size.
And, concerning modders, the way I see it, they should use the main menu screen for their tests. The creatures will be bigger, allowing to check the animations are ok, and it will be possible to add them next to another creature to have a reference for size.
I will also explain how to do that in the wiki. But IMHO, it is much more user friendly than to have to manually edit a level to add some new creature or to hack an existing one, to launch a game and pray for the expected one to spawn.

Bertram25, line 13: > IMHO, the best is to have everything at the same size.

To be completely honest, I do also think it's even simpler and may fix bugs by design. :) 
As long as it is planned to have a wiki page about this, I'm fine. Mind creating an issue about this?

Bertram25:
Latest patch looks fine.

config/creatures.cfg
hwoarangmy, line 13: done in #1185

akien-mga:
Sorry for the huge delay, I've been pretty busy lately. Will try to test it tonight and merge it then. I'm also starting to write down the detailed changelog, the release should be just around the corner.

Bertram25:
I'll be around this week-end. So the announcement should be done at the same time. :)

akien-mga:
Forgot to test and I'm away until Monday, but IMO feel free to merge and test in the dev branch.

akien-mga:
I tested a bit, especially regarding the model scaling change, and everything looks fine :)

PR #1186, opened 2016-05-15
Restored PrefixTree

tomluchowski:
Here I restore my PrefixTree class for console autocompletition

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -6,6 +6,104 @@ OpenDungeons since version 0.4.7, in reverse chronological order.
 The numbers preceded by a hash (e.g. #9) reference GitHub issue numbers on
 our repository at https://github.com/OpenDungeons/OpenDungeons
 
+### Version 0.7.0
+
+**Highlights:**
+
+* Multiplayer lobby using a master server, to find other players all over the world
+* Animated main menu with scripted events!
+* Play voice-overs on some game events
+* Vastly improved creatures behaviour, including support for spell casting,
+  elemental damage, better ranged behaviour, and many customisation features
+* New rooms: Arena, Bridges, Casino, Prison, Torture room
+* New creatures: Cultist, Elf, Dark Elf, Lich, Nature Monster, Rat, Roach, Skeleton
+* New spells: Defense, Eye of Evil, Slow, Strength, Weak
+
+**General:**
+
+* Play sounds on some game events, with voice-overs by Eugene Loza (#393)
+* Unit tests (#578) and continuous integration (#960)
+* Display stacktrace in the logs on game crash (#1155)
+* A lot of code refactoring and technical improvements!
+
+**Multiplayer:**
+
+* Master server and in-game lobby, allowing to find available servers
+  waiting for players (#777)
+  - Chat support in the lobby/seat configuration menu (#1029)
+* Improved game saving, only allowed for the server side (#938)
+
+**Gameplay:**
+
+* Creatures behaviour:
+  - New KO state as an alternative to death (#751)
+  - Vastly improved fighters behaviour (#957)
+  - Vastly improved workers behaviour (#1004)
+  - Implemented creature skills, abstract mechanism for both fighting
+    and spell casting (#959)
+    * Creature spells: Defense (#986), Slow, Strength, Weak (#1001)
+  - Elemental damage type, creating a triangle relationship between
+    strength and vulnerabilities (physical, magical, elemental) (#877)
+  - Make creature moods configurable like other parameters (#979)
+  - Make creature behaviours fully configurable (#992)
+  - Creatures grow bigger as they gain levels (#992)
+  - Creatures will use the closest room if multiple are available (#1172)
+  - General creature balancing (ongoing effort) (#985, #1001)
+* New rooms/traps:
+  - Arena, where creatures fight one on one or in a battle royale until
+    KO to earn XP (#699)
+  - Bridges (wooden for water, stone for water and lava)(#191)
+  - Casino, where creature spend their wages (#1046)
+  - Prison, turns both foes and friends into skeletons (#754)
+  - Torture room (#1058)
+* New gem tile, provides an infinite source of gold (#735)
+* Trap balancing by making them take defence skills into account (#1067)
+* New setting to select the AI level (Easy and Normal so far) (#933)
+* New Eye of Evil spell
+
+**Levels:**
+
+* Completed user maps support: official maps and user-made maps are
+  properly differentiated (#383)
+* New multiplayer level: Angel Nest (#1130)
+* New skirmish levels: Duel to Death, Forgotten Treasures (#1130)
+* Stonekeep Siege remake for bridge usage (#929)
+* Dehardcoded portal spawn rate (#1119)
+
+**UI:**
+
+* Camera management feature using the middle mouse button (#892)
+* New map creation menu; missing part is the setup of level parameters
+  like configured seats, etc. (#22)
+* Main menu usability improvements (#1050)
+* Display a scriptable 3D scene on the main menu, with lights and
+  particle effects (#1064, #1074)
+* Fixed spell cooldown display (#974)
+
+**Graphics:**
+
+* New icons for stunned and jailed (#950)
+* New creatures:
+  - Cultist, by Danimal, Dread Knight and Katarzyna Zalecka (#1181)
+  - Elf and Dark Elf, by Contmike and Danimal (#1016, #1017)
+  - Lava Spawn, by piacenti, hwoarangmy et al. (#1190)
+  - Lich, from Summoning Wars (#992)
+  - Monk, by CDmir and hwoarangmy (#1190)
+  - Nature Monster, replacing the BigKnight, by CDmir (#1069)
+  - Rat, replacing the Wyvern, by CDmir (#1072)
+  - Roach, by Danimal and Atmostatic (#962)
+  - Skeleton, from Summoning Wars (#992)
+* New objects:
+  - Room objects:
+    o Casino beer stand and poker table, by Danimal (#1054)
+    o Roulette, by hwoarangmy (#1053)
+    o Torture room object (#1115)
+  - Weapons:
+    o Basic axe, hammer, shield and sword, by YD (#1115)
+    o Bow, by YD (#1065)
+    o Crossbow, by Lamoot (#1065)
+    o Good hammers, good shields, evil shields (#1115), good swords (#1123)
+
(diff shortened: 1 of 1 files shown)


tomluchowski:
Looks like I forgot to add PrefixTree.cpp to the cmake file. Could someone do that for me ? If not, I will try that over the weekend following the next one .

hwoarangmy:
Yes, this PR do not compile. Moreover, the code do not follow code guidelines. May I ask why you are restoring this old code ? AFAIK, the console autocompletion works (and I use it pretty much) and I've not seen any issue about that.

tomluchowski:
All  I can advocate for my code is it's much faster theoretically  than the brute / naive approach used currently by oln.
It uses Trees so searching the missing postfix is proportional to the length of the prefix plus length of each possible postfix ( completition ). It's independent of total number of commands. I know that for now it doesn't matter ( modern CPU speed ), but maybe in future when we have a several dozens of commands the result would be significant. Please give me one week time to adapt my code to requested format .... this PR could be closed then I guess.

hwoarangmy:
Well, IMHO, you are using a bazooka to kill a moskito. How many console commands do you expect in the end ? 50 ? I don't think your approach will have a noticeable gain.

Using this kind of trees can be usefull when lists are big. However, it makes the code more complicated (harder to understand and debug), slower at launch (because the tree has to be built) and usually very memory consuming (because you have to store many tokens for the tree - it is the case here since you build a new PrefixTree for each different letter for each command that starts with different chars).

Apart from that, I've had a look at your code and with just a quick read, I could see memory leaks (I guess the ones @oyvindln was talking about) and some minor improvements.
Here, you are trying to use a tree for console commands when we will have at most 50 commands and we will launch 1 or 2 max per game.

OD is already pretty memory consuming (trusting the memory manager, it is around 250 Mb for test model level which is not very big) and using even more memory for a side function like console commands is not a good thing IMHO.

IMHO, this PR should not be merged for the above reasons. If you want to work on improving technically the game, why don't you start it with a profiler to see what really needs to be improved and discuss it before poping with a PR (and try to not use some shadock unreadable algorithm) ?

tomluchowski:
And the current code uses if - breaks like crazy , is that correct structural programming ? Every block should have exactly well defined entry and an exit. 
Besides maybe my implementation would prove useful , maybe we would like to have console language like python or ruby , then it's dictionary of key-words can take quite a large list.
I feel bad that my work get wasted.
We are meritocracy aren;t we , maybe after polishing the code we could find an agreement  ?

hwoarangmy:
> And the current code uses if - breaks like crazy , is that correct structural programming ? Every block should have exactly well defined entry and an exit

Where does that statement comes from ? If you have reasons to not use break (AFAIK, it is a standard instruction from C language), please let us know.
FYI, before I came to OD project, I had the same believes due to my teachers saying that it is bad to use break (just like seeing a black cat or going under a ladder). But after thinking about it, it looks good to me. I'm interested in your point of view so if you have arguments against using break. I will give mine after ^^

> maybe we would like to have console language like python or ruby

Yes, maybe. But we should first decide if we want it and what for. But I don't think it will be used for console commands anyway and, in any case, it is not a good idea to go for something complex just because we might use it in the future.

> I feel bad that my work get wasted

Well, honestly, I feel bad about it too. But I don't want to merge stuff just because not doing it would make you feel bad.
Please keep in mind that in an open source project like this one, code readability is the most important thing we should focus on. It is much more important than execution speed (at long as the game is playable, of course).
That's why I believe we need to go for simple, naive algorithms for all our needs. And, if one proves to be a bottle neck and needs to be improved, then do it and comment as optimized code is always much less readable and straight forward.

As I already said, if OD code becomes too messy, not-understandable, nobody will want to work on it as soon as the current devs will leave. The lack of activity lately have also decreased my motivation and I will probably not be very active for some time (I've uploaded everything I had ready).
That's why I'm sorry that your code is wasted but I prefer that than wasting the whole project.

> We are meritocracy aren;t we , maybe after polishing the code we could find an agreement ?

Well, if you come with something good (ie readable and that improves the game), that follows code guidelines, I will be happy to merge it.
For example, since you seem to like technical stuff, if you went for some simple implementation for culling, it would be nice to add. The PR you created some time ago was much too complex.

tomluchowski:
Pity on Prefixtree and culling then . 
The culling algorithm was made quite readable in the sense I tried to create as much classes and functions for every partitionable process the algorithm involved. The only simplification I could do more is going for separate fixed point fractional class, which would hide many of the operations ( shifting back and forth by n bits ) from the code. Simply the exact tracing of visible tiles couldn't be done simpler. Or could be ? In mathematics useful functions are hidden everywhere. I coded with two obvious presumptions : 

1) Visible tiles should stay attached, unvisible tiles should be unattached.
2) On every frame our algorithm should be as lazy as possible : only change visibilty of tiles of which status changed since last frame ( it could be otherwise : like we might have disabled all tiles by default and then enable the visible ones etc. ) 

BTW: I don't fully understand the networking processing of OD, yet I didn't vote for not merging the following iterations of it into the code.

hwoarangmy:
> BTW: I don't fully understand the networking processing of OD, yet I didn't vote for not merging the following iterations of it into the code

Well, networking has been reviewed by @Bertram25 and @akien-mga and they choose to merge after usual comments/requirements. That means they found the code to follow the guidelines and to be readable enough. Moreover, if there were some stuff you didn't understand and you want to be explained, you could have done that when the PR was released or open an issue. I will gladly answer.
When you review code, you don't have to understand everything but at least the overall idea to check it goes well with the current architecture.

Concerning the culling PR, the main reason why it was not merged was because there was a big regression on the minimap and your suggestion of going back to the old one is not acceptable IMHO. Would you think it is the good way to go if someone creates a new room that, for some reason, breaks culling to deactivate culling ?
Because you asked for some help on this point, I had to go deeper in your code but couldn't follow it. If there was no regression, it would have been merged after testing. But since you couldn't/didn't want to fix the regresion and nobody popped to do it for you, the PR couldn't be merged. And that seems pretty logic to me.

> 1) Visible tiles should stay attached, unvisible tiles should be unattached.

Seems pretty logical if you don't want them to be displayed (while I'm not sure if calling setVisible is better or worst)

> 2) On every frame our algorithm should be as lazy as possible : only change visibilty of tiles of which status changed since last frame ( it could be otherwise : like we might have disabled all tiles by default and then enable the visible ones etc. )

Seems good on the paper. But IIRC, you also added a new memory system allocation and similar stuff. The rule of thumb when coding is first to make it work and only optimize what needs to be. OD was full of code more or less preparing for futur features that might never be implemented (AS, first person camera, ...). Doing that is a pretty bad habit has it adds complexity in the code for stuff that may be implemented in a different way when it will be (if it is ever).

akien-mga:
Also remember that we are developing OpenDungeons for fun, and we have high standards as to the code quality (so that we can attract good contributors, and have a good time coding ourselves, while minimizing the risk of hidden bugs).

As you stated it on the forum, your last PRs seem to be mostly so that you have some of your own code merged in the engine to advertise on your CV. It's fine per se, contribution to open source projects is a very good argument for a developer to put on his CV, but if it's your only motivation, it's unlikely that you'll reach our standards. We don't just push code to say "I was here", but to make OpenDungeons really better, fun to hack and play for developers and players alike.

So instead of just dropping some code that hasn't been discussed previously with the rest of the team, you should try to see what would really improve the game, discuss it with the team (especially the details of the implementation), and then code it. A PR that fixes an issue is much more likely to be useful than a PR that comes out of nowhere to replace a working command completion with a more complex, albeit unnecessary implementation (that does not even build, so you probably just restored your old code and did not try to adapt it to the current code base, or maybe did not even try to compile it yourself).

I'm sorry to be harsh, but we've been discussing this over and over again. Your current behaviour make all your code contributions hard to work with, and for a free time project we're all working on for fun, they are undesirable. So please try to comply with our workflow, or just create your own project to showcase that you can make advanced C++ code and put on your CV.

tomluchowski:
>  OD was full of code more or less preparing for future features that might never be implemented (AS, first person camera, ...).

Yes, my bad, when I started I did wanted to pull all strings at once, now I code it different, without the  presumptions of future's features . 

Maybe I could modify my code so it works for both cameras : the player's one and the minimap as well . 

Thanks for having a patience for me anyway hwoarangmy :) . 

> PS  (that does not even build, so you probably just restored your old code and did not try to adapt it to the current code base, or maybe did not even try to compile it yourself).

Well it did build, but I haven't modified the CMakeBuild.txt file on the server so it does not know what to do with PrefixTree.cpp .

hwoarangmy:
> Yes, my bad, when I started I did wanted to pull all strings at once, now I code it different, without the presumptions of future's features

We are not looking for the guilty, just pointing some things out.

> Well it did build, but I haven't modified the CMakeBuild.txt file on the server so it does not know what to do with PrefixTree.cpp

@akien-mga is pointing that there clearly is a problem in your work flow as the tools you use should prevent this kind of error. That means that you upload changes will your repository and your local sources are not in sync (I assume your code did compile on your local computer). It is a bad habit and you should not do that as you are likely to upload broken code as you did.

> Maybe I could modify my code so it works for both cameras : the player's one and the minimap as well

Yes. But please keep in mind that you will probably receive the same questions/comments as you had in the original PR.
For example:
- Choose meaningfull variable names (no "myArray" and such). Excepted iteration variables.
- You introduced a new class Vector with long variables. That looks very weird to me so expect questions of why you use it (why is it called Vector ? My guess is that it is more a list of 3D points than vectors. Why do you use long ? My guess is that it is not needed)
- Please do not use FastPODAllocator. If you need to allocate something, just use new, a vector or something adapted.
- You have to follow code guidelines
- hide/show for tiles should be done in the RenderManager, not in the entity. You are working on a rendering optimization. The tile should not know if it is displayed or not.

Be aware that we will require that to merge your PR. I don't want you to work again on something that won't be merged so if you don't feel like comforming, please do not start at all.

tomluchowski:
May I use the convex hull finding algorithm from here : [http://www.boost.org/doc/libs/1_56_0/libs/geometry/doc/html/geometry/reference/algorithms/convex_hull.html]  ? After finding the convex hull of the two polygons ( from camera projections ) the road to adapt algorithm to new circumstances should be straightforward .

hwoarangmy:
> May I use the convex hull finding algorithm from here

Looks interesting. I guess using boost should not be a problem. Do you know if it would imply to embed a new library ?

tomluchowski:
> kalven    paul424: geometry is header only afaik, so no linking required.

hwoarangmy:
Then I don't see any reason not to use it

PR #1188, opened 2016-05-17
AppVeyor: don't discard 7z log output

akien-mga:
We need it to debug current build issues.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -20,7 +20,7 @@ install:
           Write-Host "Downloading MinGW32"
           appveyor DownloadFile $urlDeps -FileName $file
           Write-Host "Unzip MinGW32"
-          7z x -oC:\dev $file > $null
+          7z x -oC:\dev $file
       }
 # download dependencies
   - ps: |
@@ -29,7 +29,7 @@ install:
       Write-Host "Downloading deps"
       appveyor DownloadFile $urlDeps -FileName $file
       Write-Host "Unzip deps"
-      7z x -oC:\ $file > $null
+      7z x -oC:\ $file
   - set OD_DEPS=C:\od-deps-win32
 # We try to keep as few things in path as possible to make sure cmake won't find a needed library in some unexpected place
   - if "%generator%" == "mingw32" (set "PATH=c:\dev\MinGW\bin;C:\od-deps-win32\bin\mingw48;")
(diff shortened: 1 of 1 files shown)


akien-mga:
Looks like we have a problem with the downloads from TF to AppVeyor, it fails from time to time (and more often than not):
- In @hwoarangmy's previous PR, it failed on the MSVC build when unzipping the od-deps-win32.7z archive, but MinGW built fine: https://ci.appveyor.com/project/hwoarangmy/opendungeons-crxx4/build/1.0.230/job/kbg6khbca8x0969h
- In this PR, MSVC built fine but MinGW failed when unzipping the MinGW32.7z archive: https://ci.appveyor.com/project/Bertram25/opendungeons/build/1.0.234/job/4ts54ddqxr9ehle8
  Looking at the newly added logs, it seems that the downloaded archive was 78 KiB, which is likely bogus and explains the extraction issue.

akien-mga:
Also, why do we have AppVeyor builds running on both @hwoarangmy and @Bertram25's namespaces? :)

akien-mga:
Maybe we could try downloading with https://www.appveyor.com/docs/how-to/download-file#curl to see if it behaves better.

akien-mga:
Why do we get two Travis CI builds on this PR btw?

akien-mga:
Well it looks like AppVeyor's instructions for curl are bogus, as it gives me an error as if it was using another command than curl, while I used their example verbatim:

```
Invoke-WebRequest : Parameter cannot be processed because the parameter name 'o' is ambiguous. Possible matches include: -OutFile -OutVariable -OutBuffer.
At line:7 char:16
+     curl -fsSL -o $file $urlDeps
+                ~~
    + CategoryInfo          : InvalidArgument: (:) [Invoke-WebRequest], ParameterBindingException
    + FullyQualifiedErrorId : AmbiguousParameter,Microsoft.PowerShell.Commands.InvokeWebRequestCommand
```

_Edit:_ Reported upstream at https://github.com/appveyor/ci/issues/812

akien-mga:
AppVeyor is still failing at random... curl is more explicit about the error though:

```
cmd : curl: (18) transfer closed with 34828454 bytes remaining to read
At line:4 char:1
+ cmd /c curl -fsSL -o $file $urlDeps
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (curl: (18) tran...maining to read:String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
```

I don't know if the issue is on TF's or AppVeyor's side, but one of the two seems to close the connection during the download.

As a dirty workaround, we could try to catch the curl error, and retry the download until we get an OK error code :D

akien-mga:
Closing as this did not improve the situation (though it helped me understand the issue better).
We should look into using a cache on AppVeyor so that we don't have to download dependencies each time.

PR #1189, opened 2016-05-18
Credits update

Danimal696:
Weapons ans shields credits


akien-mga:
Nice! I think the last missing piece is TortureObject, could you add it too?

PR #1195, opened 2016-05-31
Small fixes + Added new creatures to spawn pool

hwoarangmy:


hwoarangmy:
Seems like there is an error with one of the unit test. I will have a look and fix it ASAP

Danimal696:
Strange new rankings on tiers, but i cant talk :)
I hope i get some free time in the future and get to work on something else, work + uni is harder than i ever expected

hwoarangmy:
> Strange new rankings on tiers, but i cant talk :)

Well, I've just been trying to remove the temporary units and integrate the new creatures ^^

Loobinex:
Tier ranking is indeed a bit strange, but with the available models that can't be helped. Frankly, I think there are too many creatures that are suited to be very low tier, and not enough creatures that are suited for high tier.

Hornet, Spider, Rat, Roach and Goblin are thematically all weak creatures.
Orc, Elf, Lizardman, Cultist and Lavaspawn are all perfectly suited for a tier above.
The Pitdemon and Troll are the only powerful looking creatures and can't fill two tiers by themselves.
Furthermore, the Lizardman and Orc would fit a similar niche for me, perhaps the goblin as well: Scaly humanoid brute fighter. Having three different insects is also a bit much.

This is party because you're relying on models you can find, but it is also fundamentally difficult to have so many tiers and so few creature types on the lower tiers. When tier 4 has as many types as tier 2 it is hard to sell the unique power of the former.
If you look at other strategy games in general or DK in particular, you'll see this is done differently. In DK1 you could see Horny as T4; Dragon, Vamp, Bile and Mistress as T3; the insects and demon spawn as T1 and all the other 7 types as T2 creatures. A completely different ratio, but one that makes sense.(The DK1 problem is not the ratio but the fact that you can get the T3 creatures so easily.)

akien-mga:
Code looks good, and it builds and seems to work fine, let's merge :)
I'll make a Linux binary based on this merged PR.

hwoarangmy:
> I'll make a Linux binary based on this merged PR.

Ok, cool. This will be the same as the win32 snapshots I did

PR #1197, opened 2016-06-02
Refresh entities at end of turn instead of the beginning of next turn

hwoarangmy:
@akien-mga This should be merged before making the release as there is a visual glitch in the current dev repo


akien-mga:
Thanks, redoing my binaries then. Good that I hadn't uploaded them yet :D

PR #1198, opened 2016-06-03
Culling of both cameras

tomluchowski:
As in title , now the code due to boost::geometry::convex_hull can cull both camera views at once .

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -271,6 +271,9 @@ set(OD_SOURCEFILES
 
     ${SRC}/camera/CameraManager.cpp
     ${SRC}/camera/HermiteCatmullSpline.cpp
+    ${SRC}/camera/CullingManager.cpp
+    ${SRC}/camera/CullingVectorManager.cpp
+    ${SRC}/camera/SlopeWalk.cpp
 
     ${SRC}/creatureaction/CreatureAction.cpp
     ${SRC}/creatureaction/CreatureActionCarryEntity.cpp
@@ -503,6 +506,7 @@ set(OD_SOURCEFILES
     ${SRC}/utils/MasterServer.cpp
     ${SRC}/utils/Random.cpp
     ${SRC}/utils/ResourceManager.cpp
+    ${SRC}/utils/VectorInt64.cpp
 
     ${SRC}/ODApplication.cpp
     ${SRC}/main.cpp
--- a/source/camera/CameraManager.cpp
+++ b/source/camera/CameraManager.cpp
@@ -22,12 +22,16 @@
 #include "camera/CameraManager.h"
 
 #include "gamemap/TileContainer.h"
+#include "sound/SoundEffectsManager.h"
+#include "camera/CullingManager.h"
 #include "utils/LogManager.h"
+#include "gamemap/GameMap.h"
 #include <OgreCamera.h>
 #include <OgreSceneNode.h>
 #include <OgreSceneManager.h>
 #include <OgreViewport.h>
 #include <OgreRenderWindow.h>
+
 
 #include <algorithm>
 
@@ -47,7 +51,8 @@ const Ogre::Degree ROTATION_SPEED = Ogre::Degree(90);
 //! Default orientation on the X Axis
 const Ogre::Real DEFAULT_X_AXIS_VIEW = 25.0;
 
-CameraManager::CameraManager(Ogre::SceneManager* sceneManager, TileContainer* gm, Ogre::RenderWindow* renderWindow) :
+CameraManager::CameraManager(Ogre::SceneManager* sceneManager, GameMap* gm, Ogre::RenderWindow* renderWindow) :
+    mCullingManager(nullptr),
     mCircleMode(false),
     mCatmullSplineMode(false),
     mFirstIter(true),
@@ -72,6 +77,7 @@ CameraManager::CameraManager(Ogre::SceneManager* sceneManager, TileContainer* gm
     mSceneManager(sceneManager),
     mViewport(nullptr)
 {
+    mCullingManager = new CullingManager(this);
     createViewport(renderWindow);
     createCamera("RTS", 0.02, 300.0);
     createCameraNode("RTS");
@@ -95,6 +101,10 @@ void CameraManager::createCamera(const Ogre::String& ss, double nearClip, double
 
     mRegisteredCameraNames.insert(ss);
     OD_LOG_INF("Creating " + ss + " camera...");
+}
+
+CameraManager::~CameraManager(){
+    delete mCullingManager;
 }
 
 void CameraManager::createCameraNode(const std::string& name)
@@ -574,6 +584,28 @@ void CameraManager::move(const Direction direction, double aux)
     default:
         break;
     }
+}
+
+void CameraManager::startTileCulling()
+{
+    mCullingManager->startTileCulling();
+}
+
+void CameraManager::stopTileCulling()
+{
+    mCullingManager->stopTileCulling();
+}
+
+bool CameraManager::onFrameEnded()
+{
+     mCullingManager->onFrameEnded();
+     return true;
+}
+
+bool CameraManager::onFrameStarted()
+{
+     mCullingManager->onFrameStarted();
+     return true; 
 }
 
 bool CameraManager::isCameraMovingAtAll() const
--- a/source/camera/CameraManager.h
+++ b/source/camera/CameraManager.h
@@ -33,6 +33,9 @@
 #include <set>
 
 class TileContainer;
+class CullingManager;
+class GameMap;
+class TileContainer;
 
 // The min/max camera height in tile size
 const Ogre::Real MIN_CAMERA_Z = 3.0;
@@ -62,9 +65,8 @@ public:
         fullStop
     };
 
-    CameraManager(Ogre::SceneManager* sceneManager, TileContainer* gameMap, Ogre::RenderWindow* renderWindow);
-    ~CameraManager()
-    {}
+    CameraManager(Ogre::SceneManager* sceneManager, GameMap* gameMap, Ogre::RenderWindow* renderWindow);
+    ~CameraManager();
 
     inline void circleAround(int x, int y, unsigned int radius)
     {
@@ -86,15 +88,22 @@ public:
         return mTranslateVectorAccel;
     }
 
-    bool onFrameStarted()
-    { return true; }
-    bool onFrameEnded()
-    { return true; }
+
+    inline GameMap* getGameMap(){
+        return mGameMap;
+    }
+    bool onFrameStarted();
+    bool onFrameEnded();
+
+    CullingManager* mCullingManager;
 
     /*! \brief Sets the camera to a new location while still satisfying the
     * constraints placed on its movement
     */
     void updateCameraFrameTime(const Ogre::Real frameTime);
+
+    void startTileCulling();
+    void stopTileCulling();
 
     /*! \brief Computes a vector whose z-component is 0 and whose x-y coordinates
     * are the position on the floor that the camera is pointed at.
@@ -190,7 +199,8 @@ private:
     Ogre::Camera* mActiveCamera;
     Ogre::SceneNode* mActiveCameraNode;
 
-    TileContainer* mGameMap;
+
+    GameMap* mGameMap;
 
     //! \brief Is true when a camera is flying to a given position.
     bool            mCameraIsFlying;
--- a/source/camera/CullingManager.cpp
+++ b/source/camera/CullingManager.cpp
@@ -1,0 +1,246 @@
+/*
+ *  Copyright (C) 2011-2016  OpenDungeons Team
+ *
+ *  This program is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <OgreVector3.h>
+#include <OgreCamera.h>
+
+#include "camera/CullingManager.h"
+#include "camera/CameraManager.h"
+#include "gamemap/GameMap.h"
+
+#include "entities/Creature.h"
+#include "utils/VectorInt64.h"
+#include "utils/LogManager.h"
+#include "render/RenderManager.h"
+#include <sstream>
+#include <algorithm>
+
+//! Values used to know whether to show and/or hide a mesh
+
+CullingManager::CullingManager(CameraManager* cameraManager):
+    mFirstIter(false),
+    mCm(cameraManager),
+    mCullTilesFlag(false)
+{
+    mActivePlanes =(Ogre::Plane(0, 0, 1, 0));
+    // init the Ogre vector
+    for (unsigned int i = 0; i < 8; ++i)
+    {
+        mOgreVectorsArray[i].x = static_cast<Ogre::Real>(0.0);
+        mOgreVectorsArray[i].y = static_cast<Ogre::Real>(0.0);
+        mOgreVectorsArray[i].z = static_cast<Ogre::Real>(0.0);
+    }
+}
+
+void CullingManager::cullTiles()
+{
+    mOldWalk = mWalk;
+    mWalk.mVertices.mMyArray.clear();
+    for (int ii = 0 ; ii < 8 ; ++ii)
+        mWalk.mVertices.mMyArray.push_back(VectorInt64(mOgreVectorsArray[ii]));
+
+    // create a slope -- a set of left and rigth path
+    mWalk.convex_hull();
+    mWalk.buildSlopes();
+
+
+    OD_LOG_DBG( mOldWalk.debug());
+    OD_LOG_DBG( mWalk.debug());
+
+    // reset index pointers to the begging of collections
+    mOldWalk.prepareWalk();
+    mWalk.prepareWalk();
+
+    newBashAndSplashTiles(SHOW | HIDE);
+}
+
+void CullingManager::startTileCulling()
+{
+    getIntersectionPoints();
+
+    mWalk.mVertices.mMyArray.clear();
+    for (int ii = 0 ; ii < 8 ; ++ii)
+        mWalk.mVertices.mMyArray.push_back(VectorInt64(mOgreVectorsArray[ii]));
+    mWalk.convex_hull();
+    mWalk.buildSlopes();
+    mOldWalk = mWalk;
+    mOldWalk.prepareWalk();
+    mWalk.prepareWalk();
+    hideAllTiles();
+    newBashAndSplashTiles(SHOW);
+
+    mCullTilesFlag = true;
+}
+
+
+void CullingManager::stopTileCulling()
+{
+    mCullTilesFlag = false;
+    mOldWalk = mWalk;
+    mWalk.mVertices.mMyArray.clear();
+    for (int ii = 0 ; ii < 8 ; ++ii)
+        mWalk.mVertices.mMyArray.push_back(VectorInt64(mOgreVectorsArray[ii]));
+
+    // create a slope -- a set of left and rigth path
+    mWalk.convex_hull();
+    mWalk.buildSlopes();
+
+    OD_LOG_DBG( mOldWalk.debug());
+    OD_LOG_DBG( mWalk.debug());        
+
+    // reset index pointers to the begging of collections
+    mOldWalk.prepareWalk();
+    mWalk.prepareWalk();
+
+    newBashAndSplashTiles(HIDE);
+    showAllTiles();
+}
+
+void CullingManager::hideAllTiles(void)
+{
+    GameMap* gm = mCm->getGameMap();
+    if (!gm)
+        return;
+
+    for (int jj = 0; jj < gm->getMapSizeY() ; ++jj)
+    {
+        for (int ii = 0; ii < gm->getMapSizeX(); ++ii)
+        {
+            RenderManager::getSingleton().rrDetachEntity(gm->getTile(ii, jj));
+        }
+    }
+}
+
+void CullingManager::showAllTiles(void)
+{
+    GameMap* gm = mCm->getGameMap();
+    if (!gm)
+        return;
+
+    for (int jj = 0; jj < gm->getMapSizeY() ; ++jj)
+    {
+        for (int ii = 0; ii < gm->getMapSizeX(); ++ii)
+        {
+            RenderManager::getSingleton().rrAttachEntity(gm->getTile(ii, jj));
+        }
+    }
+}
+
+void CullingManager::newBashAndSplashTiles(uint32_t mode){
+    int64_t xxLeftOld = mOldWalk.getTopLeftVertex().x;
+    int64_t xxRightOld= mOldWalk.getTopRightVertex().x;
+    int64_t xxLeft = mWalk.getTopLeftVertex().x;
+    int64_t xxRight= mWalk.getTopRightVertex().x;
+    int64_t xxp, yyp;
+    std::stringstream ss;
+    int64_t bb = (( std::min(mWalk.getBottomLeftVertex().y , mOldWalk.getBottomRightVertex().y) >> VectorInt64::PRECISION_DIGITS) - 2) << VectorInt64::PRECISION_DIGITS;
+
+    for (int64_t yy = ((std::max(mWalk.getTopLeftVertex().y , mOldWalk.getTopRightVertex().y  ) >> VectorInt64::PRECISION_DIGITS) + 2) << VectorInt64::PRECISION_DIGITS;  yy >= bb; yy -= VectorInt64::UNIT)
+    {
+        mOldWalk.notifyOnMoveDown(yy);
+        mWalk.notifyOnMoveDown(yy);
+        xxLeft = mWalk.getCurrentXLeft(yy);
+        xxLeftOld = mOldWalk.getCurrentXLeft(yy);
+        xxRight = mWalk.getCurrentXRight(yy);
+        xxRightOld = mOldWalk.getCurrentXRight(yy);
+   
+        int64_t mm = ((std::min(xxLeft, xxLeftOld) >> VectorInt64::PRECISION_DIGITS) << VectorInt64::PRECISION_DIGITS) ;
+        if(std::min(xxLeft, xxLeftOld) < std::max(xxRight,xxRightOld) )
+        {
+            for (int64_t xx = mm ; xx <= std::max(xxRight,xxRightOld) ; xx+= VectorInt64::UNIT)
+            {
+                bool bash = (xx >= xxLeftOld && xx <= xxRightOld && (yy >= mOldWalk.getBottomLeftVertex().y) && yy <= mOldWalk.getTopLeftVertex().y);
+                bool splash = (xx >= xxLeft && xx <= xxRight && (yy >= mWalk.getBottomLeftVertex().y) && yy <= mWalk.getTopLeftVertex().y);
+
+                xxp = ((xx >>VectorInt64::PRECISION_DIGITS));
+                yyp = ((yy >>VectorInt64::PRECISION_DIGITS));
+                GameMap* gm = mCm->getGameMap();
+                if(bash && splash && (mode & HIDE) && (mode & SHOW))
+                {
+                    // Nothing
+                }
+
+                else if (gm && bash && (mode & HIDE) && xxp>=0 && yyp>= 0 && xxp<mCm->getGameMap()->getMapSizeX() && yyp < mCm->getGameMap()->getMapSizeY())
+                {
+                    RenderManager::getSingleton().rrDetachEntity(gm->getTile(xxp, yyp));
+                }
+                else if (gm && splash && (mode & SHOW) && xxp>=0 && yyp>= 0 && xxp<mCm->getGameMap()->getMapSizeX() && yyp < mCm->getGameMap()->getMapSizeY())
+                {
+                    RenderManager::getSingleton().rrAttachEntity(gm->getTile(xxp, yyp));
+                }
+            }
+        }
+    }
+}
+
+bool CullingManager::getIntersectionPoints()
+{
+    const Ogre::Vector3* activeCameraVector = mCm->getActiveCamera()->getWorldSpaceCorners();
+    const Ogre::Vector3* activeCameraVector2 = mCm->getCamera("miniMapCam")->getWorldSpaceCorners();
+    for(int ii = 0 ; ii < 4; ++ii)
+        mActiveRay[ii]= Ogre::Ray (activeCameraVector[ii], activeCameraVector[ii+4] - activeCameraVector[ii]);
+
+    for(int ii = 4; ii < 8; ++ii)
+        mActiveRay[ii]= Ogre::Ray (activeCameraVector2[ii-4], activeCameraVector2[ii] - activeCameraVector2[ii-4]);
+
+    std::pair<bool, Ogre::Real> intersectionResult;
+
+    for(int ii = 0; ii < 8; ++ii)
+    {
+        intersectionResult =  mActiveRay[ii].intersects(mActivePlanes);
+        if(intersectionResult.first)
+            mOgreVectorsArray[ii]= (mActiveRay[ii].getPoint(intersectionResult.second));
+        else
+        {
+            OD_LOG_ERR("I didn't find the intersection point for " + Helper::toString(ii) + "th ray ");
+        }
+    }
+    return true;
+}
+
+bool CullingManager::onFrameStarted()
+{
+    if(mCullTilesFlag)
+        getIntersectionPoints();
+    if(mCullTilesFlag)
+        cullTiles();
+    return true;
+}
+
+bool CullingManager::onFrameEnded()
+{
+    return true;
+}
+
+/*! \brief Sort two VectorInt64 p1 and p2  to satisfy p1 <= p2 according to
+ * the value of X or Y coordiante, which depends on sortByX param .
+ */
+void CullingManager::sort(VectorInt64& p1, VectorInt64& p2, bool sortByX)
+{
+    if (sortByX)
+    {
+        if (p1.x > p2.x)
+            std::swap(p1, p2);
+    }
+    else
+    {
+        if (p1.y > p2.y)
+            std::swap(p1, p2);
+    }
+}
+
+
--- a/source/camera/CullingManager.h
+++ b/source/camera/CullingManager.h
@@ -1,0 +1,97 @@
+/*
+ *  Copyright (C) 2011-2016  OpenDungeons Team
+ *
+ *  This program is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef CULLINGMANAGER_H_
+#define CULLINGMANAGER_H_
+#include "camera/CameraManager.h"
+#include "camera/SlopeWalk.h"
+#include "entities/Tile.h"
+
+#include "utils/VectorInt64.h"
+
+#include <OgreRay.h>
+#include <OgrePlane.h>
+#include <set>
+
+class CameraManager;
+class GameMap;
+
+/*! \brief The CullingMangaer class is a class to effectivly 
+ *  manage culling methods used in game. So far there is only
+ *  one algorithm included : it is supposed to cull the Tiles.
+ *  It should be started with the method startTileCulling.
+ *  In future I plan to add creature's culling as well. 
+ 
+ * Currently I try to impment the most general polygon rasterizing algorithm , that is :
+ * 1. We choose the max and min points due to their Y value .
+ * 2. We sort all the points by the angle value in it's polar representation 
+ * ( that is going , by visiting them clockwise due to the center of the polygon )
+ * 3. Now we have two paths from which we can walk from top to down vertices : call them left and right .
+ * 4. Between each following pair of vertices we can establish a >> slope << ,
+ * which is just the the "a" in the eq. of linear form y = ax + b
+ * 5. Now start drawing our polygon Row by Row From top to bottom . 
+ * Each Row has given the most left and rightmost tile due to use of both paths prepared before 
+ * -- Left path for tracing the most Leftmost Tile , Right path teh most Rightmost Tile in each .
+ */
+class CullingManager
+{
+public:
+    static const uint32_t HIDE =  1;
+    static const uint32_t SHOW =  2;
+
+    CullingManager(CameraManager*);
+
+    void startTileCulling();
+    void startDebugging();
+
+    void stopTileCulling();
+
+    bool getIntersectionPoints();
+
+    void hideAllTiles();
+    void showAllTiles();
+
+    void cullTiles();
+
+    bool onFrameStarted();
+    bool onFrameEnded();
+
+    // set the new tiles
+    void newBashAndSplashTiles(uint32_t);
+
+    void sort(VectorInt64& p1, VectorInt64& p2, bool sortByX);
+ 
+private:
+
+    // Objects representing past and present walk around the polygon
+    SlopeWalk mWalk;
+    SlopeWalk mOldWalk;
+
+    // Array of Vector3's for keeping the intersection points of camera viewfrustrum 
+    // and the XY plane
+    Ogre::Vector3 mOgreVectorsArray[8];
+ 
+    bool mFirstIter;
+    CameraManager* mCm;
+
+    Ogre::Plane mActivePlanes;
+    Ogre::Ray mActiveRay[8];
+
+    bool mCullTilesFlag;
+};
+
+#endif // CULLINGMANAGER_H_
--- a/source/camera/CullingVectorManager.cpp
+++ b/source/camera/CullingVectorManager.cpp
@@ -1,0 +1,101 @@
+/*
+ *  Copyright (C) 2011-2016  OpenDungeons Team
+ *
+ *  This program is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "camera/CullingVectorManager.h"
+
+#include <algorithm>
+
+void CullingVectorManager::sort()
+{
+    std::vector<VectorInt64> ll[4];
+    VectorInt64 cc(0, 0, 0);
+
+    // find the center of points stored in mMyArray
+    for (auto ii : mMyArray)
+        cc = cc + ii;
+
+    cc = cc / mMyArray.size();
+
+    //Place the points into the appropriate quadrants of the (2D) Cartesian coordinate system
+    for(unsigned int ii = 0; ii < mMyArray.size() ; ++ii)
+    {
+        if(mMyArray[ii].x > cc.x && mMyArray[ii].y > cc.y)
+        {
+            ll[0].push_back(mMyArray[ii]);
+        }
+        if(mMyArray[ii].x > cc.x && mMyArray[ii].y <= cc.y)
+        {
+            ll[1].push_back(mMyArray[ii]);
+        }
+        if(mMyArray[ii].x <= cc.x && mMyArray[ii].y <= cc.y)
+        {
+            ll[2].push_back(mMyArray[ii]);
+        }
+        if(mMyArray[ii].x <= cc.x && mMyArray[ii].y > cc.y)
+        {
+            ll[3].push_back(mMyArray[ii]);
+        }
+    }
+
+    int kk = 0;
+    // sort each quadrant by tangens of angle of point in polar representation
+    for(int ii = 0; ii < 4; ++ii)
+    {
+        sortByPhi(ll[ii], cc);
+
+        for(std::vector<VectorInt64>::const_iterator jj = ll[ii].begin() ; jj != ll[ii].end(); ++jj)
+        {
+            mMyArray[kk] = *jj;
+            ++kk;
+        }
+    }
+}
+
+void CullingVectorManager::zoom(double zz)
+{
+    VectorInt64 cc(0, 0, 0);
+
+    for (auto& ii : mMyArray)
+        cc = cc + ii;
+    
+    cc = cc / mMyArray.size();
+    for (auto& ii : mMyArray)    
+    {
+        ii = cc + (ii - cc)*zz;    
+    }
+}
+
+void CullingVectorManager::sortByPhi(std::vector<VectorInt64> &ll, VectorInt64& cc)
+{
+    if (ll.size() < 2)
+        return;
+
+    std::sort(ll.begin(), ll.end(),
+              [&cc](VectorInt64 vv, VectorInt64 ww)
+              {
+                  return static_cast<double>(vv.y - cc.y) / static_cast<double>(vv.x - cc.x) > static_cast<double>(ww.y - cc.y) / static_cast<double>(ww.x - cc.x);
+              }
+        );
+}
+
+VectorInt64& CullingVectorManager::operator[](int ii)
+{
+    while(ii < 0)
+        ii += mMyArray.size();
+
+    return mMyArray[ii % mMyArray.size()];
+}
--- a/source/camera/SlopeWalk.cpp
+++ b/source/camera/SlopeWalk.cpp
@@ -1,0 +1,294 @@
+/*
+ *  Copyright (C) 2011-2016  OpenDungeons Team
+ *
+ *  This program is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "camera/SlopeWalk.h"
+#include "utils/LogManager.h"
+#include "utils/Helper.h"
+
+#include <algorithm>
+#include <iostream>
+#include <sstream>
+
+#include <boost/geometry.hpp>
+#include <boost/geometry/geometries/point_xy.hpp>
+#include <boost/geometry/geometries/polygon.hpp>
+#include <boost/geometry/geometries/register/point.hpp>
+#include <boost/geometry/geometries/adapted/boost_tuple.hpp>
+
+BOOST_GEOMETRY_REGISTER_POINT_2D(VectorInt64, int64_t, boost::geometry::cs::cartesian, x, y)
+//! Values used to know whether to show and/or hide a mesh
+typedef VectorInt64 Point;
+typedef boost::geometry::model::polygon<Point> Polygon;
+
+template <typename Point>
+void get_coordinates(Point const& p)
+{
+    OD_LOG_DBG( "get__coordinates : " + Helper::toString(double(boost::geometry::get<0>(p))/VectorInt64::UNIT) + ", " + Helper::toString( double(boost::geometry::get<1>(p))/VectorInt64::UNIT )) ;
+}
+
+void SlopeWalk::buildSlopes()
+{
+    //A value of how much we are to enlarge our polygon
+    //choosen experimentally too large values slows down
+    //the culling and may provide some glitches
+    mVertices.sort();
+    mRightSlopes.clear();
+    mLeftSlopes.clear();
+    mRightVertices.clear();
+    mLeftVertices.clear();
+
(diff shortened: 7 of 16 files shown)


hwoarangmy:
@tomluchowski Could you please merge the commits so that there is only 1 functionality per commit ?
As I already said, splitting into several commits is a good habit but each commit should represent one functionality. For example, you could have one for culling in the main map and one in the minimap.
Then, if there are code changes requested, you can add commits to adapt (to avoid too much editing).

But with PRs like this one, history of the files will become a mess in only a few days and reading code to understand better what's going on when looking for some regression becomes not possible.
I will review the code when the commits are merged. In the meanwhile, I will test to see how it behaves.

tomluchowski:
: naah by choice I appended to the akien's mga " Add changelog for 0.7.0"
Hopefully it's not a great blunder.

hwoarangmy:
"Add changelog for 0.7.0" is not the best name I could think about for such a commit. Please rename it so the history is not broken. Keep in mind that if we look for file history, commits must be a minimum meaningfull. And in this case, it is missleading as I wouldn't expect such a commit to be culling code.

hwoarangmy:
It seems also like your PR do not compile

akien-mga:
Also I appear as author, and it contains my whole changelog additions, which are already in the development branch.

I would do:

```
git format-patch -1  // keep the 0001-*.patch file around
git remote add upstream https://github.com/OpenDungeons/OpenDungeons
git fetch upstream
git reset --hard upstream/development
// edit the patch to remove the part about RELEASE-NOTES.md
git apply 0001-*.patch  // reapply the patch, without committing, solve the conflict with RELEASE-NOTES.md
git status  // check that the changes are fine
git commit -a "Good name for a commit about culling"
```

tomluchowski:
Well just look why travis is broken , it cannot upgrade it's repos : 
`The command "sudo apt-get update -qq; sudo apt-get install -y libcegui-0.8-dev" failed and exited with 100 during .`

I will fix my commit to meaningful name when I am back home.

hwoarangmy:
yes but appveyor complains at least of constexpr. Since we also support MSVS2013 and MinGW compilers, we must not use not supported c++11 features.

And since you are going to redo your commit, also note that you have uploaded a file EntityBase.cpp and .h that are not in use anymore.

tomluchowski:
BTW : Could you and akien-mga relate to the issues I posted on our forum  ?

tomluchowski:
ok , hopefully I am done .

CMakeLists.txt
```diff
@@ -271,6 +271,9 @@ set(OD_SOURCEFILES
 
     ${SRC}/camera/CameraManager.cpp
     ${SRC}/camera/HermiteCatmullSpline.cpp
+    ${SRC}/camera/CullingManager.cpp
+    ${SRC}/camera/DummyArrayClass.cpp
```
akien-mga: This file does not exist, so the build fails:

```
CMake Error at CMakeLists.txt:652 (add_executable):
  Cannot find source file:
    /home/travis/build/OpenDungeons/OpenDungeons/source/camera/DummyArrayClass.cpp
  Tried extensions .c .C .c++ .cc .cpp .cxx .m .M .mm .h .hh .h++ .hm .hpp
  .hxx .in .txx
```

source/camera/CameraManager.h
```diff
@@ -62,7 +66,7 @@ class CameraManager
         fullStop
     };
 
-    CameraManager(Ogre::SceneManager* sceneManager, TileContainer* gameMap, Ogre::RenderWindow* renderWindow);
+    CameraManager(Ogre::SceneManager* sceneManager, GameMap* gameMap, Ogre::RenderWindow* renderWindow);
```
hwoarangmy, line 68: Why did you change TileContainer to GameMap ? TileContainer is not a very useful abstract level since it is almost not used but why did you change ?

source/camera/CameraManager.h
```diff
@@ -48,6 +51,7 @@ enum class ViewModes : uint16_t
 
 class CameraManager
 {
+    friend class CullingManager;
```
hwoarangmy: Please do not use friendship unless there is a real reason. It makes search and code reading not straight forward. If you need access to some of the CameraManager variables, just add the corresponding getters/setters so that we know at first glance if a private variable is private or not.

source/camera/CameraManager.h
```diff
@@ -86,16 +90,20 @@ class CameraManager
...
+    bool onFrameEnded();
+
+    CullingManager* mCullingManager;
 
     /*! \brief Sets the camera to a new location while still satisfying the
     * constraints placed on its movement
     */
     void updateCameraFrameTime(const Ogre::Real frameTime);
 
+    void startCreatureCulling();
```
hwoarangmy: From what I've seen creatures are not culled yet. Please do not add the functions if they are not implemented. A simple TODO would be better if you don't want to forget. Moreover, it makes sense to split the commits in this case.
When creature culling will be implemented, it will be the moment to add the function. That will help to keep history clean rather than creating some functions in a commit and implementing them in some later one.

source/camera/CullingManager.cpp
```diff
@@ -0,0 +1,292 @@
...
+    for (int ii = 0 ; ii < 8 ; ++ii)
+        mWalk.mVertices.mMyArray.push_back(Vector3i(mOgreVectorsArray[ii]));
+
+    // create a slope -- a set of left and rigth path
+    mWalk.convex_hull();
+    mWalk.buildSlopes();
+
+    if(mDebug)
+    {
+        LogManager::getSingleton().logMessage(LogMessageLevel::NORMAL,__FILE__,__LINE__ ,mOldWalk.debug());
```
hwoarangmy: Please use OD_LOG_INF for corresponding level for logs. In this case, if you want debug logs, it would be good to use OD_LOG_DBG (instead of the flag).

source/camera/CullingManager.cpp
```diff
@@ -0,0 +1,292 @@
...
+{
+    GameMap* gm = mCm->mGameMap;
+    if (!gm)
+        return;
+
+    for (int jj = 0; jj < gm->getMapSizeY() ; ++jj)
+    {
+        for (int ii = 0; ii < gm->getMapSizeX(); ++ii)
+        {
+            gm->getTile(ii,jj)->hide();
```
hwoarangmy: Culling is a pure client side function. Calling the render manager here would be better than adding a proxy in the game entity class.

source/camera/CullingManager.cpp
```diff
@@ -0,0 +1,292 @@
...
+    for (int64_t yy = ((std::max(mWalk.getTopLeftVertex().y , mOldWalk.getTopRightVertex().y  ) >> mPrecisionDigits) + 2) << mPrecisionDigits;  yy >= bb; yy -= Unit)
+    {
+        mOldWalk.notifyOnMoveDown(yy);
+        mWalk.notifyOnMoveDown(yy);
+        xxLeft = mWalk.getCurrentXLeft(yy);
+        xxLeftOld = mOldWalk.getCurrentXLeft(yy);
+        xxRight = mWalk.getCurrentXRight(yy);
+        xxRightOld = mOldWalk.getCurrentXRight(yy);
+   
+        // if(mDebug)
```
hwoarangmy: Please do not let dead code. If there is a need, add a debug log. If not, just remove it.

BTW, you should not use streams for logs. It is much better to use OD_LOG_DBG macro because if the log level is not high enough, the message will not be computed.

source/camera/CullingManager.cpp
```diff
@@ -0,0 +1,292 @@
...
+                }
+                xxp = ((xx >>mPrecisionDigits));
+                yyp = ((yy >>mPrecisionDigits));
+                GameMap* gm = mCm->mGameMap;
+                if(bash && splash && (mode & HIDE) && (mode & SHOW))
+                {
+                    // Nothing
+                }
+
+                else if (gm && bash && (mode & HIDE) && xxp>=0 && yyp>= 0 && xxp<mCm->mGameMap->getMapSizeX() && yyp < mCm->mGameMap->getMapSizeY())
```
hwoarangmy: Can the gamemap really be null ?
Moreover, the check you are doing here is also done in GameMap::getTile. It would be better to check if the tile returned by GameMap::getTile is null.

source/camera/CullingManager.cpp
```diff
@@ -0,0 +1,292 @@
...
+
+                if (mDebug)
+                {
+                    ss<< " x " <<  (xx >> mPrecisionDigits )<< " y " << (yy >> mPrecisionDigits ) << " " << (splash && (mode & SHOW)) << (bash && (mode & HIDE)) << std::endl;
+            
+                }
+                xxp = ((xx >>mPrecisionDigits));
+                yyp = ((yy >>mPrecisionDigits));
+                GameMap* gm = mCm->mGameMap;
+                if(bash && splash && (mode & HIDE) && (mode & SHOW))
```
hwoarangmy, line 172: Can the mode be HIDE and SHOW ? That's sounds pretty weird.

source/camera/CullingManager.cpp
```diff
@@ -0,0 +1,292 @@
...
+                    gm->getTile(xxp, yyp)->show();
+                }
+            }
+        }
+    }
+}
+
+bool CullingManager::getIntersectionPoints()
+{
+    const Ogre::Vector3* myvector = mCm->getActiveCamera()->getWorldSpaceCorners();
```
hwoarangmy: please avoid the my... as variable name. Here, activeCameraVector or at least actCamVector sounds better.

source/camera/CullingManager.cpp
```diff
@@ -0,0 +1,292 @@
...
+    std::pair<bool, Ogre::Real> intersectionResult;
+
+    for(int ii = 0; ii < 8; ++ii)
+    {
+        intersectionResult =  mMyRay[ii].intersects(mMyplanes);
+        if(intersectionResult.first)
+            mOgreVectorsArray[ii]= (mMyRay[ii].getPoint(intersectionResult.second));
+        else
+        {
+            std::cerr<< "I didn't find the intersection point for " << ii <<"th ray"<<std::endl;
```
hwoarangmy: Please use OD_LOG_ERR for errors.

source/camera/CullingManager.h
```diff
@@ -0,0 +1,104 @@
...
+#include "utils/Vector3i.h"
+
+#include <OgreRay.h>
+#include <OgrePlane.h>
+#include <set>
+
+class CameraManager;
+class GameMap;
+
+/*! \brief The CullingMangaer class is a class to effectivly
```
hwoarangmy, line 33: Some typos but overall, it helps understanding what you are trying to do. That's good.

source/camera/CullingManager.h
```diff
@@ -0,0 +1,104 @@
...
+ * 3. Now we have two paths from which we can walk from top to down vertices : call them left and right .
+ * 4. Between each following pair of vertices we can establish a >> slope << ,
+ * which is just the the "a" in the eq. of linear form y = ax + b
+ * 5. Now start drawing our polygon Row by Row From top to bottom . 
+ * Each Row has given the most left and rightmost tile due to use of both paths prepared before 
+ * -- Left path for tracing the most Leftmost Tile , Right path teh most Rightmost Tile in each .
+ */
+class CullingManager
+{
+    friend class GameMap;
```
hwoarangmy: Please avoid friendship again.

source/camera/CullingManager.h
```diff
@@ -0,0 +1,104 @@
...
+    bool onFrameStarted();
+    bool onFrameEnded();
+
+    // set the new tiles
+    void newBashAndSplashTiles(int64_t);
+
+    void sort(Vector3i& p1, Vector3i& p2, bool sortByX);
+
+    //! \brief Activate/deactivate debug output.
+    bool mDebug;
```
hwoarangmy: As I said, it is better to use OD_LOG_DBG rather than a boolean.

source/camera/CullingManager.h
```diff
@@ -0,0 +1,104 @@
...
+
+    // Objects representing past and present walk around the polygon
+    SlopeWalk mWalk;
+    SlopeWalk mOldWalk;
+
+    // Array of Vector3's for keeping the intersection points of camera viewfrustrum 
+    // and the XY plane
+    Ogre::Vector3 mOgreVectorsArray[8];
+ 
+    bool mFirstIter;
```
hwoarangmy: Sounds like a bad name

source/camera/CullingManager.h
```diff
@@ -0,0 +1,104 @@
...
+    CameraManager* mCm;
+
+    Ogre::Plane mMyplanes;
```
hwoarangmy: Please avoid the my... variable name.

source/camera/DummyArrayClass.h
```diff
@@ -0,0 +1,42 @@
...
+#include <array>
+#include <vector>
+
+/*! \brief The DummyArrayClass class is a wrapper 
+ *  around the std::array mainly to overload the 
+ *  operator [].        
+ */
+
+
+class DummyArrayClass
```
hwoarangmy: DummyArray sound like a bad name. Be it a wrapper with only a few changes or something that have additional features, it is the vector manager for the culling. So, CullingVectorManager makes more straight forward what this class is used for than DUmmyArrayClass.

source/camera/SlopeWalk.cpp
```diff
@@ -0,0 +1,321 @@
...
+#include <iostream>
+#include <sstream>
+
+#include <boost/geometry.hpp>
+#include <boost/geometry/geometries/point_xy.hpp>
+#include <boost/geometry/geometries/polygon.hpp>
+#include <boost/geometry/geometries/register/point.hpp>
+#include <boost/geometry/geometries/adapted/boost_tuple.hpp>
+
+using namespace boost::geometry;
```
hwoarangmy: In most of our code, we don't use using. It is better to use the complete name when calling boost functions to make easier to identify them.

source/camera/SlopeWalk.cpp
```diff
@@ -0,0 +1,321 @@
...
+typedef model::polygon<Point> Polygon;
+
+template <typename Point>
+void get_coordinates(Point const& p)
+{
+    using boost::geometry::get;
+    std::stringstream ss;
+    ss << "get__coordinates : " ; 
+    ss << double(get<0>(p))/Unit << ", " << double(get<1>(p))/Unit << std::endl;
+    LogManager::getSingleton().logMessage(LogMessageLevel::NORMAL,__FILE__,__LINE__,ss.str());
```
hwoarangmy: Again, using OD_LOG_DBG without using a stream would be better

source/camera/SlopeWalk.cpp
```diff
@@ -0,0 +1,321 @@
...
+
+SlopeWalk::SlopeWalk():
+    mLeftVertexPassed(false),
+    mRightVertexPassed(false)
+{
+}
+
+void SlopeWalk::buildSlopes()
+{
+    int ii;
```
hwoarangmy: you should define such local variables in the for loops

source/camera/SlopeWalk.cpp
```diff
@@ -0,0 +1,321 @@
...
+
+    // find minimal and maximal values of y coordinate at left and right path
+    findMinMaxLeft(mVertices.mMyArray);
+    findMinMaxRight(mVertices.mMyArray);
+
+    // for each pair of consecutive vertexes on right path create a slope value , which is equal to 'a' as in equation ax + b = 0
+    // also put slopes of value 0 at the begging and the end of path
+
+    mRightSlopes.push_back(0);
+    for(ii = mTop_right_index; ii != mDown_right_index ; ++ii,ii%=mVertices.mMyArray.size()  )
```
hwoarangmy: Wow, that's pretty unreadable...

source/camera/SlopeWalk.h
```diff
@@ -0,0 +1,106 @@
...
+ */
+
+class SlopeWalk
+{
+
+    friend class CullingManager;
+
+private:
+
+    std::deque<long long> mRightSlopes;
```
hwoarangmy: it would be better to use sized variables such as uint32_t when needed as long long and such are plateform dependant and may create unexpected bugs.

source/camera/SlopeWalk.h
```diff
@@ -0,0 +1,106 @@
...
+    std::deque<long long> mLeftSlopes;
+    std::deque<long long>::iterator mRightSlopeIndex;
+    std::deque<long long>::iterator mLeftSlopeIndex;
+
+
+    // index numbers in myArray of Vertices 
+    // belonging to the right path
+    std::deque<int> rightVertices;
```
hwoarangmy: please follow code guidelines

source/camera/SlopeWalk.h
```diff
@@ -0,0 +1,106 @@
...
+    std::deque<int> leftVertices;
+    // point the Vertex currently processed 
+    // on the right path
+    std::deque<int>::iterator rightVerticesIndex;
+    // pint the Vertex currenly processed
+    // on the left path
+    std::deque<int>::iterator leftVerticesIndex;
+    bool passLeftVertex();
+    bool passRightVertex();
+    int mDown_right_index;
```
hwoarangmy: please do not use underscores for variable names

source/entities/GameEntity.h
```diff
@@ -452,6 +455,7 @@ class GameEntity
 
     void fireEntityRemoveFromGameMap();
 
+
```
hwoarangmy: please do not add useless empty lines

source/gamemap/MiniMapCamera.cpp
```diff
@@ -50,7 +50,7 @@ static const Ogre::Real NB_TILES_DISPLAYED_IN_MINIMAP = 30.0;
 static const Ogre::Radian ANGLE_CAM = Ogre::Degree(90.0);
 static const Ogre::Real CAM_HEIGHT = NB_TILES_DISPLAYED_IN_MINIMAP * 0.5 / Ogre::Math::Tan(ANGLE_CAM * 0.5);
 static const Ogre::uint TEXTURE_SIZE = 512;
-static const Ogre::Real MIN_TIME_REFRESH_SECS = 0.5;
+static const Ogre::Real MIN_TIME_REFRESH_SECS = 0.05;
```
hwoarangmy, line 53: this may hurt pretty bad game performances to increase this much refresh rate. The best would be to somehow try to know if the camera is moving and refresh more often if it is the case

source/gamemap/TileContainer.h
```diff
@@ -21,6 +21,7 @@
 #include <cassert>
 #include <list>
 #include <vector>
+#include "entities/Tile.h"
```
hwoarangmy: Looks like an error

source/modes/ConsoleCommands.cpp
```diff
@@ -530,6 +536,11 @@ void addConsoleCommands(ConsoleInterface& cl)
                   cSrvAddCreature,
                   {AbstractModeManager::ModeType::GAME, AbstractModeManager::ModeType::EDITOR});
 
+    // cl.addCommand("startTileCulling",
```
hwoarangmy: Why do you add a command commented ? If it is usefull, just let it be. If not, then just remove it.

source/network/ODClient.cpp
```diff
@@ -46,6 +46,8 @@
 #include "utils/Helper.h"
 #include "utils/LogManager.h"
 #include "ODApplication.h"
+#include "camera/CullingManager.h"
```
hwoarangmy: Looks like an error

source/render/ODFrameListener.cpp
```diff
@@ -38,6 +38,7 @@
 #include "utils/Helper.h"
 #include "utils/LogManager.h"
 #include "utils/MakeUnique.h"
+#include "camera/CullingManager.h"
```
hwoarangmy: looks like an error

source/render/RenderManager.cpp
```diff
@@ -29,6 +29,7 @@
 #include "entities/RenderedMovableEntity.h"
 #include "entities/Tile.h"
 #include "entities/Weapon.h"
+
```
hwoarangmy: Why did you add an emty line?

source/render/RenderManager.cpp
```diff
@@ -665,6 +666,19 @@ void RenderManager::rrTemporalMarkTile(Tile* curTile)
     ent->setVisible(bb);
 }
 
+void RenderManager::rrDetachEntity(GameEntity* curEntity)
+{
+    Ogre::SceneNode* node = mSceneManager->getSceneNode(curEntity->getOgreNamePrefix() + curEntity->getName() + "_node");
```
hwoarangmy: You have GameEntity::getEntityNode that would avoid to search the whole scene. AFAIK, it is set for all game entities.

source/render/RenderManager.cpp
```diff
@@ -665,6 +666,19 @@ void RenderManager::rrTemporalMarkTile(Tile* curTile)
     ent->setVisible(bb);
 }
 
+void RenderManager::rrDetachEntity(GameEntity* curEntity)
```
hwoarangmy: When you add new code, please avoid the curEntity and, more generally, cur...
Just use the simple name like GameEntity\* entity.

source/utils/Vector3i.h
```diff
@@ -0,0 +1,58 @@
...
+
+#ifndef VECTOR3I_H_
+#define VECTOR3I_H_
+
+#include <cstdint>
+#include <iostream>
+#include <OgreVector3.h>
+
+const int mPrecisionDigits = 20;
+const int Unit = (1 << mPrecisionDigits);
```
hwoarangmy: Thats a very bad name for a global constant and it would probably fit better in the Vector3i class

source/utils/Vector3i.h
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+#ifndef VECTOR3I_H_
+#define VECTOR3I_H_
+
+#include <cstdint>
+#include <iostream>
+#include <OgreVector3.h>
+
+const int mPrecisionDigits = 20;
```
hwoarangmy: in the code guidelines, constant variables are usually in capital letters with words separated by underscores. Moreover, it would fit better in the Vector3i class

source/utils/Vector3i.h
```diff
@@ -0,0 +1,58 @@
...
+const int Unit = (1 << mPrecisionDigits);
+
+
+//! \brief A custom vector used commonly between the CameraManager and the CullingManager classes
+class Vector3i
```
hwoarangmy: Note that here, the name is misleading. usually, i refers to int which is most of the case int32. I would not expect a int64 in this case.

hwoarangmy:
The feature seems to work pretty well. I've tried to navigate a bit and seen no big problem. The refresh in the minimap is acceptable to me.
However, it would be better to avoid increasing the minimap refresh frequency because it is cpu consuming and it makes the player hand blinking when over UI even worst.
Do you think it would be possible to only refresh tile culling when the camera moves ?

@tomluchowski Note that your PR needs to be rebased

source/camera/CullingManager.h
```diff
@@ -0,0 +1,104 @@
...
+    Ogre::Ray mMyRay[8];
+
+    bool mCullCreaturesFlag;
```
hwoarangmy: Again, until creature culling is implemented, please do not add code related to it. It will be much easier to read if all the related code is inside one commit.

source/camera/CullingManager.cpp
tomluchowski, line 172: Yes, in most frames we both look for Tiles to HIDE , as well as ones to SHOW :)

source/camera/CullingManager.cpp
tomluchowski: The old in variables name does not implies that the code is itself is old.

source/camera/CullingManager.cpp
tomluchowski: Could you explain the diffrence between OD_LOG_INF and OD_LOG_DBG ?

source/camera/CullingManager.cpp
tomluchowski: I don't understand that , proxy in the game entity class ?

source/camera/SlopeWalk.cpp
tomluchowski: So how do I do without stream class ? How do I quickly append and convert from certain types ?

source/gamemap/MiniMapCamera.cpp
tomluchowski, line 53: Yeap I am also a fan of a more and more effective culling algorithms .

source/utils/Vector3i.h
tomluchowski: Unit or mPrecisionDigits ?

source/utils/Vector3i.h
tomluchowski: so rename it to Vectorint64, right ?

source/camera/CullingManager.cpp
hwoarangmy, line 172: Yes and usually, it is one or the other. Using the values 1 and 2 for hide and show instead of a boolean (what would be expected in a case when a something is visible or not) makes think that a tile can be in both states. Is it the case here ?

source/camera/CullingManager.cpp
hwoarangmy: I'm not talking about old but about the commented code

source/camera/CullingManager.cpp
hwoarangmy: OD_LOG_INF is for LogMessageLevel::NORMAL
OD_LOG_DBG is for LogMessageLevel::TRIVIAL

source/camera/CullingManager.cpp
hwoarangmy: you used Tile::hide and Tile::show as proxies to call the render manager (ie it is only a call to RenderManager). Here, I'm telling you that you should call directly the render manager as culling only concerns the culling code (from the tile point of view, being shown or not doesn't matter)

source/camera/SlopeWalk.cpp
hwoarangmy: you can use Helper::toString

source/utils/Vector3i.h
hwoarangmy: Both. But Unit is worst because so common that you may have conflicts when including other libraries or other class members. That's why it is a bad practice to use so common variable names especially in a header.

Concerning mPrecisionDigits, the problem is that you followed class member variable naming. That means that when someone will read your code, he will think mPrecisionDigits is a class member variable while it is a constant.

source/utils/Vector3i.h
hwoarangmy: yes, it seems better. But note that you should use a capital letter for different words. Here, it would be VectorInt64.

source/gamemap/MiniMapCamera.cpp
hwoarangmy, line 53: Note that there is CameraManager::isCameraMovingAtAll()
My guess is that it seems to be pretty close to what you need to know if the minimap should be refreshed or not.

source/camera/CameraManager.h
tomluchowski, line 68: That was a choice it seems./

source/camera/CullingManager.cpp
tomluchowski: it looks nicer and more abstract, we don't look how the hide and show methods are implemented .

source/camera/CullingManager.cpp
tomluchowski, line 172: Nope you are wrong it's not either one or the other.
In cullTiles I do call 
`newBashAndSplashTiles(SHOW | HIDE);`

source/camera/CullingManager.h
tomluchowski: Could the variable refresh rate of minimap be the next pull requests? I have done quite much I think already ... .

source/camera/SlopeWalk.cpp
tomluchowski: It's locality is fine ; in that scope I am using  it notice 
`rightVertices.push_back(ii);`
for that example.

source/modes/ConsoleCommands.cpp
tomluchowski: I would lovely have both startTileCulling and stopTileCulling . However I cannot cope with the code ; could you correct the code below : 

```
    cl.addCommand("startTileCulling",
                  "start culling of the tiles",
                  cStartTileCulling,
                  {AbstractModeManager::ModeType::GAME, AbstractModeManager::ModeType::EDITOR});
```

ConsoleInterface::CommandClientFunction_t, ConsoleInterface::CommandServerFunction_t,
What the hell are those ?

tomluchowski:
Ehem rebased to what ? The upstream right ? 

EDIT : OKI, I am done , now please merge in. 
EDIT2 : 

Could the variable refresh rate of minimap be the next pull requests? I have done quite much I think already ... .

source/camera/CullingManager.cpp
hwoarangmy: You were talking about coupling some days ago. Here, you are coupling GameEntity with the culling logic. That means that if one day, you somehow change something about culling, you may have to change the tile. From an object point of view, it is not good.

Concerning polymorphism (since it is the main argument you are pointing out):
1 - ATM, all entities are shown/hiden the same way. Even you seem to think that it is not to be overriden since the method is not virtual
2 - The show/hide logic is simple and not likely to change in the futur or for new game entities
3 - If one really wants to do some shadock show/hide feature, it can easily be changed

It is a choice to go one way or the other. But IMHO, the gain with low coupling is better than with polymorphism

source/camera/SlopeWalk.cpp
hwoarangmy: Yes but the stop condition is ii != mDownLeftIndex . That means that pushing mDownLeftIndex is enough.

source/camera/CameraManager.h
hwoarangmy, line 68: ok, then it is a random change. Hopefully, GameMap is compatible with TileContainer. Seems like luck is on our side...

source/utils/VectorInt64.h
```diff
@@ -22,37 +22,37 @@
 #include <iostream>
 #include <OgreVector3.h>
 
-const int mPrecisionDigits = 20;
-const int Unit = (1 << mPrecisionDigits);
+const int PRECISION_DIGITS = 20;
+const int UNIT = (1 << PRECISION_DIGITS);
```
hwoarangmy: I've not looked close enough to know what UNIT is used for but when I said "Unit" was not a good name, "UNIT" is not much better ;-)

Something like VECTOR3INT64_UNIT seems better. The idea is to not have such common variable names to avoid possible conflicts (and usually, such conflicts are pretty hard to debug if you includ a header that only defines UNIT if it is not).

source/utils/VectorInt64.h
```diff
@@ -22,37 +22,37 @@
...
 
 
 //! \brief A custom vector used commonly between the CameraManager and the CullingManager classes
-class Vector3i
+class VectorInt64
 {
 public:
-    Vector3i():
+    VectorInt64():
```
hwoarangmy, line 35: Sorry to have a comment again here but Vector3Int64 would be better :-(

source/camera/CullingVectorManager.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+void CullingVectorManager::sort()
+{
+    std::vector<VectorInt64> ll[4];
+    VectorInt64 cc(0, 0, 0);
+
+    // find the center of points stored in mMyArray
+    for (auto ii : mMyArray)
+        cc = cc + ii;
+    cc = cc / mMyArray.size();
```
hwoarangmy, line 31: after a for boucle without braces, it is better to add an empty line to help reading

hwoarangmy:
There are still a few things to fix but overall, it doesn't look bad. IMHO, it can be merged once the last requested stuff are ok. Then please just merge everything in one commit.

> Could the variable refresh rate of minimap be the next pull requests?

ok. Do you want some help on this ?

tomluchowski:
OK, I am done I believe, could someone do the merge , @hwoarangmy : help would be not needed , but let's leave that for another commit.

source/camera/CameraManager.cpp
```diff
@@ -47,14 +51,16 @@ const Ogre::Degree ROTATION_SPEED = Ogre::Degree(90);
...
+CameraManager::CameraManager(Ogre::SceneManager* sceneManager, GameMap* gm, Ogre::RenderWindow* renderWindow) :
+    mCullingManager(nullptr),
     mCircleMode(false),
     mCatmullSplineMode(false),
     mFirstIter(true),
     mRadius(0.0),
     mCenterX(0),
     mCenterY(0),
     mAlpha(0.0),
+
```
hwoarangmy: useless empty line

source/camera/CameraManager.cpp
```diff
@@ -72,6 +78,7 @@ CameraManager::CameraManager(Ogre::SceneManager* sceneManager, TileContainer* gm
     mSceneManager(sceneManager),
     mViewport(nullptr)
 {
+    mCullingManager = new CullingManager(this);
```
hwoarangmy, line 80: If I'm not mistaken, there is a memory leak

source/camera/CameraManager.cpp
```diff
@@ -576,6 +583,30 @@ void CameraManager::move(const Direction direction, double aux)
     }
 }
 
+void CameraManager::startTileCulling()
+{
+
```
hwoarangmy: useless empty line

source/camera/CameraManager.cpp
```diff
@@ -576,6 +583,30 @@ void CameraManager::move(const Direction direction, double aux)
...
+    mCullingManager->startTileCulling();
+}
+
+void CameraManager::stopTileCulling()
+{
+
```
hwoarangmy: useless empty line

source/camera/CullingManager.cpp
```diff
@@ -0,0 +1,253 @@
...
+
+#include "entities/Creature.h"
+#include "utils/VectorInt64.h"
+#include "utils/LogManager.h"
+#include "render/RenderManager.h"
+#include <sstream>
+#include <algorithm>
+
+//! Values used to know whether to show and/or hide a mesh
+static const int HIDE =  1;
```
hwoarangmy: This should be static in the CullingManager class to avoid problems with same constant name (pretty likely since names are very common)

source/camera/CullingManager.cpp
```diff
@@ -0,0 +1,253 @@
...
+static const int SHOW =  2;
```
hwoarangmy: Same

source/camera/CullingManager.cpp
```diff
@@ -0,0 +1,253 @@
...
+    OD_LOG_DBG( mOldWalk.debug());
+    OD_LOG_DBG( mWalk.debug());        
+
+    // reset index pointers to the begging of collections
+    mOldWalk.prepareWalk();
+    mWalk.prepareWalk();
+
+    newBashAndSplashTiles(HIDE);
+    showAllTiles();
+
```
hwoarangmy: useless empty line

source/camera/CullingManager.cpp
```diff
@@ -0,0 +1,253 @@
...
+}
+
+void CullingManager::hideAllTiles(void)
+{
+    GameMap* gm = mCm->getGameMap();
+    if (!gm)
```
hwoarangmy, line 116: IMHO, there is no need to check for game map being null as it is already expected not to be null everywhere else. But no need to change if you prefer letting like that

source/camera/CullingManager.cpp
```diff
@@ -0,0 +1,253 @@
...
+
+                xxp = ((xx >>PRECISION_DIGITS));
+                yyp = ((yy >>PRECISION_DIGITS));
+                GameMap* gm = mCm->getGameMap();
+                if(bash && splash && (mode & HIDE) && (mode & SHOW))
+                {
+                    // Nothing
+                }
+
+                else if (gm && bash && (mode & HIDE) && xxp>=0 && yyp>= 0 && xxp<mCm->getGameMap()->getMapSizeX() && yyp < mCm->getGameMap()->getMapSizeY())
```
hwoarangmy, line 177: As I said, using gm->getTile(xxp, yyp) and checking if returned tile is null would be more readable and would avoid a if with 7 checks in one line

source/camera/CullingManager.cpp
```diff
@@ -0,0 +1,253 @@
...
+                {
+                    RenderManager::getSingleton().rrDetachEntity(gm->getTile(xxp, yyp));
+                }
+                else if (gm && splash && (mode & SHOW) && xxp>=0 && yyp>= 0 && xxp<mCm->getGameMap()->getMapSizeX() && yyp < mCm->getGameMap()->getMapSizeY())
```
hwoarangmy, line 181: same

source/camera/CullingManager.cpp
```diff
@@ -0,0 +1,253 @@
...
+                }
+            }
+        }
+    }
+}
+
+bool CullingManager::getIntersectionPoints()
+{
+    const Ogre::Vector3* activeCameraVector = mCm->getActiveCamera()->getWorldSpaceCorners();
+    const Ogre::Vector3* activeCameraVector2 = mCm->getCamera("miniMapCam")->getWorldSpaceCorners();
```
hwoarangmy, line 193: Here, you are coupling culling with the minimap camera. I guess this would segfault if we use the old minimap. Using the minimap to have the coordinates of the displayed tiles would be a better way to go.

This can be done later however but would require an issue

source/camera/CullingManager.cpp
```diff
@@ -0,0 +1,253 @@
...
+    for(int ii = 0 ; ii < 4; ++ii)
+        mActiveRay[ii]= Ogre::Ray (activeCameraVector[ii], activeCameraVector[ii+4] - activeCameraVector[ii]);
+
+    for(int ii = 4; ii < 8; ++ii)
+        mActiveRay[ii]= Ogre::Ray (activeCameraVector2[ii-4], activeCameraVector2[ii] - activeCameraVector2[ii-4]);
+
```
hwoarangmy, line 199: too may empty lines

source/camera/CullingVectorManager.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+void CullingVectorManager::zoom(double zz)
+{
+    VectorInt64 cc(0, 0, 0);
+
+    for (auto& ii : mMyArray)
+        cc = cc + ii;
+    
+    cc = cc / mMyArray.size();
+
```
hwoarangmy: empty line

source/camera/CullingVectorManager.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+    for (auto& ii : mMyArray)    
+    {
+        ii = cc + (ii - cc)*zz;    
+    }
+
```
hwoarangmy: empty line

source/camera/SlopeWalk.cpp
```diff
@@ -0,0 +1,313 @@
...
+BOOST_GEOMETRY_REGISTER_POINT_2D(VectorInt64, int64_t, boost::geometry::cs::cartesian, x, y)
+//! Values used to know whether to show and/or hide a mesh
+typedef VectorInt64 Point;
+typedef boost::geometry::model::polygon<Point> Polygon;
+
+template <typename Point>
+void get_coordinates(Point const& p)
+{
+    OD_LOG_DBG( "get__coordinates : " + Helper::toString(double(boost::geometry::get<0>(p))/UNIT) + ", " + Helper::toString( double(boost::geometry::get<1>(p))/UNIT )) ;
+
```
hwoarangmy: empty line

source/camera/SlopeWalk.cpp
```diff
@@ -0,0 +1,313 @@
...
+}
+
+// reset indexes to the begginging of containers 
+void SlopeWalk::prepareWalk()
+{
+    mLeftSlopeIndex =  mLeftSlopes.begin();
+    mRightSlopeIndex = mRightSlopes.begin();
+    leftVerticesIndex = leftVertices.begin();
+    rightVerticesIndex = rightVertices.begin();
+
```
hwoarangmy: empty line

source/camera/SlopeWalk.cpp
```diff
@@ -0,0 +1,313 @@
...
+    rightVerticesIndex++;
+    mRightSlopeIndex++;
+    return rightVerticesIndex == rightVertices.end();
+
+}
+
+// Check whether we pass a new Vertex on the left or right path, if so notify about it 
+bool SlopeWalk::notifyOnMoveDown(int64_t newyIndex)
+{
+
```
hwoarangmy: empty line

source/camera/SlopeWalk.cpp
```diff
@@ -0,0 +1,313 @@
...
+    return leftVerticesIndex == leftVertices.end();
+}
+
+// What to do when passing one Vertex down on the Right Path 
+bool SlopeWalk::passRightVertex()
+{
+    rightVerticesIndex++;
+    mRightSlopeIndex++;
+    return rightVerticesIndex == rightVertices.end();
+
```
hwoarangmy: empty line

source/camera/SlopeWalk.cpp
```diff
@@ -0,0 +1,313 @@
...
+}
+// get the value of slope currently pointed by index on the right path
+int64_t SlopeWalk::getCurrentXRight(int64_t yy){
+    if(mRightSlopeIndex != mRightSlopes.begin())
+    {
+        return getPreviousRightVertex().x + ((*mRightSlopeIndex) * (yy - getPreviousRightVertex().y))/UNIT ;
+    }
+    else
+        return getCurrentRightVertex().x;
+
```
hwoarangmy: empty line

source/camera/SlopeWalk.h
```diff
@@ -0,0 +1,104 @@
...
+    std::deque<int64_t>::iterator mRightSlopeIndex;
+    std::deque<int64_t>::iterator mLeftSlopeIndex;
+
+
+    // index numbers in myArray of Vertices 
+    // belonging to the right path
+    std::deque<int> rightVertices;
+    // index numbers in myArray of Vertices
+    // belonging to the left path
+    std::deque<int> leftVertices;
```
hwoarangmy: please follow code guidelines

source/camera/SlopeWalk.h
```diff
@@ -0,0 +1,104 @@
...
+    // point the Vertex currently processed 
+    // on the right path
+    std::deque<int>::iterator rightVerticesIndex;
```
hwoarangmy: please follow code guidelines

source/camera/SlopeWalk.h
```diff
@@ -0,0 +1,104 @@
...
+    // pint the Vertex currenly processed
+    // on the left path
+    std::deque<int>::iterator leftVerticesIndex;
```
hwoarangmy: please follow code guidelines

source/modes/ConsoleCommands.cpp
```diff
@@ -473,6 +473,12 @@ Command::Result cTriggerCompositor(const Command::ArgumentList_t& args, ConsoleI
     return Command::Result::SUCCESS;
 }
 
+Command::Result cStartTileCulling(const Command::ArgumentList_t& args, ConsoleInterface& c, AbstractModeManager&)
```
hwoarangmy: This function is not used. It should be removed or used.

source/modes/GameEditorModeBase.cpp
```diff
@@ -132,6 +133,12 @@ GameEditorModeBase::~GameEditorModeBase()
     // Delete the potential pending event messages
     for (EventMessage* message : mEventMessages)
         delete message;
+
+    // Clear up any events and chat messages.
+    mRootWindow->getChild("GameChatWindow/GameChatText")->setText("");
```
hwoarangmy: Please pay attention to not upload code that have been changed. You always complain that you are too much reviewed but I can't remember any PR without loads of old code restored. I don't know how you setup your dev environment but there is clearly a problem.

source/network/ODClient.cpp
```diff
@@ -47,6 +47,7 @@
 #include "utils/LogManager.h"
 #include "ODApplication.h"
 
+
```
hwoarangmy: empty line

source/utils/VectorInt64.cpp
```diff
@@ -0,0 +1,67 @@
+/*
+ *  Copyright (C) 2011-2014  OpenDungeons Team
```
hwoarangmy: BTW, we are in 2016 ATM

source/utils/VectorInt64.h
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+#ifndef VECTOR3I_H_
+#define VECTOR3I_H_
+
+#include <cstdint>
+#include <iostream>
+#include <OgreVector3.h>
+
+const int PRECISION_DIGITS = 20;
```
hwoarangmy: This should be static in the VectorInt64 class for same reasons as above

source/utils/VectorInt64.h
```diff
@@ -0,0 +1,58 @@
...
+const int UNIT = (1 << PRECISION_DIGITS);
```
hwoarangmy: same

hwoarangmy:
I've not signaled all the random empty lines because there were too many. Please at least remove when there are too many or when they hurt readability (like in a constructor init list).

source/modes/GameEditorModeBase.cpp
tomluchowski: When there is conflict in git I choose the piece which has the code in it, that's my secret to be honest , and it seems that it does not do it's job very well ./

hwoarangmy: No, the secret is to check what have been done in the conflicting PR and do the same. Here, some code had been moved and you re-added it.

source/camera/CullingManager.cpp
tomluchowski, line 177: So you proposed another line nested if inside else-if clause ? If I correctly understand ?

hwoarangmy, line 177: In this case, I would propose something like:

```
        GameMap* gameMap = mCm->getGameMap();
        if(std::min(xxLeft, xxLeftOld) < std::max(xxRight,xxRightOld) )
        {
            for (int64_t xx = mm ; xx <= std::max(xxRight,xxRightOld) ; xx+= UNIT)
            {
                bool bash = (xx >= xxLeftOld && xx <= xxRightOld && (yy >= mOldWalk.getBottomLeftVertex().y) && yy <= mOldWalk.getTopLeftVertex().y);
                bool splash = (xx >= xxLeft && xx <= xxRight && (yy >= mWalk.getBottomLeftVertex().y) && yy <= mWalk.getTopLeftVertex().y);

                xxp = ((xx >>PRECISION_DIGITS));
                yyp = ((yy >>PRECISION_DIGITS));
                if(bash && splash && (mode & HIDE) && (mode & SHOW))
                    continue;

                Tile* tile = gameMap->getTile(xxp, yyp);
                if(tile == nullptr)
                    continue;

                if (bash && (mode & HIDE))
                {
                    RenderManager::getSingleton().rrDetachEntity(tile);
                    continue;
                }

                if (splash && (mode & SHOW))
                {
                    RenderManager::getSingleton().rrAttachEntity(tile);
                    continue;
                }
            }
        }

```

hwoarangmy:
I've had a quick look and seen that 
SlopeWalk::mLeftVertexPassed
SlopeWalk::mRightVertexPassed

are set but not used. Is it some forgotten test variables ?

tomluchowski:
@hwoarangmy : They are used in bool SlopeWalk::notifyOnMoveDown(int64_t newyIndex)
EDIT : but not too much . :P

source/camera/CullingManager.cpp
tomluchowski, line 177: At first I laughed, but after further investigation I was able to read that code  . 
If I could have something to say about my code that would be the only site not to change it because I hate using continue and break .

hwoarangmy, line 177: Well, I have nothing against it but I would prefer to have real arguments instead of "using break is the first step to hell".
But since you don't feel like giving ones, I will give mines for using break/continue.
1 - It allows to clearly separate different parts of code that should stop or go to next iteration. In my code, it is pretty obvious that "if(bash && splash && (mode & HIDE) && (mode & SHOW))", we don't want to proceed the tile. In yours, you have to check that all other conditions are "else if" and not if of anything else.
2 - It allows to isolate conditions that should stop processing the loop. For example, if we assume you want to proceed tiles with X even, Y even, fullness > 0 and type = dirt, your code would be something like:

```
for(Tile* tile : tiles)
{
   if((tile->getX() % 2) || (tile->getY() % 2) || (tile->getFullness() <= 0) || (tile->getType() != TileType::Dirt))
   {
      // Nothing
   }
   else
   {
      // Processing code
   }
}
```

If, for some reason, you or someone else have to re-write this code, he is likely to forget a condition and at review, it is easy to miss the forgotten part. And I am talking about a real case since this exact problem happened to me when improving the room absorbtion code. It have been merged without seeing I had forgotten one of the conditions. Thus, we had a pretty nasty segfault in some specific cases.

My code would be more like:

```
for(Tile* tile : tiles)
{
   if(tile->getX() % 2)
      continue;
   if(tile->getY() % 2)
      continue;
   if(tile->getFullness() <= 0)
      continue;
   if(tile->getType() != TileType::Dirt)
      continue;

   // Processing code
```

I guess you will agree that it is not as likely to miss a conditions like that.

3 - It allows to have less imbricated code. See my previous example. And, for another example, as you said, in your code, if you are to test if returned tile from the gamemap is null instead of checking the conditions that will be checked again in getTile, it will add another imbrication. Not in my code.

4 - Because the conditions are more isolated, it allows to see what has been changed at the first glance at code review. For example, if you add or remove one of the conditions in the previous example, in your code, that will change the one line with the list of conditions. In my version, you will add/remove a specific part of the code that will be easy to identify.

5 - For the same reason as 4, it helps keeping code history more readable.

To be honest, before I joined OD team I had the same feeling as yours about continue/break and I never used it (I had never thought about it but since my teachers always told me that using them was bad, I didn't use them). But after seeing it used in OD code and thinking about it, it became obvious that it is much better as the only reason I see to not use them is the same you gave so far: "it is bad to use them". But when I think about it, all the reasons I can find tell me the opposite. And keeping code history clean and better readability is something that should be above everything else, especially in an open source project like this one.

Please feel free to comment what you think of my arguments and do not hesitate to tell if you see something else. It is a free time project and we are all here to have fun and increase our skills. Discussing about stuff like that is a good way for that.

source/camera/SlopeWalk.h
```diff
@@ -0,0 +1,96 @@
...
+private:
+
+    std::deque<int64_t> mRightSlopes;
+    std::deque<int64_t> mLeftSlopes;
+    std::deque<int64_t>::iterator mRightSlopeIndex;
+    std::deque<int64_t>::iterator mLeftSlopeIndex;
+
+    // index numbers in myArray of Vertices 
+    // belonging to the right path
+    std::deque<int32_t> rightVertices;
```
hwoarangmy: you really don't want to follow code guidelines...

source/camera/CullingManager.cpp
```diff
@@ -0,0 +1,246 @@
...
+    for (int jj = 0; jj < gm->getMapSizeY() ; ++jj)
+    {
+        for (int ii = 0; ii < gm->getMapSizeX(); ++ii)
+        {
+            RenderManager::getSingleton().rrAttachEntity(gm->getTile(ii, jj));
+        }
+    }
+}
+
+void CullingManager::newBashAndSplashTiles(int64_t mode){
```
hwoarangmy: Note that here, you are using mode as an array of bytes. using a int64_t is not usefull for an array that can have only 2 values. Moreover, for this kind of usage, we usually use an unsigned int such as uint32_t.
Since there are still changes to do, it would be nice to change that at the same time.

source/camera/CullingManager.h
```diff
@@ -0,0 +1,97 @@
...
+ * 4. Between each following pair of vertices we can establish a >> slope << ,
+ * which is just the the "a" in the eq. of linear form y = ax + b
+ * 5. Now start drawing our polygon Row by Row From top to bottom . 
+ * Each Row has given the most left and rightmost tile due to use of both paths prepared before 
+ * -- Left path for tracing the most Leftmost Tile , Right path teh most Rightmost Tile in each .
+ */
+class CullingManager
+{
+public:
+    static const int HIDE =  1;
```
hwoarangmy: that would be better to use the same variable type as used in newBashAndSplashTiles. In this case, the most fitted would be uint32_t. That will avoid implicit type conversion.

source/camera/CullingManager.h
```diff
@@ -0,0 +1,97 @@
...
+    static const int SHOW =  2;
```
hwoarangmy: same

hwoarangmy:
@tomluchowski It can be done later but do you think it would be hard to split the CullingManager so that we can use one instance for the main screen and another for the minimap ?
That would decrease coupling between the CullingManager and the rest of the classes.

tomluchowski:
That would be increadibly hard if not impossible. The whole algorithm needs to compute the convex_hull of ALL points exactly ONCE.

source/camera/SlopeWalk.h
tomluchowski: What have I missed in code guidelines that I fail to implement in here ?

source/camera/CullingManager.cpp
tomluchowski: uint32_t. is g++ specific afaik.

source/camera/SlopeWalk.h
hwoarangmy: class member variables names must start with m. That is, you should rename this variable to mRightVertices.

source/camera/CullingManager.cpp
tomluchowski, line 177: Maybe we could get to a compromise writting the condition by each clause in separate line : 

```
if((tile->getX() % 2) 
|| (tile->getY() % 2) 
|| (tile->getFullness() <= 0) 
|| (tile->getType() != TileType::Dirt))
```

EDIT : For me and most beginners the mine code style is just easier to read and understand . Probably I am not as skilled and wordy as you are ; but if you insist on changing that I would have to change that because it won't be merged into upstream. Wonder which code @Bertram and @akien-mga found better to read ...

source/camera/CullingManager.cpp
hwoarangmy: well, it works with gcc, MinGW and MSVS2013 (it is used everywhere in OD code) so all the supported platforms. That's why it should be prefered.

tomluchowski: ohh I meant uint , never mind then .

tomluchowski:
Could you also answer my post on the forum ?

source/camera/CullingManager.cpp
hwoarangmy, line 177: As I said, this is not blocking merging. I was just detailing stuff we had talked about in another PR. But believe me, if you start reviewing code, you will start to pay attention to that.
The first PRs I've done were more than 50 files changed with not well splitted stuff (also due to the fact that the code was less mature as it currently is). And thinking about it, it was just unreadable ^^

source/camera/SlopeWalk.cpp
```diff
@@ -0,0 +1,294 @@
...
+    boost::geometry::correct(polygon);
+
+    assert(boost::geometry::num_points(polygon) == 9);
+
+    Polygon hull;
+    mVertices.mMyArray.clear();
+
+    boost::geometry::convex_hull(polygon, hull);
+    boost::geometry::for_each_point(boost::geometry::exterior_ring(hull), [this](Point &p){this->mVertices.mMyArray.push_back(p);});
```
hwoarangmy, line 292: For the next time, please split this into several lines to make clear there is a function here. Not blocking

hwoarangmy:
I would have prefered to merge that after the release but at our current pace, that may be long. I have a PR almost done for allowing to change the minimap (switching from the old and the camera one) and if it gets merged before yours, you will have to make sure it works. I guess you would curse me for many generations if that was to happen ^^

Sorry for all the requested changes but the code looks much better now and is mergeable as for me

> Could you also answer my post on the forum ?

Which one ?

PR #1213, opened 2016-06-08
Added a timeout to chat messages

hwoarangmy:
The chat message box will be displayed:
- for 30 seconds after receiving a message (harcoded, easy to move where we want later)
- When the edit box to write a chat message is displayed (when enter is pressed)

The chat messages are only removed at game start. That means that history can be browse during the whole game


akien-mga:
Works fine and code looks good :)

PR #1217, opened 2016-06-15
Allow to use the old minimap or the new one in the user settings

hwoarangmy:
The minimap can be changed in the settings. The new minimap will be chosen at the start of the next new game (no need to reload the executable)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -371,6 +371,7 @@ set(OD_SOURCEFILES
 
     ${SRC}/gamemap/GameMap.cpp
     ${SRC}/gamemap/MapHandler.cpp
+    ${SRC}/gamemap/MiniMap.cpp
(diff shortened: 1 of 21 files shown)


tomluchowski:
hwoarangmy : I can write much simpler version of culling switcher between Main window or both. All you must do is to change the number of corners supplied to the algorithm . ( 4 vs 8 )

hwoarangmy:
Well, that's what I've done. The rest is just adapting code to guidelines. And I also moved the culling manager to the gamemode to avoid a failing assert because culling was still ongoing when the gamemap was cleared (at game quit).
The only thing that may look a bit more complex is the use of a bit array to cull tiles. But since there might be more than 1 culling manager, it's the simplest I can think of.

akien-mga:
Code looks good :)

akien-mga:
I'll run codespell once to see if I catch some typos.

PR #1220, opened 2016-06-18
Example culling switch between only one window or both

tomluchowski:
Here's example of culling switch between only for one window vs both . What the hw. has done  is totally obscure to me -- especially adding a new int value to each Tile to control it's visibility.

### DIFF
--- a/source/camera/CullingManager.cpp
+++ b/source/camera/CullingManager.cpp
@@ -32,6 +32,7 @@
 //! Values used to know whether to show and/or hide a mesh
 
 CullingManager::CullingManager(CameraManager* cameraManager):
+    mWindowCulling(CullingType::SHOW_ONLY_MAIN_WINDOW),
     mFirstIter(false),
     mCm(cameraManager),
     mCullTilesFlag(false)
@@ -50,7 +51,7 @@ void CullingManager::cullTiles()
 {
     mOldWalk = mWalk;
     mWalk.mVertices.mMyArray.clear();
-    for (int ii = 0 ; ii < 8 ; ++ii)
+    for (int ii = 0 ; ii < int(mWindowCulling) ; ++ii)
         mWalk.mVertices.mMyArray.push_back(VectorInt64(mOgreVectorsArray[ii]));
 
     // create a slope -- a set of left and rigth path
@@ -73,7 +74,7 @@ void CullingManager::startTileCulling()
     getIntersectionPoints();
 
     mWalk.mVertices.mMyArray.clear();
-    for (int ii = 0 ; ii < 8 ; ++ii)
+    for (int ii = 0 ; ii < int(mWindowCulling) ; ++ii)
         mWalk.mVertices.mMyArray.push_back(VectorInt64(mOgreVectorsArray[ii]));
     mWalk.convex_hull();
     mWalk.buildSlopes();
@@ -92,7 +93,7 @@ void CullingManager::stopTileCulling()
     mCullTilesFlag = false;
     mOldWalk = mWalk;
     mWalk.mVertices.mMyArray.clear();
-    for (int ii = 0 ; ii < 8 ; ++ii)
+    for (int ii = 0 ; ii < int(mWindowCulling) ; ++ii)
         mWalk.mVertices.mMyArray.push_back(VectorInt64(mOgreVectorsArray[ii]));
 
     // create a slope -- a set of left and rigth path
--- a/source/camera/CullingManager.h
+++ b/source/camera/CullingManager.h
@@ -50,6 +50,12 @@ class GameMap;
 class CullingManager
 {
 public:
+
+
+    enum class CullingType{
+        SHOW_ONLY_MAIN_WINDOW = 0x04,
+        SHOW_BOTH = 0x08
+    } mWindowCulling;
     static const uint32_t HIDE =  1;
     static const uint32_t SHOW =  2;
 
--- a/source/camera/SlopeWalk.cpp
+++ b/source/camera/SlopeWalk.cpp
@@ -279,11 +279,11 @@ void SlopeWalk::convex_hull(){
     Polygon polygon;
     boost::geometry::append(polygon, mVertices.mMyArray);
 
-    assert(boost::geometry::num_points(polygon) == 8);
+    // assert(boost::geometry::num_points(polygon) == 8);
 
     boost::geometry::correct(polygon);
 
-    assert(boost::geometry::num_points(polygon) == 9);
+    // assert(boost::geometry::num_points(polygon) == 9);
 
     Polygon hull;
     mVertices.mMyArray.clear();


hwoarangmy:
What I've done is to add a bit array to the tile so that it can be hiden/shown depending on if it is culled or not. Basically, if the bit array is 0, nobody wants to show the tile and it is hiden. If it is 1, 2 or 3, it means that the main window, the minimap or both want to show the tile and it is shown.
That's all.

The difference your code and mine is that my code is more generic and allows to have new features that will show the tiles if they need.
Your code has high coupling with the other classes (for example, it will segfault if the camera name from the minimap is changed).
Since the minimap is generic, there can be new types that won't work exactly the same and the higher the coupling is, the harder it may become to add the feature.

tomluchowski:
What features do you plan to add that this is MUST HAVE ? 
Adding additional flag into Tiles destroys all the charm and work I put in culling algorithm. 
There are two options : either we want a minimap which uses camera ( call that culling minimap ) or the minimap does not use the camera and therefore does not do any culling ( let's call it non-culling minmap ) . And I don't see any third option, despite how many generic new minimaps you want to introduce .... I don't bother the other code upgrades, but culling flag inside Tile is nasty ...

hwoarangmy:
I must say that I would also have prefered to keep the culling code in its own part without messing with the Tile code. I am currently testing my code about culling entities on tiles. I will have a look at the impact to not use the bit array by using something similar to your code.
But in any case, for creature culling, we need to keep in game entity the node state (if it is attached to the parent node or not). That's because there are other actions that can attach/detach nodes (picking up a creature or when a worker carries a dead creature). So, I don't see how to avoid some flag in the game entity anyway (even if not related to culling only).

tomluchowski:
Yeap, I did try to write the creature culling and tracing all possiblities why creature might be hidden - that gave me a headache. Anyway creature culling ain't tile culling -- it won't speed up fps much as my test showed.

hwoarangmy:
I've reverted the changes to use only one culling manager but the issue with minimap refresh is back and requires to increase refresh frequency just like you did (which causes the keeper hand to blink even more than it currently does).
IMHO, it is better to live with a variable that we would have prefered to not exist rather than hurting player experience.
Note that from a performance point of view, in the current code, we use 2 culling manager while in the original one (yours) there was only one. On the other hand, the refresh is not the same (on the main map, it is computed each frame while it it done only twice per seconds for the minimap). I've not done any checks but since it makes the keeper hand blink, I would say it may hurt performances.
@OpenDungeons/developers WDYT ?

> Anyway creature culling ain't tile culling -- it won't speed up fps much as my test showed.

Well, I've not done tests but since 1 - we have tile culling and creatures (and the other entities as well) are associated with the tiles they are walking on, it would be a waste to stop here (even if there might be issues and regression with such feature - not the best to add while talking about releasing). Moreover, creatures have much more faces than tiles so it might not be neutral to cull them.

> especially adding a new int value to each Tile to control it's visibility

Apart from the current discussion, note that this is an interesting pattern. The basic utility is to use a bit array to enable or disable features that can be enabled/disabled by different means. For example, if you create the program for a car radio, you will have some driver utilities allowing to play, stop a sound an change volume level. Now, if we assume we want to add a mute key, the naive approach will be to call some mute/unmute driver. But now, if you add other features like bluetooth phone or navigation, you will want to mute the sound when it starts and unmute when it stops. The naive approach will also be to call the same mute functions. However, that will mean that if you manually mute the sound, if you get a call, the sound will unmute when you're done. That's not what you will expect as a user (in my car, that's the behaviour).
Another approach will be to identify the features that may want to mute (key, phone, navigation), associate each with a bit (1, 2, 4) and use a bit array. Then, when a feature is activated, it turns the corresponding bit on and we know that sound should be muted. When all are turned off, the sound is unmuted. And if many concurrently happen, there is no problem.

The approach with the tiles is the same. At start, we set the value to 0 (the tile is hidden) and if the main map or the minimap needs to show a tile, it turns the corresponding bit on.

tomluchowski:
What's the connection between refresh rate of minimap and drawing a keeper's hand ?

hwoarangmy:
For some reason, when the minimap is rendered, it makes the keeper hand blink when is is over the UI (over CEGUI widgets). The more often the minimap is refreshed, the more it blinks.

tomluchowski:
> Note that from a performance point of view, in the current code, we use 2 culling manager while in the original one (yours) there was only one. On the other hand, the refresh is not the same (on the main map, it is computed each frame while it it done only twice per seconds for the minimap). I've not done any checks but since it makes the keeper hand blink, I would say it may hurt performances.

Do I understand correctly ? You suggest the approach of two culling managers might be faster , because the second one is called only twice a second , while the first one with every frame ? 

EDIT : also if you plan to stay your way with the tiles, mind to remove the convex_hull finding part of the algorithm , as it is not needed for four verticies ....

hwoarangmy:
> Do I understand correctly ? You suggest the approach of two culling managers might be faster , because the second one is called only twice a second , while the first one with every frame ?

I was talking about the fact that refreshing the minimap 10 times more than it used to be is probably CPU consuming since it creates a visible glitch (the keeper hand blinking more). Concerning the culling itself, using 2 managers is probably worst (especially since they look more or less at the same tiles) but refreshing one of them only twice per second may make a difference. In any case, I don't thnik the performance of culling computation in itself is very relevant regarding the minimap drawing.
Note that I had run gprof to check the performance of the game before the camera minimap and the most cpu consuming task was the minimap drawing.

> EDIT : also if you plan to stay your way with the tiles, mind to remove the convex_hull finding part of the algorithm , as it is not needed for four verticies ....

It is not a religion, I'm just trying to figure out what is the best from the player point of view (as the result is the most important IMHO). And from this point of view, using only one culling manager makes the minimap have to be refreshed much more which makes the keeper hand blink more. It also creates a glitch while scrolling on the minimap as you had already noticed yourself. Both problems disapear when using 2 culling managers so, yes, I think it is better to stay like that. If you have an idea how to improve that, I'm open. But I've tried to revert my changes to have only one culling manager and couldn't find a way to avoid the minimap glitch when moving.

Concerning the convex_hull finding part of the algorithm, please feel free to create a PR if you see improvements as you know this part better than I do.

tomluchowski:
Yeap, the drawn minimap is actually slower than the rendered 3d one !!!
For the keeper hand :  for me and most players it would be one mess , whether it blinks twice a second or twenty a second , this is something we should fix.

hwoarangmy:
> For the keeper hand : for me and most players it would be one mess , whether it blinks twice a second or twenty a second , this is something we should fix.

Definitely. But I don't think it will be an easy task as I've had a quick look some time ago but couldn't find anything similar. There is already an issue about it but I don't think it will be fixed anytime soon.

tomluchowski:
Well, I am bit lost inside that code , where is this second culling manager for minimap we are talking about ?

hwoarangmy:
It's there:
https://github.com/OpenDungeons/OpenDungeons/blob/development/source/gamemap/MiniMapCamera.cpp#L76

BTW, if you want, you can have a look at my last PR as it is about culling too. Note that you said you want to practive C++ and reviewing PRs is a pretty good way IMHO.

PR #1222, opened 2016-06-21
Use culling on entities on tiles

hwoarangmy:


tomluchowski:
Why do you mix Creature culling and Tile Culling into one function : setParentNodeDetach ? Could tile be pickedUp, or carried by another creature ? Today I ended up parsing the previous commit and this makes culling even more complicated .... Could you stay with the code for Tiles in newBashAndSplash and just set  / unset proper flags for it's child entities ( in this case creatures ) ? This would make the code less complicated ... Besides you still use the fooling if - return pattern ( fooling the reader :) ) .

hwoarangmy:
> Could you stay with the code for Tiles in newBashAndSplash and just set / unset proper flags for it's child entities

In any case, the tile has to be aware if it is culled or not. That's because even if the camera doesn't move (thus, culling will not do anything), the creatures (or more generally, the entities) might move and go from a culled tile to a visible tile (or the opposite).

> Why do you mix Creature culling and Tile Culling into one function : setParentNodeDetach ? Could tile be pickedUp, or carried by another creature ?

No but that's the only game entity in this case (all others can). The only way I can see to achieve what you ask is to use polymorphism but I'm not sure it will be more readable as it would mean to spread the culling (more generally the node attachement) in several places and we might forget some (for example maplight can be picked up but not carried). It would also mean to use different variables for node attachement for different cases which may lead to more complex code and more variables for entities using more than tile culling (ie all entities except tiles).
Note that there are already many stuff on game entities that do not work for all their childs (for example, they have functions for picking, dropping, taking damage...). Having a takeDamage function in the tile does sound more weird to me than having not applicable bit arrays (which are not used within the tile btw).
If you have a better idea, I'm open to discussion so don't hesitate to give some quick details (no need to open a PR but I don't see something better than what I did - which is the reason why I did it this way ;) ).

source/entities/Tile.cpp
```diff
@@ -1860,6 +1868,29 @@ bool Tile::removeWorkerDigging(const Creature& worker)
...
+    mTileCulling = (value ? mTileCulling | mask : mTileCulling & ~mask);
+
+    if(mTileCulling == CullingType::HIDE)
+    {
+        // We cull the tile
+        setParentNodeDetach(EntityParentNodeAttach::DETACH_CULLING, true);
+        for(GameEntity* entity : mEntitiesInTile)
+            entity->setParentNodeDetach(EntityParentNodeAttach::DETACH_CULLING, true);
+
+        return;
```
tomluchowski: Can you leave the return and just use good, old if else branching statement ?

source/entities/GameEntity.cpp
```diff
@@ -616,4 +615,20 @@ void GameEntity::notifyFightPlayer(Tile* tile)
...
+{
+    // We save the current attach  state
+    bool oldState = (mEntityParentNodeAttach == EntityParentNodeAttach::ATTACHED);
+    // We compute the new attach state
+    mEntityParentNodeAttach = (value ? mEntityParentNodeAttach | mask : mEntityParentNodeAttach & ~mask);
+
+    // If the attach state changed, we do what is needed
+    bool newState = (mEntityParentNodeAttach == EntityParentNodeAttach::ATTACHED);
+    if(oldState == newState)
+        return;
```
tomluchowski, line 628: if(oldState != newState){
    if(newState)
        RenderManager::getSingleton().rrAttachEntity(this);
    else
        RenderManager::getSingleton().rrDetachEntity(this);
}

tomluchowski:
Could you rename setTileCulling and setParentNodeDetach to setTileCullingFlag and setParentNodeDetachFlag, that would suggest the program reader that we are using set/ unset bit flags pattern in here ...

tomluchowski:
Hmm I tested your commit and the gain seems to be ~3 fps.

tomluchowski:
Now as we cull everything, we should disable Ogre3d side culling entirely ,,.,,

source/entities/Tile.cpp
hwoarangmy: ok

hwoarangmy:
> Could you rename setTileCulling and setParentNodeDetach to setTileCullingFlag and setParentNodeDetachFlag

ok

> Now as we cull everything, we should disable Ogre3d side culling entirely ,,.,,

Well, it might be worth trying even if out of the scope of this PR

source/entities/GameEntity.cpp
hwoarangmy, line 628: In this case, I don't agree because it is clear that oldState == newState means no change so it is obvious that there is nothing more to do.
IMHO, your suggestion adds 1 more imbrication and doesn't help reading.

tomluchowski, line 628: It does it means : to get here I must first meet condition oldState!=newState which must be deduced in your case. So if I am in RenderManager::getSingleton().rrAttachEntity(this); I explicitly know that newState is true and OldState is false.

tomluchowski:
Hmmm setting DefaultSceneManager speeds up nothing; if you wish just merge in this commit if I cannot convince you about nested if statements ...

hwoarangmy:
Well, I prefer to wait for @akien-mga or @Bertram25 to merge

tomluchowski:
You could also make setTileCullingFlag and setParentNodeDetachFlag a single argument function , by eliminating the boolean value. Just add the proper unsetParentNodeDetachFlag and unsetTileCullingFlag functions respectivly, where the boolean value was false .

oyvindln:
I haven't done any performance testing, but the game seems to run fine, so I can merge if you don't want to wait for @akien-mga or @Bertram25

PR #1260, opened 2016-09-13
Improved creature behaviour to stop working only when sleepy/hungry

hwoarangmy:


akien-mga:
Huh, I had the bad feeling for some days that I had left something unachieved here... a pending PR :D
Reviewing now.

akien-mga:
Looks good :)

PR #1262, opened 2016-10-27
Start of implementation of window and input handling using SFML

### LINKED ISSUE #1041: Key bindings not working on Linux X11
The keybindings do not work in 0.6.0. 

Windowed Mode, Debian and two displays.

oyvindln:
Not functional in windows yet, though that should be simple to fix.

Changing resolution in game is also not implemented (and the list of resolutions may not be working properly yet).

Enable it by setting OD_USE_SFML_WINDOW to true

This should hopefully fix #1041 and eventually let us avoid having to depend on OIS which is not really maintained any more.

Unfortunately, using SFML for the window may prevent us from using the D3D renderer (I don't know if it's currently enabled or not.)


CMakeLists.txt
```diff
@@ -483,8 +495,11 @@ find_package(Threads REQUIRED)
 find_package(OIS REQUIRED)
 find_package(OGRE REQUIRED)
 find_package(CEGUI REQUIRED)
-find_package(SFML 2 REQUIRED COMPONENTS Audio System Network)
-
+if(UNIX)
+    find_package(SFML 2 REQUIRED COMPONENTS Audio System Network Window Graphics)
```
akien-mga, line 495: The other dependencies should likely be added only if `OD_USE_SFML_WINDOW` is true.

CMakeLists.txt
```diff
@@ -198,6 +199,14 @@ if(WIN32)
     add_definitions(-DOIS_DYNAMIC_LIB)
 endif()
 
+if(OD_USE_SFML_WINDOW)
+    if(WIN32)
+        message(WARNING "SFML window is not implemented on windows yet!")
```
akien-mga: Is it not implemented on Windows upstream, or in our project?

oyvindln: It's just in our project, there are some small differences when the window is created, and I haven't done the windows specific part yet. (see http://www.ogre3d.org/forums/viewtopic.php?f=1&t=79384#p499268)

CMakeLists.txt
oyvindln: Yup.

oyvindln:
Seems to be a small issue when compiling on MSVC I have to fix too judginb by the build servers: 

```
c:\projects\opendungeons\source\modes\gamemode.cpp(81): error C2797: 'GameMode::mPreviousMousePosition': list initialization inside member initializer list or non-static data member initializer is not implemented [C:\projects\opendungeons\OpenDungeons.vcxproj]
```

oyvindln:
Ahh I acidentally merged this, but It's probably not an issue.

PR #1264, opened 2016-11-22
Added missing dependency

abenkovskii:
Not sure if it's required on all platforms or just on Linux and which version to recommend (1.3.0 worked for me), but without it the cmake refused to proceed on Fedora.


akien-mga:
Indeed, that's a required dependency.

PR #1269, opened 2016-12-27
Fixed bug hiding traps for all players + compilation fixes

hwoarangmy:


source/entities/Building.cpp
```diff
@@ -115,11 +115,19 @@ bool Building::canBuildingBeRemoved()
...
-        if(!p.second->mSeatsVision.empty())
+        for(Seat* seat : p.second->mSeatsVision)
         {
+            if(seat->getPlayer() == nullptr)
+                continue;
+            if(!seat->getPlayer()->getIsHuman())
+                continue;
+
             ret = false;
             break;
```
akien-mga, line 126: The `break;` is redundant here.

akien-mga:
Looks good :)

source/entities/Building.cpp
hwoarangmy: no, it's not the same for loop. The idea is that we want to check if a human player has seen the room when built and not since it is destroyed. As long as it is the case, we do not remove it.

PR #1275, opened 2017-11-15
Camera boundaries + optional Autoscroll

tomluchowski:
As in title, please merge in when you test my changes :)


oyvindln:
The autoscrolling works fine, except that it seems to only work upwards and to the left, but not to the right or downwards.

tomluchowski:
How about now , oln ?

oyvindln:
It works better now!

PR #1276, opened 2017-11-23
 Changed the write pass to HardwarePixelBuffer to be B R G instead of A B G R 

tomluchowski:
Now the code has the minimum changes to work with Ogre 1 10 9, please merge in when tested....


paroj:
you should rather refactor the code to use [PixelBox::setColourAt](https://ogrecave.github.io/ogre/api/1.10/class_ogre_1_1_pixel_box.html#ad9be8442d595e5208ea80e59041b59d8) which will always do the right thing.

oyvindln:
Yeah using that seems like a much better idea.

oyvindln:
Okay see #1277

PR #1277, opened 2017-11-27
Fix minimap pixel buffer issues.

oyvindln:
Yeah the commit history is a bit messy, but we can use squash and merge.

Use setPixelColourAt to make sure the pixel format is correct. We could probably improve some stuff here later.

### DIFF
--- a/source/gamemap/MiniMapDrawn.cpp
+++ b/source/gamemap/MiniMapDrawn.cpp
@@ -235,16 +235,16 @@ void MiniMapDrawn::update(Ogre::Real timeSinceLastFrame, const std::vector<Ogre:
 
     auto output = mPixelBuffer->lock(mPixelBox, Ogre::HardwareBuffer::HBL_NORMAL);
 
-    size_t x = 0;
-    size_t y = 0;
+    size_t xx = 0;
+    size_t yy = 0;
     for(const Color& color : mTiles)
     {
         // TODO: This is probably a bit inefficient at the moment.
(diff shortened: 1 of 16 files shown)


source/gamemap/MiniMapDrawn.cpp
```diff
@@ -233,14 +233,19 @@ void MiniMapDrawn::update(Ogre::Real timeSinceLastFrame, const std::vector<Ogre:
...
-    Ogre::uint8* pDest;
-    pDest = static_cast<Ogre::uint8*>(mPixelBuffer->getCurrentLock().data) - 1;
+    auto output = mPixelBuffer->lock(mPixelBox, Ogre::HardwareBuffer::HBL_NORMAL);
 
+    size_t xx = 0;
+    size_t yy = 0;
     for(const Color& color : mTiles)
     {
-        drawPixelToMemory(pDest, color.RR, color.GG, color.BB);
+        // TODO: This is probably a bit inefficient at the moment.
```
paroj, line 242: https://ogrecave.github.io/ogre/api/1.10/class_ogre_1_1_hardware_pixel_buffer.html#ae12e40bc3a6461d93b7bfae20e58c866

oyvindln: Yup that was what I was thinking about when writing that comment.

oyvindln:
Ok, so I'm merging this for now, though MinimapDrawnFull is still bugged on larger maps.

PR #1281, opened 2017-12-04
Camera proportional to its highness above the ground

tomluchowski:
as in title PLUS Added background picture of stars .

### DIFF
--- a/materials/scripts/Background.material
+++ b/materials/scripts/Background.material
@@ -1,0 +1,25 @@
+// Background
+material Background
+{
+    receive_shadows off
+    depth_write on
+    technique
+    {
+        pass
+        {
+            ambient 0.0 0.0 0.0 1.0
+            diffuse 0.8 0.8 0.8 1.0
+            specular 0.0 0.0 0.0 1.0 2.0
+            emissive 1.0 1.0 1.0 1.0
+            depth_write on
+            texture_unit
+            {
+                texture starsedit.jpg
+                tex_address_mode wrap
+                scale 1.0 1.0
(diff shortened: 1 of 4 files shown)


oyvindln:
I haven't tested the other changes yet, but for the picture there should be some license info (I'm going to assume you didn't take it yourself.)

EDIT: And it could probably be scaled down a bit to save on texture memory.

oyvindln:
Other than those issues the changes seem to work fine. I'm a bit unsure about the choice of background image though, maybe we could use something slightly less imposing, or darken the image a bit.

hwoarangmy:
hi @tomluchowski. Nice to see a PR with 1 comit per functionality, clear comments and that follows code conventions. I will try it to see what niceties you came with ^^

hwoarangmy:
It seems to work well. I guess your background is better than a black background so let's merge. Note that it is probably not related to this PR but I've noticed that when you go to the south, the camera doesn't allow to go under the first tile. But on north, i can go 1 screen above the last tile. My guess is that it is because we consider the tile on the bottom of the vision to allow or not the camera to move.

Bertram25:
Hey there, Happy New Year to everyone :D 

Sorry to be a bit the party crasher, is the picture open licensed?
Is it possible to add that in the credits file? https://github.com/OpenDungeons/OpenDungeons/blob/development/CREDITS

oyvindln:
Yeah, it's in the [CREDITS](https://github.com/OpenDungeons/OpenDungeons/blob/development/CREDITS#L209) file already.

PR #1283, opened 2018-01-01
Patch for 64 bit systems, where toString(uint64_t d) and toString(siz…

cyruscousins:
…e_t d) were overlapping declarations (and definitions), breaking the build.

I remove these definitions from the 64 bit build.  There's probably a cleaner way to do this, but this should leave 32 bit builds alone while fixing 64 bit builds.


oyvindln:
Seems to work fine, merging.

PR #1286, opened 2018-04-17
Unbreak build on 32bit platforms

jbeich:
#1283 fixed only x86_64 but x86 is still broken:
```c++
source/utils/Helper.cpp:278:17: error: redefinition of 'toString'
    std::string toString(size_t d)
                ^
source/utils/Helper.cpp:265:17: note: previous definition is here
    std::string toString(uint32_t d)
                ^
1 error generated.
```

@devnexen, can you check this PR doesn't break OpenBSD build?

### DIFF
--- a/source/utils/Helper.cpp
+++ b/source/utils/Helper.cpp
@@ -274,7 +274,7 @@ namespace Helper
     {
         return TTostring(d);
     }
-    #ifndef __x86_64
+    #if defined(__OpenBSD__) && defined(__LP64__)
     std::string toString(size_t d)
     {
         return TTostring(d);
--- a/source/utils/Helper.h
+++ b/source/utils/Helper.h
@@ -138,7 +138,7 @@ namespace Helper
     std::string toString(uint32_t d);
     std::string toString(int64_t d);
     std::string toString(uint64_t d);
-    #ifndef __x86_64
+    #if defined(__OpenBSD__) && defined(__LP64__)
     std::string toString(size_t d);
     #endif
     std::string toString(const Ogre::Vector3& v);


devnexen:
It does but not because of this change but most likely because C++11 mode as overrides methods need to have override "annotation" but will look further later.

devnexen:
The last change breaks indeed ... with size_t it works better.

devnexen:
Maybe in order to solve your issue might be better to cast some calls parameters (do not know if it is possible or too tedious ... did not really look), what do you think ?

devnexen:
I pushed a PR https://github.com/OpenDungeons/OpenDungeons/pull/1288 let me know what you think.

jbeich:
#1288 looks more ugly than the statu quo. Let's drop 75888ad91ee4 for now. I don't know C++ enough, anyway.

jbeich:
Does it build fine on OpenBSD now? If so I'd like this merged.

devnexen:
Your branch ? Did not try yet.

devnexen:
It seems ... I ve built only Helper.o object but the rest still stands e.g. override markers which I addressed in my own PR (e.g. I use clang 6).

jbeich:
@devnexen, can you try to build this PR on OpenBSD again?

devnexen:
Sorry to repeat but can t build because of numerous missing override markers :-) your Helper changes seem ok though but unless callers are compiled ... can t know for sure.

jbeich:
`override` issue is unrelated. Feel free to merge it with this PR when testing. I'm getting really confused by your feedback. Did`toString(unsigned long long d)` actually work? If so I'd like to avoid ugly conditionals.

devnexen:
Just done that seems working.

jbeich:
Can you drop `toString` changes from #1288 to avoid merge conflict?

jbeich:
... to avoid merge conflict.

devnexen:
done.

hwoarangmy:
Hi, thanks for the PR. I will have a look ASAP

hwoarangmy:
Looks good. Thanks for the PR

PR #1288, opened 2018-04-17
override marker added to missing methods.

devnexen:
few casts left.

### DIFF
--- a/source/network/ODServer.cpp
+++ b/source/network/ODServer.cpp
@@ -1507,20 +1507,20 @@ bool ODServer::processClientNotifications(ODSocketClient* clientSocket)
                         // We look for the Skirmish or multiplayer prefix and keep it.
                         uint32_t indexSk = fileLevel.find(SAVEGAME_SKIRMISH_PREFIX);
                         uint32_t indexMp = fileLevel.find(SAVEGAME_MULTIPLAYER_PREFIX);
-                        if((indexSk != std::string::npos) && (indexMp == std::string::npos))
+                        if(((size_t)indexSk != std::string::npos) && ((size_t)indexMp == std::string::npos))
                         {
                             // Skirmish savegame
                             ss << SAVEGAME_SKIRMISH_PREFIX;
                             ss << fileLevel.substr(indexSk + SAVEGAME_SKIRMISH_PREFIX.length());
 
                         }
-                        else if((indexSk == std::string::npos) && (indexMp != std::string::npos))
+                        else if(((size_t)indexSk == std::string::npos) && ((size_t)indexMp != std::string::npos))
                         {
                             // Multiplayer savegame
                             ss << SAVEGAME_MULTIPLAYER_PREFIX;
                             ss << fileLevel.substr(indexMp + SAVEGAME_MULTIPLAYER_PREFIX.length());
                         }
-                        else if((indexSk != std::string::npos) && (indexMp != std::string::npos))
+                        else if(((size_t)indexSk != std::string::npos) && ((size_t)indexMp != std::string::npos))
                         {
                             // We found both prefixes. That can happen if the name contains the other
                             // prefix. Because of filename construction, we know that the lowest is the good
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -151,8 +151,11 @@ else()
     if(OD_ENABLE_WARNINGS)
         # Help getting compilation warnings
         set(OD_OPT_FLAGS "${OD_OPT_FLAGS} -Wall -Wcast-align -Wcast-qual -Wctor-dtor-privacy -Wdisabled-optimization -Wold-style-cast")
-        set(OD_OPT_FLAGS "${OD_OPT_FLAGS} -Woverloaded-virtual -Wredundant-decls -Wswitch-default -Wstrict-null-sentinel -Winvalid-pch")
+        set(OD_OPT_FLAGS "${OD_OPT_FLAGS} -Woverloaded-virtual -Wredundant-decls -Winvalid-pch")
         set(OD_OPT_FLAGS "${OD_OPT_FLAGS} -Wswitch-default -Wstack-protector")
+        if (NOT CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+          set(OD_OPT_FLAGS "${OD_OPT_FLAGS} -Wstrict-null-sentinel")
+        endif()
         #for later https://stackoverflow.com/questions/5088460/flags-to-enable-thorough-and-verbose-g-warnings/9862800#9862800:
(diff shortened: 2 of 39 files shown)


source/utils/Helper.cpp
```diff
@@ -274,11 +274,16 @@ namespace Helper
     {
         return TTostring(d);
     }
-    #ifndef __x86_64
+    #if !defined(__LP32__) || defined(__OpenBSD__)
```
jbeich: `__LP32__` doesn't exist and `!defined(__ILP32__)` would break x86_64 again. `size_t` is `uint64_t` aka `unsigned long` on Linux/FreeBSD x86_64.

source/utils/Helper.cpp
```diff
@@ -274,11 +274,16 @@ namespace Helper
...
     std::string toString(size_t d)
     {
         return TTostring(d);
     }
+    #else
+    std::string toString(unsigned long long d)
```
jbeich: Unused on non-OpenBSD and may break on build on 32bit platforms where `uint64_t` is `unsigned long long` but `toString(uint64_t d)` is already defined earlier.

source/utils/Helper.cpp
devnexen: The only alternative I see is then creating special rules for OpenBSD and general rules for the rest. __LP32__ does not really exist just realise clang create this macro possibly.

jbeich: If build broke because `size_t` != `uint64_t` on 64bit platform then `toString(size_t d)` is useful.

source/utils/Helper.cpp
```diff
@@ -262,23 +262,42 @@ namespace Helper
     {
         return TTostring(d);
     }
+    std::string toString(int64_t d)
+    {
+        return TTostring(d);
+    }
+    #ifndef __OpenBSD__
```
jbeich: Better move `toString(size_t d)` undef `#if defined(__OpenBSD__) && defined(__LP64__)` then drop `#ifdef __LP64__`.

jbeich (changes requested):
Before 8568c7d14fb8 the code worked fine **as is** on both x86 and x86_64. I'm not sure why you're trying to complicate ifdefs.

source/utils/Helper.cpp
devnexen: I propose we merge your PR then will apply your changes to this one keeping the override fixes ... What do you think ?

jbeich: Sorry, I had a typo in my comment (edited). OpenBSD probably only needs `toString(size_t d)` only on 64bit platforms.

jbeich: If my #1286 introduces a regression on OpenBSD then it shouldn't be merged.

devnexen: Yes ideally the toString might be better as generics but that would require too much changes I fear.

devnexen: I ve tested my branch on FreeBSD and OpenBSD 64 bits so far by the way.

jbeich: i386 or x86 (32bit) is not dead yet. It has Tier1 support on FreeBSD.
```c++
source/gamemap/TileContainer.cpp:400:52: error: call to 'toString' is ambiguous
            OD_LOG_ERR("Unknown neighbor index=" + Helper::toString(i));
                                                   ^~~~~~~~~~~~~~~~
source/utils/LogManager.h:32:155: note: expanded from macro 'OD_LOG_ERR'
#define OD_LOG_ERR(_message)                      LogManager::getSingleton().logMessage(LogMessageLevel::CRITICAL, __FILE__, __LINE__, (std::string("") + _message))
                                                                                                                                                          ^~~~~~~~
source/utils/Helper.h:131:17: note: candidate function
    std::string toString(float f, unsigned short precision = 6);
                ^
source/utils/Helper.h:132:17: note: candidate function
    std::string toString(double d, unsigned short precision = 6);
                ^
source/utils/Helper.h:133:17: note: candidate function
    std::string toString(int8_t d);
                ^
source/utils/Helper.h:134:17: note: candidate function
    std::string toString(uint8_t d);
                ^
source/utils/Helper.h:135:17: note: candidate function
    std::string toString(int16_t d);
                ^
source/utils/Helper.h:136:17: note: candidate function
    std::string toString(uint16_t d);
                ^
source/utils/Helper.h:137:17: note: candidate function
    std::string toString(int32_t d);
                ^
source/utils/Helper.h:138:17: note: candidate function
    std::string toString(int64_t d);
                ^
source/utils/Helper.h:144:17: note: candidate function
    std::string toString(uint64_t d);
                ^
```

devnexen: Ok so since your PR for Helpers change seem to work (tested locally your changes with my own branch) might be good to merge yours then will rebase this one. What do you think ?

jbeich: If the changes here won't touch `toString` helpers simply rebase against  `development` branch. No need to wait for #1286 to be merged. Obviously, history would look better if you squash reverts, split unrelated bits into separate commits and document compiler error messages.

devnexen: OK will try later in the day. Cheers.

devnexen:
I hope it looks a bit better now. Sorry for the overly long review. Regards.

stavrossk:
Hi, thanks for the great game. Does the 32-bit Windows version work on 64-bit Windows?

hwoarangmy:
Hi, thanks for the PR. I will have a look ASAP

source/network/ODServer.cpp
```diff
@@ -1507,20 +1507,20 @@ bool ODServer::processClientNotifications(ODSocketClient* clientSocket)
                         // We look for the Skirmish or multiplayer prefix and keep it.
                         uint32_t indexSk = fileLevel.find(SAVEGAME_SKIRMISH_PREFIX);
                         uint32_t indexMp = fileLevel.find(SAVEGAME_MULTIPLAYER_PREFIX);
-                        if((indexSk != std::string::npos) && (indexMp == std::string::npos))
+                        if(((size_t)indexSk != std::string::npos) && ((size_t)indexMp == std::string::npos))
```
hwoarangmy, line 1510: I guess it would have been better to define indexSk as a size_t instead of casting every time there is a compare

hwoarangmy:
The PR looks good. Merging

source/network/ODServer.cpp
hwoarangmy: Actually, it seems like Travis build fails because of that. I'm fixing it

devnexen: Yes I pushed a change too late few minutes after you merged :-) but feel free.

hwoarangmy: I already created the PR, thanks ^^

Issue #1280, opened 2017-12-03
Accommodate Ogre 1.11 deprecations and changes


oyvindln:
- [ ] Replace `Ogre::SharedPtr::isNull` and related since it's deprecated.
- [ ] Do camera operations on scene node camera is attached to (doing it directly is deprecated)
- [ ] Do something about the large number of string lookups of nodes, as [`OGRE_NODE_STORAGE_LEGACY`](https://www.ogre3d.org/2017/09/02/ogre-1-10-mid-term-report) will no longer be supported in 1.11 and using `OGRE_NODE_STORAGE_LEGACY=OFF` makes loading large maps extremely, using time in string lookups when calling `mSceneManager->hasEntity` in  `RenderManager::rrRefreshTile`.
- [ ] Look into using the new HLMS, which should make materials much easier to deal with.
- [ ] Try to get [this](https://bitbucket.org/cegui/cegui/issues/1157/ogre-renderer-does-not-work-with-ogre-110) CEGUI issue fixed as it prevents using c++11 enabled Ogre which will start being the default.

akien-mga:
Ogre 1.11 was released, and one more thing to fix would be:
```
/home/iurt/rpmbuild/BUILD/opendungeons-0.7.1/source/render/ODFrameListener.h:33:10: fatal error: OgreWindowEventUtilities.h: No such file or directory
 #include <OgreWindowEventUtilities.h>
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
(edited in OP)

> Try to get [this](https://bitbucket.org/cegui/cegui/issues/1157/ogre-renderer-does-not-work-with-ogre-110) CEGUI issue fixed as it prevents using c++11 enabled Ogre which will start being the default.

I think they might have fixed this in the `v0-8` branch without checking/updating your issue, so you might want to try again. I just built the tip of the `v0-8` branch against Ogre 1.11 without issues (you can see that there were some patches for compatibility with Ogre 1.10/1.11 in March 2018).

----

Since distro packages will start upgrading their Ogre version (as I did for Mageia), they'll run into issues building OpenDungeons, so I think we should have a good at this soon-ish and make a 0.7.2 release once we've checked that the new version works well.

[...]

PR #1291, opened 2018-06-03
Fix build with OGRE >= 1.11.0

akien-mga:
Initial patch by @wally-mageia to fix build against OGRE >= 1.11.0, I have yet to test it myself (building the new OGRE locally on Mageia 6).

Wally just told me:

> \<wally_\> Akien: actually we cheat a bit as we're using -Wno-deprecated-declarations in CXXFLAGS
\<wally_\> so my patch isn't complete

But I'm putting it up nevertheless to start the process of porting OD to OGRE 1.11.
@oyvindln mentioned he had started some work on this too and might already be some steps further, we'll see.

Travis and AppVeyor should both fail as they're still building against OGRE 1.11.

Part of #1280.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -514,7 +514,7 @@ if(OD_USE_SFML_WINDOW)
 else()
     find_package(SFML 2 REQUIRED COMPONENTS Audio System Network)
 endif()
-if((OGRE_VERSION_MAJOR LESS 1) AND (OGRE_VERSION_MINOR LESS 9))
+if("${OGRE_VERSION}" VERSION_LESS "1.9.0")
     message(FATAL_ERROR "OGRE version >= 1.9.0 required")
 endif()
 
@@ -702,7 +702,14 @@ else()
     message(STATUS "Plugin path dbg: " ${OD_OGRE_PLUGIN_DIR_DBG})
 endif()
 
-#Do the configuration
+# Do the configuration
+# Convoluted check because old CMake has no VERSION_GREATER_EQUAL,
+# we want this true for 1.11 or later since plugins have to be declared manually.
+if("${OGRE_VERSION}" VERSION_LESS "1.11")
+    set(OD_OGRE_VERSION_SPECIFIC_PLUGINS "")
+else()
+    set(OD_OGRE_VERSION_SPECIFIC_PLUGINS "Plugin=Codec_FreeImage")
+endif()
 configure_file(${CMAKE_CONFIG_DIR}/plugins.cfg.in ${CMAKE_BINARY_DIR}/plugins.cfg)
 if(WIN32)
     configure_file(${CMAKE_CONFIG_DIR}/plugins_d.cfg.in ${CMAKE_BINARY_DIR}/plugins_d.cfg)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -626,6 +626,7 @@ target_link_libraries(
 
     #libraries
     ${OGRE_LIBRARIES}
+    ${OGRE_Bites_LIBRARIES}
     ${OGRE_RTShaderSystem_LIBRARIES}
     ${OIS_LIBRARIES}
     ${CEGUI_LIBRARIES}
--- a/source/render/MovableTextOverlay.cpp
+++ b/source/render/MovableTextOverlay.cpp
@@ -38,7 +38,8 @@ ChildOverlay::ChildOverlay(const Ogre::String& fontName, Ogre::Real charHeight,
     mForcedHeight(-1),
     mCharHeight(charHeight),
     mTimeToDisplay(0),
-    mFont(Ogre::FontManager::getSingleton().getByName(fontName))
+    // FIXME: Move FontManager usage to ResourceManager somehow and dehardcode "GUI"?
+    mFont(Ogre::FontManager::getSingleton().getByName(fontName, "GUI"))
 {
 #if defined(OGRE_VERSION) && OGRE_VERSION < 0x10A00
     if (mFont.isNull())
--- a/cmake/config/plugins.cfg.in
+++ b/cmake/config/plugins.cfg.in
@@ -7,6 +7,9 @@ PluginFolder=@OD_OGRE_PLUGIN_DIR_REL@
 Plugin=RenderSystem_GL
 Plugin=Plugin_ParticleFX
 
+# Version-specific plugins (e.g. codecs for 1.11+)
+@OD_OGRE_VERSION_SPECIFIC_PLUGINS@
+
 #This is a card specific library, installed on some systems but not on others
 #Cg is also obsolete and nonfree
 #Plugin=Plugin_CgProgramManager
--- a/cmake/config/plugins_d.cfg.in
+++ b/cmake/config/plugins_d.cfg.in
@@ -7,6 +7,9 @@ PluginFolder=@OD_OGRE_PLUGIN_DIR_DBG@
 Plugin=RenderSystem_GL_d
 Plugin=Plugin_ParticleFX_d
 
+# Version-specific plugins (e.g. codecs for 1.11+)
+@OD_OGRE_VERSION_SPECIFIC_PLUGINS@
+
 #This is a card specific library, installed on some systems but not on others
 #Cg is also obsolete and nonfree
 #Plugin=Plugin_CgProgramManager_d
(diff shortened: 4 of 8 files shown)


CMakeLists.txt
```diff
@@ -513,8 +513,8 @@ if(OD_USE_SFML_WINDOW)
 else()
     find_package(SFML 2 REQUIRED COMPONENTS Audio System Network)
 endif()
-if((OGRE_VERSION_MAJOR LESS 1) AND (OGRE_VERSION_MINOR LESS 9))
-    message(FATAL_ERROR "OGRE version >= 1.9.0 required")
+if("${OGRE_VERSION}" VERSION_LESS "1.11.0")
```
akien-mga: Previous check was bogus BTW, it wouldn't fail if OGRE was 1.6.0 :P

akien-mga:
BTW, reflecting on how it took *years* for Debian to package a suitable version of CEGUI for OD, and knowing that the same might happen with OGRE 1.10/1.11 given the compatibility breakage, it might be worth keep OGRE 1.9 support if it doesn't require too many hacks.

But let's work first on OGRE 1.11 compat, and we can then see what would be involved to keep OGRE 1.9 compat too.

hwoarangmy:
It would also be nice to update appveyor/travis files to build against ogre 1.10 so that it is easy to know if a PR compiles or not.
And if the game should work with ogre 1.9, a build variant would allow to test both.

oyvindln:
I do actually have it up and running without "-Wno-deprecated-declarations", but I got some issues with divide-by-zero in the culling functions, so I didn't make a PR yet.

tomluchowski:
@oyvindln : Ehem you mean mine code gives a divide by zero  .... ? (<embarassed>)

CMakeLists.txt
paroj: your life will be much easier if you just drop the FindOGRE.cmake script, see
https://ogrecave.github.io/ogre/api/latest/setup.html

akien-mga:
I'll rework this in coming days, for now I'm making sure that everything builds and works fine with both 1.9.x and 1.10.x.

akien-mga:
Rebased and simplified, it should now still build with Ogre 1.9, 1.10 and 1.11+.

I've tested the build successfully with 1.11.0, though I got this weird issue:
```
An exception has occurred: ItemIdentityException: Can not find codec for 'png' image format.
Supported formats are: dds ktx pkm pvr. in Codec::getCodec at /home/iurt/rpmbuild/BUILD/ogre-1.11.0/OgreMain/src/OgreCodec.cpp (line 66)
```
Seems like an issue with my distro's Ogre package, I'm now packaging 1.11.2 and checking if/why the PNG support is broken.

akien-mga:
Turns out PNG support was moved to opt-in plugins in Ogre 1.11; I hacked the `plugins.cfg` generation to handle it, but we could also do it directly from code to avoid relying on this config file (see e.g. patch in https://aur.archlinux.org/packages/gazebo/?comments=all&O=0&PP=10 ).

akien-mga:
Next 1.11+ issue is the following:
```
An exception has occurred: FileNotFoundException: Cannot locate resource fonts/LiberationSans-10.font in resource group General. in ResourceGroupManager::openResource at /home/akien/Mageia/Checkout/cauldron/ogre/BUILD/ogre-1.11.2/OgreMain/src/OgreResourceGroupManager.cpp (line 698)
```

It might be related to what is mentioned here: https://github.com/OGRECave/ogre/blob/master/Docs/1.11-Notes.md#cmake-build
> New OGRE_RESOURCEMANAGER_STRICT options:
>
> - legacy mode that searches through all groups (slow)
> - pedantic mode that forces you to specify a resource group (cumbersome)
> - new strict mode that defaults to the default group

Ogre now builds by default with the strict mode (2), while I suppose the previous mode was the now deprecated 0 one, which looks up for resources everywhere. So we need to better define how to find our font resources, but I'm not sure how it's done (@Bertram25?).

paroj:
I guess its in "GUI"
https://github.com/OpenDungeons/OpenDungeons/blob/6bc6e053fb64c6109cae9473519cb0999b267baf/cmake/config/resources.cfg.in#L9-L12

Bertram25:
@akien-mga I don't recall a thing on this one. I'll need to test but it would have to wait at least one more week on my side. (Sry :S)
@paroj 's suggestion looks relevant. :) Safe enough to try, I guess.

akien-mga:
"GUI" is indeed the correct resource location, but I'm not sure where and how to define that I want to use it for loading fonts?

```
(13:07:32) (ResourceManager) [NORMAL] Resource in section: GUI Type: FileSystem Path: ./gui
(13:07:32) (ResourceManager) [NORMAL] Resource in section: GUI Type: FileSystem Path: ./gui/fonts
(13:07:32) (ResourceManager) [NORMAL] Resource in section: GUI Type: FileSystem Path: ./gui/schemas
(13:07:32) (ResourceManager) [NORMAL] Resource in section: Graphics Type: FileSystem Path: ./materials/RTShaderLib
(13:07:32) (ResourceManager) [NORMAL] Resource in section: Graphics Type: FileSystem Path: ./materials/RTShaderLib/materials
(13:07:32) (ResourceManager) [NORMAL] Resource in section: Graphics Type: FileSystem Path: ./materials/scripts
(13:07:32) (ResourceManager) [NORMAL] Resource in section: Graphics Type: FileSystem Path: ./materials/textures
(13:07:32) (ResourceManager) [NORMAL] Resource in section: Graphics Type: FileSystem Path: ./models
(13:07:32) (ResourceManager) [NORMAL] Resource in section: Graphics Type: FileSystem Path: ./particles
(13:07:32) (ResourceManager) [NORMAL] Resource in section: Music Type: FileSystem Path: ./music
(13:07:32) (ResourceManager) [NORMAL] Resource in section: Sound Type: FileSystem Path: ./sounds
```

akien-mga:
@Bertram25 Any idea how to specify what resource section to use for give resources? My limited Ogre knowledge is starting to show :P

paroj:
its the groupName argument used in various places e.g. here https://ogrecave.github.io/ogre/api/latest/class_ogre_1_1_font_manager.html#a8d7fce6d9152fd1830d98e3235a8756f

akien-mga:
Thanks @paroj, I could fix it with this hint. Actually the first font-related crash I was stuck on was due to CEGUI not being able to resolve `fonts/*.ttf` even though both `gui` and `gui/fonts` are in the resource group... Removing `fonts/` from the CEGUI scheme fixed it.

Then yeah I had to use "GUI" in `FontManager::getByName` to fix a crash when starting a game.

akien-mga:
So I'm considering this ready to merge in that it does fix building with OGRE 1.11+, and the game is playable.

There are some issues that will need to be investigated further though, e.g. the resources overlay in-game no longer shows up (might be another CEGUI configuration issue though).

I've tested with Ogre 1.10.12 again to make sure that everything works as it should with this PR merged.

Bertram25:
@akien-mga Sorry, a bit late on that one. Cool that you could fix it. :) I'll be setupping things with Ogre 1.11 this week-end so that next time, I can get some clues faster.

source/render/MovableTextOverlay.cpp
```diff
@@ -38,7 +38,8 @@ ChildOverlay::ChildOverlay(const Ogre::String& fontName, Ogre::Real charHeight,
     mForcedHeight(-1),
     mCharHeight(charHeight),
     mTimeToDisplay(0),
-    mFont(Ogre::FontManager::getSingleton().getByName(fontName))
+    // FIXME: Move FontManager usage to ResourceManager somehow and dehardcode "GUI"?
```
Bertram25, line 41: Good point, I guess this could be wrapped up into some ResourceManager accessor, so that the "GUI" string could come from const of the same location?
See: https://github.com/OpenDungeons/OpenDungeons/blob/development/source/utils/ResourceManager.cpp#L77

E.g: (No idea if this works :])
```c++
const std::string ResourceManager::RESOURCEGROUPGUI = "GUI";
...
FontPtr ResourceManager::getFontByName(const std::string& fontName) const
{
    return Ogre::FontManager::getSingleton().getByName(fontName, );
}
...
mFont(ResourceManager::getSingleton().getFontByName(fontName));
```

akien-mga:
Superseded by #1300.

PR #1292, opened 2018-06-03
CMake: Drop unused and obsolete CPack config

akien-mga:


akien-mga:
BTW @oyvindln, I see you've used the "Squash and merge" option to merge this one. We don't have strong guidelines for its usage as far as I remember, but I tend to prefer a normal merge with merge commit if the PR already has a correct git history - and squash and merge only when there's a lot of history that we want to get rid of in the PR.

oyvindln:
Noted

Issue #1268, opened 2016-12-27
Remove dependency to OIS


hwoarangmy:
@oyvindln I think you have started working on this point. Are you still working on it ?

oyvindln:
I started on it yea, and committed some stuff, I haven't worked much on OD lately though.

hwoarangmy:
wow, that's fast answering ^^
Are you still on it ? Do you want to assign this issue ? If you want, I can help as I should have some free time.

hwoarangmy:
@oyvindln BTW, I've tried your PR and it seems I get some weird results with windows
![sfml_window](https://cloud.githubusercontent.com/assets/7493706/21505067/57b76180-cc64-11e6-8815-ef63962243ce.png)

Please tell me if you want me to test something

oyvindln:
Hm, no idea why it would be like that. Does it happen if OD_USE_SFML_WINDOW is not set to ON(in cmake)?

hwoarangmy:
If I don't set it, display is as it used to be. Note that I've tried after compiling with MinGW. I've not tried with MSVS2013. Have you ?

oyvindln:
I used MSVS2015 when I was testing on windows. I can't remember having that issue, but I'll have a look.

oyvindln:
I seem to have some path issue that makes some of the file paths contain a null character, I will have to try to solve that first before it's possible to test.

hwoarangmy:
well, don't worry about that, i will test tomorrow. I have functional environments with MSVS2013 and MinGW (which are what we said we officially support). I will tell you if the issue is the same or not.

hwoarangmy:
I've tried the MSVS2013 version. With OpenGL renderer, I get the same behaviour. With the D3D renderer, as expected, it doesn't work. I've not checked what happens in details but since we explicitely use the open gl renderer, I didn't expect it to work.
Note that the glitch also happens in the main screen so it is not related to ingame settings. It is like something is messed with the z buffer and meshes overlap.

hwoarangmy:
@oyvindln I've had a quick look and it seems to work when creating the SFML window with a dept of 32:
```
sf::RenderWindow sfmlWindow(sf::VideoMode(w, h, sf::VideoMode::getDesktopMode().bitsPerPixel)
, windowTitle, style, sf::ContextSettings(32, 0, 2, 1));
```

Does it also work for you ?
[...]

PR #1295, opened 2018-10-23
OIS: Remove dir name from includes

akien-mga:
With the release of OIS 1.4, upstream renamed their include folder
from `/usr/include/OIS` to `/usr/include/ois`, so our includes break.

`/usr/include/ois` is already part of their pkgconfig include path
so we get it via CMake.

Needs testing on Windows (@hwoarangmy?).


akien-mga:
Should fix what @goupildb reported in https://github.com/OpenDungeons/OpenDungeons/issues/1268#issuecomment-385430296 (and where I mistakenly blamed Debian for it, but it's actually upstream OIS that did this weird compat breakage).

akien-mga:
Rebased.

akien-mga:
Prior to this commit we already had one include that did not use the `OIS/` folder, so I assume that it will work fine on Windows.
```
$ rg 'include [<"]OIS[^/]'
source/render/Gui.h
29:#include <OISMouse.h>
```

PR #1300, opened 2018-12-26
Ogre 1.11 support + snapcraft package

paroj:
includes/ replaces #1291 

https://snapcraft.io/opendungeons


source/utils/ResourceManager.cpp
```diff
@@ -133,6 +134,8 @@ void ResourceManager::setupDataPath(boost::program_options::variables_map& optio
         {
             mGameDataPath.append("/");
         }
+
+        mGameDataPath = Ogre::FileSystemLayer::resolveBundlePath(mGameDataPath);
```
paroj, line 139: this can also resolve the mac bundle path, but I cannot test that and I wanted to keep changes minimal

Bertram25:
Looks sane to me. Is it ok to merge guys?

tomluchowski:
Would love to see OD working on Ogre 1.11, however I don't have Ogre 1.11 package yet and I ' d  have to ask my fellow packager to build it after it is certain that OD works. That's why I stay silent .

paroj:
if your distribution supports snaps, you can get a prebuild package of OD+Ogre1.11 from the link above.

(automatically built via https://snapcraft.io/build)

tomluchowski:
thanks for the snap tip, it installs fine.
However the package seems to have failures -- first of all the ambient light is not set or game colors with ogre3d 1 11 0  looks especially bleach.

Second flaw I noticed is the camera no longer moving properly - I don''t even mention my autoscroll feature -- but the keyboard A W S D does not work neither. ...

paroj:
od1 11
this is how it looks for me. Seemed pretty similar to the screenshots section. Also WASD works.

Which distribution do you use? Drivers? Ogre log?

paroj:
autoscroll only works in right or down directions for me.

tomluchowski:
I am on OPenSuse 15 + nvidia prop. drivers

Well, to be honest I have two LCD's and it seems on the old one the colors are just fine ....
After setting 'grab keyboard ' , 'grab mouse' scrolling and A W S D works fine .
Also the glitches from my culling code seems to be less visible ! :), did you change my code, or it is an ogre3d improvment in attaching/ detaching tiles ... ? 
The only cons now is it won't load large maps -- the 250x250 and 400x400 .... and whether it was working last time, I don't remember cause I made my branch ( and it happens from master / development as well)  to crash on free'ing a null pointer ( 0 address) after choosing ANY map . The only working version of OD is your version ( plus the one in Opensuse builds) but nothing works that I have build my own. 

Would love to have OD code compiling and working again ... so maybe let's merge that patch in ? 



P.S. I always  thought about you as a very busy person on ogre3d forum , and you find time for OD , respect to you :+1:

paroj:
the only changes are made to OD are the ones you see in the diff here.

the 250x250 map still loads, but it takes several minutes now due to #1280

you can create your custom build using the included snapcraft.yml. Though I have no idea how well it works on OpenSuse.
You could try to use a persistent Ubuntu Container as in https://snapdocs.labix.org/clean-build-using-lxc/4157
Alternatively you can abuse https://snapcraft.io/build to build & publish on each commit.

Yeah.. from time to time I still have to check whether anybody is still using Ogre..

akien-mga:
Thanks @paroj, the additional changes look good!

source/utils/ResourceManager.cpp
```diff
@@ -133,6 +134,10 @@ void ResourceManager::setupDataPath(boost::program_options::variables_map& optio
         {
             mGameDataPath.append("/");
         }
+
+#if OGRE_VERSION > 0x10900
```
akien-mga, line 138: For the reference, this check was incorrect, as it would be true for 1.9.1 and later until 1.10.0.
Fixed in #1303.

PR #1302, opened 2019-02-08
More tile types for the same neightbourhood; creatures catch and drop at the point pointed by the "hand" fingers

tomluchowski:
As in title , now it is possible to more then one tile type for the given neightbourhood index. To do this it is enough to copy your model and materials file(s) in appropriate directory. Then somewhere in the config/tileset.cfg change : e.g. : 

        1111	Dirt_fl_1111.mesh	Dirt	0.0	0.0	90.0

into 

        1111
        [oneOf]
	Dirt_fl_1111A.mesh	Dirt	0.0	0.0	90.0
	Dirt_fl_1111B.mesh	Dirt	0.0	0.0	90.0
        [/oneOf]

if you have Dirt_fl_1111A.mesh and Dirt_fl_1111B.mesh prepared in the model diretory.

Second feature is we now can pickup and drop creatures by the point pointed by hand fingers, before it was pretty messed up, the game took the plane of highness 1,5 instead of 0,0 found the intersection and then claimed it was intersection with the Ground ( x = 0 . y = 0 ) .....


tomluchowski:
@hwoarangmy , @Bertram25 : Could you check this out and merge, as well as other outstanding PR's ?

akien-mga:
For the reference: we're working with @tomluchowski on splitting this PR into three separate PRs for the unrelated changes, to make them easier to review and merge independently.

PR #1306, opened 2019-02-21
Multiple tile types for the same neighbourhood

tomluchowski:
As in title, now it is possible to more then one tile type for the given neighbourhood index. To do this it is enough to copy your model and materials file(s) in appropriate directory. Then somewhere in the config/tileset.cfg change : e.g. :

```
    1111	Dirt_fl_1111.mesh	Dirt	0.0	0.0	90.0
```

into

```
    1111
    [oneOf]
Dirt_fl_1111A.mesh	Dirt	0.0	0.0	90.0
Dirt_fl_1111B.mesh	Dirt	0.0	0.0	90.0
    [/oneOf]
```

if you have Dirt_fl_1111A.mesh and Dirt_fl_1111B.mesh prepared in the model diretory.


akien-mga:
Copied the description over from #1302.

PR #1307, opened 2019-02-21
Camera sliding

tomluchowski:
This PR removes the bug of constant camera sliding even when no mouse button was pressed. Now you can randomly look around the scene with middle-mouse button without causing the camera to slide.


akien-mga:
Thanks!

PR #1308, opened 2019-02-21
creatures catch and drop at the point pointed by the "hand" fingers

tomluchowski:
We now can pickup and drop creatures by the point pointed by hand fingers, before it was pretty messed up, the game took the plane of highness 1,5 instead of 0,0 found the intersection and then claimed it was intersection with the Ground ( x = 0 . y = 0 ) .....


max-carroll (approved):
changes look good

PR #1309, opened 2019-02-28
Resized buttons, its size should be now proportional to the display size

tomluchowski:
As in title resized buttons, maybe worth noting is that their sizes is exactly the percentage of a display not percentage PLUS some constant as it is possible with FALAGARD/CEGUI ...

### DIFF
--- a/gui/ModeEditor.layout
+++ b/gui/ModeEditor.layout
@@ -106,8 +106,8 @@
             <Property name="AlwaysOnTop" value="True" />
         </Window>
         <Window type="OD/TabControl" name="MainTabControl" >
-            <Property name="Area" value="{{0,0},{1,-133},{1,-200},{1,0}}" />
-            <Property name="TabHeight" value="{0,35}" />
+            <Property name="Area" value="{{0,0},{0.80,0},{1,0},{1,0}}" />
+            <Property name="TabHeight" value="{0.15,0}" />
             <Property name="RiseOnClickEnabled" value="False" />
             <Window type="DefaultWindow" name="Tiles" >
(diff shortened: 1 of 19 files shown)


akien-mga:
It looks pretty messed up IMO:
screenshot_20190304_092627

That's on 1920x1080.

tomluchowski:
Why ? I asked dozens of times what new outlook should be .... Most of the people said it doesn't mind non-square buttons .... btw write in the console uncheckall or something to get all the rooms. The remove button is a special button that's why I put it in "special " position ....

tomluchowski:
@Bertram25 seems you are the last hope for those three PR's to merge in, care to see and push them ?

akien-mga:
> Why ? I asked dozens of times what new outlook should be .... Most of the people said it doesn't mind non-square buttons .... btw write in the console uncheckall or something to get all the rooms. The remove button is a special button that's why I put it in "special " position ....

Can't you see in my screenshot what's wrong with the current state?

- Action buttons are *huge* compared to the rest of the UI
  * Tab button are still small, tabs themselves are small
  * Minimap buttons are still small, and even get out of the bottom menu due to the change
  * Upper menu buttons are also super small
  * Action buttons overflow their bounding UI panel

I'm not against upscaling buttons, but it needs to be done properly. Everything should be scaled up, while making sure that the GUI elements can adapt to upscaled icons properly. This should also be an option where players can define by what factor they want to scale the UI.

max-carroll (approved):
Nothing really to approve here, looks ok

max-carroll (commented):
oh didnt see others comments

PR #1312, opened 2020-02-05
Code polish

tomluchowski:
Some code polishing + rotatable cannon + port to ogre 1 12 2 ....


materials/scripts/Bow.material
```diff
@@ -11,7 +11,7 @@ material Bow : RTSS/NormalMapping_MultiPass
             // Override the normal map.
             rtshader_system
             {
-                lighting_stage normal_map BowNormal.png object_space 0 bilinear 1 -1.0
+                lighting_stage normal_map BowNormal.png tangent_space 0 bilinear 1 -1.0
```
paroj, line 14: I tried your branch. There are visible corruptions using tangent_space, the meshes seem to expect object_space. Rather then doing this you should upgrade to Ogre v1.12.5, which fixes this bug

tomluchowski: IT's nice to see someone carrying for OD when others don't. Many thanks for looking at my code.  I will, do what you suggest, but some time must pass for OpenSuse to prepare proper packages -- btw. have you ever thought of preparing ogre packages so make install cooperates with checkinstall ? Yeah , that's a wishfull thinking , someone could excpect that from OD as well ... OR I will acquire Ogre 1. 12. 5 by cmd 'make install', but I want this code to be useful for others and most people won't install libraries by 'make install' ...

paroj: on linux there is no easy way to distribute a C++ SDK, as compilers  and STL versions vary between distributions. Also even with the SDK, you would have to re-build the other deps like cegui yourself.

To work around all of this, I use snapcraft which allows building everything from scratch in an easy way, e.g. OD: https://github.com/paroj/OpenDungeons/blob/snapcraft/snapcraft.yaml

PR #1313, opened 2020-02-19
Rename appdata to metainfo

Polynomial-C:
See https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html
chapter 2.1.2 "Filesystem locations"

Signed-off-by: Lars Wendler <polynomial-c@gentoo.org>


akien-mga:
The file should also be renamed `.metainfo.xml`.

akien-mga:
Sorry for the delay. PR looks good, I force pushed a rebase to hopefully fix the CI build, let's see how it goes.

akien-mga:
Oh well Travis CI is shutting down so that's not going to fix CI anyway... Changes are safe enough so let's merge.

akien-mga:
Thanks!

PR #1316, opened 2020-03-08
use upstream RTShaderLib with Ogre 1.12+

paroj:
@tomluchowski this way it is backwards compatible


CMakeLists.txt
```diff
@@ -510,7 +510,9 @@ find_package(OIS REQUIRED)
 find_package(OGRE)
 if(NOT ${OGRE_FOUND})
     find_package(OldOGRE REQUIRED)
+    set(RTSHADER_DIR "materials/RTShaderLib")
 else()
+    set(RTSHADER_DIR "${OGRE_MEDIA_DIR}/RTShaderLib") # use upstream version
```
paroj, line 515: @tomluchowski with this line the ogre folder is referenced. No need to copy anything

tomluchowski: aha ;)

PR #1324, opened 2023-01-06
Fix Lighting

paroj:
- lighting is now done in a single pass instead of drawing everything 3x
- Ogre 13 required
- also includes hand positioning fix by @tomluchowski 

 before | after
---- | ----
before | after


oyvindln:
Oh, nice. I'm not really actively keeping up this project and not sure if anyone else is either at the moment but can do a merge if you want. Been considering trying to port the logic and assets to godot or some other game engine but I already have another foss project I'm busy with.

If you are interested on working on this further we could maybe add you as a maintainer if @akien-mga is fine with it.

paroj:
> If you are interested on working on this further we could maybe add you as a maintainer

I dont know whether I will find the time to actually improve the game logic, but I can keep the engine in a workable state so maybe someone else can.
I currently,  also do the snap packaging https://snapcraft.io/opendungeons to give this at least some visibility as I think OD is too good to let it die just like that..

